PR #2, opened 2019-11-25
Initial stab at sgx-sys crate

npmccallum:
The create test currently fails. It needs more work. Likewise, lots of
stuff is unimplemented. However, this should be good enough to start the
shape of the library.


sgx-sys/src/ioctl/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::convert::TryInto;
+use std::io::{Error, Result};
+use std::os::raw::{c_int, c_uint, c_ulong, c_void};
+use std::os::unix::io::AsRawFd;
+use std::ptr::null;
+
+pub mod sgx;
+
+extern "C" {
+    pub fn ioctl(fd: c_int, request: c_ulong, ...) -> c_int;
```
lkatalin: Learned something new here about variadic functions being allowed in FFIs.

sgx-sys/src/ioctl/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub trait Ioctl {
+    const REQUEST: c_ulong;
+
+    fn ioctl(&self, fd: &impl AsRawFd) -> Result<c_uint> {
+        let r = unsafe {
+            ioctl(
+                fd.as_raw_fd(),
+                Self::REQUEST,
+                self as *const _,
```
lkatalin: I'm not familiar with this syntax and will look it up.

sgx-sys/src/ioctl/sgx.rs
```diff
@@ -0,0 +1,85 @@
...
+#[derive(Debug)]
+pub struct Create<'a>(u64, PhantomData<&'a ()>);
+
+impl<'a> Ioctl for Create<'a> {
+    const REQUEST: c_ulong = 1_074_308_096; // SGX_IOC_ENCLAVE_CREATE
+}
+
+impl<'a> Create<'a> {
+    pub fn new(secs: &'a crate::secs::Secs) -> Self {
+        Create(secs as *const _ as _, PhantomData)
```
lkatalin: Same for this, I need to learn this syntax.

sgx-sys/src/ioctl/sgx.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct AddPages<'a> {
+    src: u64,
+    offset: u64,
+    length: u64,
+    secinfo: u64,
+    flags: PageFlags,
+    phantom: PhantomData<&'a ()>,
```
lkatalin: This is just here so that we can use the lifetime `<'a>` for the struct, right?

sgx-sys/src/ioctl/sgx.rs
```diff
@@ -0,0 +1,85 @@
...
+#[derive(Debug)]
+pub struct SetAttribute<'a>(u64, PhantomData<&'a ()>);
+
+impl<'a> Ioctl for SetAttribute<'a> {
+    const REQUEST: c_ulong = 1_074_308_099; // SGX_IOC_ENCLAVE__SET_ATTRIBUTE
+}
+
+impl<'a> SetAttribute<'a> {
+    pub fn new(fd: &'a impl AsRawFd) -> Self {
+        SetAttribute(fd.as_raw_fd() as _, PhantomData)
```
lkatalin: This one makes a little more sense to me because we want to return something that contains an unnamed `u64`.

sgx-sys/src/ioctl/sgx.rs
```diff
@@ -0,0 +1,85 @@
...
+// TODO: Implement Init::new()
+//impl<'a> Init<'a> {
+//    pub fn new(sig: &'a crate::sig::Sig) -> Self {
+//        Init(sig as *const _ as _, PhantomData)
+//    }
+//}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct SetAttribute<'a>(u64, PhantomData<&'a ()>);
```
lkatalin: I need to look into `<&'a ()>` ...

sgx-sys/src/ioctl/mod.rs
npmccallum: The syntax is just a cast from a reference to a pointer. The underscore here simply means: "Rust please infer the type." Rust can infer the type because you can only cast from a reference to a pointer of the same type.

sgx-sys/src/ioctl/sgx.rs
npmccallum: Cast the `Secs` reference to a `Secs` pointer. Then cast the `Secs` pointer to a `u64`. Have Rust infer the types (`_`).

sgx-sys/src/ioctl/sgx.rs
npmccallum: Because of the kernel interface, we are passing a pointer cast to a `u64`. This discards the lifetime of the reference. We need to make sure that Rust doesn't delete the object behind the reference while we still point to it as a `u64`. The `PhantomData` basically makes the object reference live as long as the struct we're instantiating.

sgx-sys/src/ioctl/sgx.rs
npmccallum: This is just a reference with lifetime `'a` to a struct with no contents. It is pretty much the Rust equivalent to `void*` in C. We're basically saying that the `SetAttribute` instance, and therefore the `u64` which is really a pointer, has a lifetime of `'a`. But since we have no way to do that on a `u64`, we do it on a reference inside `PhantomData` which the compiler will remove.

PR #4, opened 2019-11-26
Add existing CI tests from other Enarx repos

### LINKED ISSUE #3: Port existing Travis test suite from other repos
Currently, we run a few tests on Enarx's other repos via Travis. With development in the namesake Enarx repo picking up, we'd like those tests here too.

mbestavros:
`enarx/enarx` now has a Travis config! Resolves #3.

Tagging @npmccallum for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+# Copyright 2019 Red Hat
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+language: rust
+rust:
+  - stable
+  - beta
+  - nightly
+matrix:
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
+install:
+  - rustup component add rustfmt
+  - cargo install --force cargo-audit
+script:
+  - cargo fmt -- --check
+  - cargo audit
+  - cargo test


.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
+install:
+  - rustup component add rustfmt
+  - cargo install --force cargo-audit
+  - cargo generate-lockfile
+script:
+  - cargo fmt -- --check
+  - git rev-list origin/master.. | xargs -n1 -I{} sh -c '! git rev-parse {}^2'
```
steveeJ: This line is currently failing on CI. I'm not sure how to fix it though. What's the purpose of it?

mbestavros: Merge commit detection. It was working in GH Actions, but I'll need to tweak it to make it work in Travis.

npmccallum:
@mbestavros Let's disable the merge commit detection for now and get *some* tests running. We can come back to this.

mbestavros:
@npmccallum Agreed, and done. The current test run still fails b`enclave_create` and needs a `cargo fmt` run, but otherwise is good to go.

steveeJ:
> The current test run still fails b`enclave_create` and needs a `cargo fmt` run, but otherwise is good to go.

We need to gate the platform-specific code with appropriate features (or more ideal specific platform configs but I doubt they're available today. This means we need to add something like `#[cfg(feature = "test-sgx") ]` above https://github.com/enarx/enarx/blob/f7e44895f22802c366484a9e7cf00cbcf017a116/sgx-sys/src/lib.rs#L121 and then call the tests on CI without that feature. We can enable that feature on SGX capable CI runners in the future.

mbestavros:
I believe @lkatalin said she'd be adding that check in standup today -- but it sounded like we wanted a check that figured out if the box was SGX2 capable without needing to set a feature.

Either way, happy to make changes to the test config if needed once that's in place.

steveeJ:
> but it sounded like we wanted a check that figured out if the box was SGX2 capable without needing to set a feature.

@lkatalin I'm curious what your plan was. I'm in favor of a feature either way, which we could set after determining SGX2 capability on CI before we run cargo.

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+matrix:
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
+install:
+  - rustup component add rustfmt
+  - cargo install --force cargo-audit
+  - cargo generate-lockfile
+script:
+  - cargo fmt -- --check
```
npmccallum: Currently, this fails. Can you submit a second commit which makes this pass?

lkatalin:
@steveeJ @mbestavros It looks like we want something like [this](https://github.com/enarx/sev/blob/cbc5d20a7eab4a3e5910f6b1f6111715bba13b11/tests/api.rs#L48):
`#[cfg_attr(has_sgx)]`

based on a test that is run with a [build script](https://github.com/enarx/sev/blob/master/build.rs).

I can add these now. How should I add to this PR? @mbestavros Should I PR into your fork, or...?

npmccallum:
@lkatalin Just push as an independent PR. I'll merge that first.

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+  - beta
+  - nightly
+matrix:
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
+install:
+  - rustup component add rustfmt
+  - cargo install --force cargo-audit
+  - cargo generate-lockfile
```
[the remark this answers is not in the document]
mbestavros: That makes sense to me. As long as the test still works without it (and it should) I think that's valuable. I think it was just a holdover from other repo configs that don't have a lockfile committed.

Either way, I've removed it in the latest revision.

.travis.yml
npmccallum: Nevermind. I already did it.

PR #5, opened 2019-11-26
create passing

lkatalin:
```
running 3 tests
test secs::test::align ... ok
test secs::test::size ... ok
test test::enclave_create ... ok

```

### DIFF
--- a/sgx-sys/src/lib.rs
+++ b/sgx-sys/src/lib.rs
@@ -58,6 +58,8 @@ mod test {
         secs.base = 8192;
         secs.ssa_frame_size = 4096;
         secs.attributes = secs::Attributes::MODE_64_BIT;
+        secs.miscselect = secs::MiscSelect::EXINFO;
+        secs.xfrm = (secs::Xfrm::X87 | secs::Xfrm::SSE);
         Enclave::create(&secs).unwrap();
     }
 }
--- a/sgx-sys/src/secs.rs
+++ b/sgx-sys/src/secs.rs
@@ -32,6 +32,23 @@ bitflags! {
     }
 }
 
+bitflags! {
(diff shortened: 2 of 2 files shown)


sgx-sys/src/lib.rs
```diff
@@ -58,6 +58,8 @@ mod test {
         secs.base = 8192;
         secs.ssa_frame_size = 4096;
         secs.attributes = secs::Attributes::MODE_64_BIT;
+        secs.miscselect = secs::MiscSelect::EXINFO;
+        secs.xfrm = 0x3;
```
npmccallum: First off, awesome job!

The big question is: what belongs in `xfrm`?

Currently, `xfrm` is defined as a `u64` (https://github.com/enarx/enarx/blob/master/sgx-sys/src/secs.rs#L50). This probably isn't correct. Is this a bundle of flags? If so, we should turn this into a bitflags type (see: https://github.com/enarx/enarx/blob/master/sgx-sys/src/secs.rs#L28).

lkatalin: Thanks - I'm looking into this today.

npmccallum: @lkatalin I can confirm that your patch indeed does fix the tests.

lkatalin: @npmccallum That's great! 

I turned the `xfrm` into a set of bigflags based on the description of `xcr0` from https://en.wikipedia.org/wiki/Control_register and turned on the last two bits, so the test still passes. I'm still not sure if this is the correct value. I may send an email about it.

npmccallum: Good enough for me.

lkatalin:
@npmccallum I updated the commit message; hopefully this is better?

PR #12, opened 2019-12-04
Add fields and defaults for TCS, SigStruct

lkatalin:
Added fields to the TCS and SigStruct and some default implementations. Also a cargo fmt that seems to have adjusted some commas, etc.


lkatalin:
@npmccallum Let me know if this is closer to what you had in mind. I:

- Took out some implementations that may not be useful yet and just left `Default`s
- Used `default()` for fields that are zeros in structs that can't use `#[derive(Default)]`
- Used `#[derive(Default)]` for structs that can use it (i.e. ones without long arrays as fields)
- Removed `openssl` and `chrono` dependencies for now (though I don't see a way around them for the future)

I checked on the whitespace. Any whitespace or comma changes are the result of `cargo fmt`. I'm also not sure why it's showing a diff, ex. in `Cargo.toml`, because my current version of that file also doesn't have whitespace after the end of line 10 (`bigflags`).

lkatalin:
PS thanks for the reference for BCD. I'll check on this once we add the `chrono` dependency back in.

npmccallum:
@lkatalin A dependency is certainly necessary in the future. The question is **who** has the dependency.

PR #13, opened 2019-12-05
Add the paged library

npmccallum:


npmccallum:
Everyone, please tell me if this is a bad idea.

Basically, I got sick of manually managing all the alignment and padding myself. I also found that alignment and padding was required in some cases, but not others, for the same data type. If this approach is sane, it allows us to decouple alignment and padding from the type definitions. All the tests pass.

paged/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+pub trait PageSize: Copy + Clone + Default {
```
steveeJ: missing docstring

npmccallum:
@steveeJ I implemented a bunch of additional, optional traits. I also added documentation. The only major change is that `Page::new()` is now `impl<S, T> From<T> for Page<S, T>`.

paged/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+//!
+//! struct Bar(u32, [u8; 4096]);
+//!
+//! assert_eq!(size_of::<Bar>(), size_of::<(u32, [u8; 4096])>());
+//! assert_eq!(align_of::<Bar>(), align_of::<u32>());
+//! assert_eq!(size_of::<Page<Size4k, Bar>>(), 8192);
+//! assert_eq!(align_of::<Page<Size4k, Bar>>(), 4096);
+//! ```
+//!
+//! This means that you get the alignment and padding when you need it,
```
steveeJ: Would you mind pointing out (again) what _this_ refers to? I expected an aha-effect but didn't experience one, so I'm likely misunderstanding something.

paged/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+//! assert_eq!(size_of::<Foo>(), 8);
+//! assert_eq!(align_of::<Foo>(), align_of::<u32>());
+//!
+//! // The size and allocation of Page<Size4k, Foo> are paged.
+//! assert_eq!(size_of::<Page<Size4k, Foo>>(), 4096);
+//! assert_eq!(align_of::<Page<Size4k, Foo>>(), 4096);
+//!
+//! struct Bar(u32, [u8; 4096]);
+//!
+//! // The size and allocation of Foo are normal.
```
steveeJ:
```suggestion
//! // The size and allocation of Bar are normal.
```

paged/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+//! assert_eq!(size_of::<Bar>(), size_of::<(u32, [u8; 4096])>());
+//! assert_eq!(align_of::<Bar>(), align_of::<u32>());
+//!
+//! // The size and allocation of Page<Size4k, Foo> are paged.
```
steveeJ:
```suggestion
//! // The size and allocation of Page<Size4k, Bar> are paged.
```

paged/src/lib.rs
npmccallum: Fixed.

paged/src/lib.rs
npmccallum: Fixed.

PR #14, opened 2019-12-05
Split up sgx-sys into three crates

npmccallum:
sgx-types: low-level types from the SGX docs
  sgx-traits: high-level enclave API traits
     sgx-sys: implementation of the traits using ioctl()s


sgx-sys/src/lib.rs
```diff
@@ -12,80 +12,104 @@
...
+                size_of_val(&page),
+            )
+        };
+
+        let si = SecInfo::new(Perms::R | Perms::W, PageType::Tcs);
+        self.add_slice(&data, offset, Flags::MEASURE, si)
+            .map(|_| unsafe { Handle::new(offset) })
+    }
+
+    fn add_struct<T>(
```
lkatalin: I would like to understand a bit more about how this function `add_struct` differs from `add_slice` down below and the use cases for each. Why are they two separate functions? And why are they both different from `add_tcs`? I don't yet understand the different behavior we want to get from each (although I see the use of the new `Handle` and `Page<Size4k, _>`).

sgx-traits/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+// limitations under the License.
+
+use std::marker::PhantomData;
+use std::io::Result;
+
+use sgx_types::{secs::Secs, tcs::Tcs, sigstruct::SigStruct};
+use sgx_types::secinfo::{SecInfo, Flags as Permissions};
+use bitflags::bitflags;
+
+/// An offset reference with neither read nor write capabilities
```
lkatalin: Was there more to this sentence?

sgx-traits/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl<'h, T> Handle<'h, T> {
+    pub unsafe fn new(ptr: usize) -> Self {
+        Handle(ptr as u64, PhantomData)
+    }
+}
+
+bitflags! {
+    pub struct Flags: u64 {
+        const MEASURE = 1 << 0;
```
lkatalin: Where in the docs does this come from? I can't find a reference.

sgx-traits/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    fn new(secs: Secs) -> Result<Self>;
+
+    fn add_tcs(
+        &mut self,
+        tcs: Tcs,
+        offset: usize
+    ) -> Result<Handle<'b, Tcs>>;
+
+    fn add_struct<T>(
```
lkatalin: Same question here about the difference between `add_struct` and `add_slice`.

sgx-types/src/tcs.rs
```diff
@@ -1,10 +1,9 @@
-const SGX_TCS_RESERVED_SIZE: usize = 4024;
-
 /// Thread Control Structure (TCS) is an enclave page visible in its address
 /// space that defines an entry point inside the enclave. A thread enters inside
 /// an enclave by supplying address of TCS to ENCLU(EENTER). A TCS can be entered
 /// by only one thread at a time.
-#[repr(C, align(4096))]
+#[derive(Debug, Default)]
```
lkatalin: I'm happy to see that we can use this now that we don't need the large `reserved` field as padding.

sgx-types/src/ssa.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl ExitInfo {
+    pub fn new(et: ExitType, exc: Exception) -> Self {
+        let et = match et {
+            ExitType::Hardware => 0b011 << 8,
+            ExitType::Software => 0b110 << 8,
+        };
+
+        let exc = match exc {
+            Exception::Divider => 0x00,
```
lkatalin: Where are these values from?

sgx-types/src/ssa.rs
```diff
@@ -0,0 +1,135 @@
+#![allow(clippy::unreadable_literal)]
+
+#[derive(Copy, Clone, Debug)]
+pub enum Exception {
+    Divider,
```
lkatalin: How do we know that these are all the types of `Exception`s? (Curious where this came from.)

sgx-types/src/ssa.rs
```diff
@@ -0,0 +1,135 @@
...
+            Exception::FloatingPoint => 0x10,
+            Exception::AlignmentCheck => 0x11,
+            Exception::Simd => 0x13,
+        };
+
+        ExitInfo(1 << 31 | et | exc)
+    }
+
+    pub fn exit_type(self) -> Option<ExitType> {
+        const MASK: u32 = 1 << 31 | 0b111 << 8;
```
lkatalin: Same question about these values.

sgx-types/src/secinfo.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+bitflags! {
+    pub struct Flags: u8 {
+        const R = 1 << 0;
+        const W = 1 << 1;
+        const X = 1 << 2;
+        const PENDING = 1 << 2;
+        const MODIFIED = 1 << 2;
+        const PR = 1 << 2;
```
lkatalin: Maybe add a comment that `PR` is restricted permissions and that this comes from 38.11.11.

sgx-types/src/secinfo.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+#[repr(u8)]
+pub enum PageType {
+    Secs = 0,
+    Tcs = 1,
+    Reg = 2,
+    Va = 3,
```
lkatalin: I found a reference for this in 38.18.

sgx-sys/src/lib.rs
npmccallum: Basically, everything is just `add_slice()`. The `add_struct()` and `add_tcs()` functions are just convenience wrappers.

sgx-traits/src/lib.rs
npmccallum: Nope. The first line of a doc comment is a sentence fragment with no period.

sgx-traits/src/lib.rs
npmccallum: It doesn't. This flag is part of the kernel interface. Basically, the docs describe two instructions: `EADD` and `EMEASURE`. The kernel exposes just one call `add_pages()` and offers a flag for whether or not to measure them.

sgx-traits/src/lib.rs
npmccallum: `add_slice()` is the kernel API. The other two are convenience wrapper functions.

sgx-types/src/ssa.rs
npmccallum: They are the Intel CPU exception constants. I found a list in the SGX docs. But I'd like to improve this section generally.

sgx-types/src/ssa.rs
npmccallum: See above. I'd like to see if the `x86_64` crate has these types. If not, we should submit them and link against `x86_64`...

sgx-types/src/ssa.rs
npmccallum: These are in the SGX docs. This section still needs improvement.

sgx-types/src/secinfo.rs
npmccallum: Yes, please. And actually, the `add_pages()` kernel `ioctl()` just ignores most of the flags here. So that needs to be documented too. Lots of room to improve still.

sgx-types/src/secinfo.rs
npmccallum: Great! Maybe submit an improvements PR on top of this?

PR #15, opened 2019-12-06
Conditional compilation for enclave_create() test

lkatalin:


sgx-sys/build.rs
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+fn main() {
+    use std::path::Path;
+
+    if Path::new("/dev/sgx").exists() {
```
npmccallum: We probably want to test for `/dev/sgx/enclave`.

sgx-sys/src/lib.rs
```diff
@@ -117,6 +117,7 @@ impl<'e> EnclaveTrait<'e> for Enclave<'e> {
 mod test {
     use super::*;
 
+    #[cfg(has_sgx)]
```
npmccallum: This code conditionally compiles the test. We don't want that. We want to conditionally skip (ignore) the test. See how `sev` does this.

Why the difference? Ignored tests are still compiled. There is testing value in ensuring the tests still compile. Further, ignored tests are still included in the testing output, showing explicitly that they were skipped. This helps with debugging.

lkatalin: Okay, it makes sense now, thanks. I thought we didn't want to compile it.

PR #18, opened 2019-12-06
Remove outdated alignment and size tests in sgx-types/secs.rs

lkatalin:
These tests are outdated now that the `paging` crate exists.


npmccallum:
Tests fail for a different reason. This patch still improves things. I'll merge it.

PR #19, opened 2019-12-06
Remove reserved fields

lkatalin:
- ~Removed `reserved` fields since we don't need them for alignment/sizing~
- Removed extraneous `bitflags` line
- Changed doc comment to regular comment to fix build error


lkatalin:
@npmccallum Updated this; no longer removing `reserved` fields.

PR #21, opened 2019-12-06
Created types for Xfrm, Attributes, MiscSelect in sgx-types

lkatalin:
Also updated the `enclave_create()` test in `sgx-sys` to use these new types and keep the test from failing.

### DIFF
--- a/sgx-sys/src/lib.rs
+++ b/sgx-sys/src/lib.rs
@@ -106,15 +106,13 @@ mod test {
     #[cfg_attr(not(has_sgx), ignore)]
     #[test]
     fn enclave_create() {
-        use sgx_types::secs::*;
+        use sgx_types::*;
 
         let mut secs = Secs::default();
         secs.size = 8192;
         secs.base = 8192;
         secs.ssa_frame_size = 4096;
-        secs.attributes = Attributes::MODE_64_BIT;
-        secs.miscselect = MiscSelect::EXINFO;
-        secs.xfrm = Xfrm::X87 | Xfrm::SSE;
+        secs.xfrm = xfrm::Xfrm::default();
         Builder::new(secs).unwrap();
     }
 }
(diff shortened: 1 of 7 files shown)


sgx-types/src/miscselect.rs
```diff
@@ -0,0 +1,12 @@
+bitflags::bitflags! {
+    /// Section 38.7.2
+    pub struct MiscSelect: u32 {
+        const EXINFO = 1 << 0;
+    }
+}
+
+impl Default for MiscSelect {
+    fn default() -> Self {
+        MiscSelect::EXINFO
```
npmccallum: I think we need to default to `MiscSelect::empty()`. We can't be sure that `EXINFO` is setup by whoever calls this.

sgx-sys/src/lib.rs
```diff
@@ -106,15 +106,15 @@ mod test {
...
         let mut secs = Secs::default();
         secs.size = 8192;
         secs.base = 8192;
         secs.ssa_frame_size = 4096;
-        secs.attributes = Attributes::MODE_64_BIT;
-        secs.miscselect = MiscSelect::EXINFO;
-        secs.xfrm = Xfrm::X87 | Xfrm::SSE;
+        secs.attributes = attributes::Attributes::MODE_64_BIT;
+        secs.miscselect = miscselect::MiscSelect::EXINFO;
+        secs.xfrm = xfrm::Xfrm::X87 | xfrm::Xfrm::SSE;
```
npmccallum: I don't think these lines are required anymore because these values are the defaults.

sgx-types/src/miscselect.rs
npmccallum: Never mind, the kernel enforces the presence of this flag. Leave it as it is.

sgx-sys/src/lib.rs
lkatalin: It still wants `xfrm` specified to pass the test. I think this is because `SECS` defaults to `std::mem::zeroed()`.

PR #24, opened 2019-12-10
Split testaso!() into three test functions

npmccallum:
This makes it both clearer what testaso!() does and it provides separate
output for each kind of test during test execution.


lkatalin:
I think this is a good change.

PR #25, opened 2019-12-10
Clean up Offset type

npmccallum:
This removes the lifetime enforcement I thought we could make work, but realized we can't.

### DIFF
--- a/sgx-types/src/lib.rs
+++ b/sgx-types/src/lib.rs
@@ -73,28 +73,12 @@ pub mod tcs;
 
 use core::marker::PhantomData;
 
-/// An offset reference with neither read nor write capabilities
-///
-/// The Offset struct allows the creation of an opaque reference to a
-/// type that cannot be read or written. Neither the lifetime nor the
-/// type are discarded. This allows us to refer to an offset inside
-/// an enclave without fear that it will be dereferenced. The size of
-/// the Offset is always 64 bits with natural alignment. Therefore,
-/// the Offset type can be embedded in structs.
 #[repr(transparent)]
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


sgx-types/src/lib.rs
```diff
@@ -73,28 +73,12 @@ pub mod tcs;
...
-/// The Offset struct allows the creation of an opaque reference to a
-/// type that cannot be read or written. Neither the lifetime nor the
-/// type are discarded. This allows us to refer to an offset inside
-/// an enclave without fear that it will be dereferenced. The size of
-/// the Offset is always 64 bits with natural alignment. Therefore,
-/// the Offset type can be embedded in structs.
 #[repr(transparent)]
 #[derive(Debug)]
-pub struct Offset<'h, T>(u64, PhantomData<&'h T>);
+pub struct Offset<T: ?Sized>(u64, PhantomData<T>);
```
lkatalin: Looks like this removes the lifetime and also the `Sized` bound, which I don't totally understand yet - why do we not want `T` to be `Sized`?

npmccallum:
The lifetime does not work because we need to write a lifetime that is bound to either the `Builder` and the `Enclave`. This is because the `Offset` is valid starting with the `Builder` object and then ending with the `Enclave` object. But we destroy the `Builder` to create the `Enclave`. There is no way to migrate the lifetime of the `Object`. We could tie it to the lifetime of the inner `File` object. But there is no way to expose the lifetime of the inner object to the outside. So we're stuck.

sgx-types/src/lib.rs
npmccallum: Because `[u8]` is not Sized. This is helpful for the `add_slice()` function.

lkatalin:
That's a good explanation! Thank you!

PR #26, opened 2019-12-10
Implement StateSaveArea

npmccallum:


sgx-types/src/ssa.rs
```diff
@@ -118,8 +125,57 @@ pub struct GprSgx {
     pub gsbase: u64,
 }
 
+bitflags! {
+    /// Section 38.9.2.2, Table 38-13
+    pub struct PageFault: u32 {
```
lkatalin: This is a good addition.

Issue #29, opened 2019-12-11
SGX: Add `from_hash()` implementation to `Signature`


lkatalin:
The hash is a SHA256 of an enclave binary.

mbestavros:
Abandoned.

PR #33, opened 2019-12-11
Add implementation for Signature created from mrenclave hash

lkatalin:
Refers to #29

### DIFF
--- a/sgx-types/src/sig.rs
+++ b/sgx-types/src/sig.rs
@@ -110,6 +110,20 @@ pub struct Signature {
     pub q2: RsaNumber, // value used in RSA signature verification
 }
 
+impl Signature {
+    /// Creates a signature from a mrenclave hash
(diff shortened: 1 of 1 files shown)


npmccallum:
@lkatalin I'm not sure what the value of a constructor from a hash without the other information is.

sgx-types/src/sig.rs
```diff
@@ -110,6 +110,20 @@ pub struct Signature {
     pub q2: RsaNumber, // value used in RSA signature verification
 }
 
+impl Signature {
+    /// Creates a signature from a mrenclave hash
+    pub fn from_hash(hash: [u8; 32]) -> Self {
```
steveeJ: would it make sense to introduce a stronger type for the hash?

npmccallum:
@lkatalin What's the status of this PR?

lkatalin:
Closed because we are doing this from the `sgx-hasher` crate instead.

PR #37, opened 2019-12-11
Add Apache 2.0 LICENSE file

npmccallum:


haraldh:
Don't we want LICENSE per crate?

npmccallum:
@haraldh Probably yes. However, GitHub looks for this file in particular to advertise the license.

PR #40, opened 2019-12-12
add vmsyscall crate

haraldh:


vmsyscall/src/error.rs
```diff
@@ -0,0 +1,321 @@
+//! errno definition
+//!
+//! copied from Redox
```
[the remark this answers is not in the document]
haraldh: yeah, we might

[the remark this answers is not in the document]
npmccallum: @haraldh I agree with @steveeJ . No code copying please. Should we reject this PR altogether?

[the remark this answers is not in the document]
haraldh: I'll update the PR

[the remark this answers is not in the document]
haraldh: To clarify, this is only one file out of the whole crate
https://gitlab.redox-os.org/redox-os/syscall
and I just need the errno consts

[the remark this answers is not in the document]
haraldh: maybe I should just use https://crates.io/crates/errno-codes and add more dependencies

haraldh:
here we go again

vmsyscall/src/error.rs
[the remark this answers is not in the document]
haraldh: meh, `errno-codes` is not no_std

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    Mmap(Result<usize, Error>),
+    Mremap(Result<usize, Error>),
+    Munmap(Result<i32, Error>),
+    Mprotect(Result<i32, Error>),
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub enum Error {
+    Errno(i32),
+    NotImplemented,
```
npmccallum: Isn't there already `ENOSYS`?

haraldh: right

haraldh:
> Why is this dual licensed?

all my work was dual licensed... can change

haraldh:
updated licencse and added errno and syscall consts

npmccallum:
@haraldh I need the syscall and errno consts too. Can we make a separate crate for these? Should we make them enums?

haraldh:
removed the syscall and errno consts

haraldh:
includes no main Cargo.toml and Cargo.lock changes

haraldh:
added Cargo.toml ... otherwise travis is useless

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! syscall serialize/deserialize
+//!
+//! Currently it uses a hard coded page and an I/O trigger.
+//! We might want to switch to MMIO.
+
+#![no_std]
+
```
steveeJ: Maybe ` #![deny(missing_docs)]` and add docs to satisfy it?

haraldh: done

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 Red Hat, Inc.
```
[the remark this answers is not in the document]
haraldh: done

npmccallum:
@haraldh I think this needs a rebase.

vmsyscall/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+authors = ["Harald Hoyer <harald@redhat.com>"]
+license = "Apache-2.0"
+edition = "2018"
+
+categories = [
+    "no-std",
+]
+
+keywords = [
+    "errno",
```
npmccallum: `errno` is now a different crate.

vmsyscall/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+    "syscall",
+    "no_std",
+]
+
+[dependencies]
+serde = { version = "1.0", default-features = false, features = ["derive"] }
```
npmccallum: I don't remember how cargo handles `no_std` in this case. But I think the consumer of this crate will also have to specify `no_std` as a feature for `serde`. This is just a note for my future recall. :)

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! syscall serialize/deserialize
+//!
+//! Currently it uses a hard coded page and an I/O trigger.
+//! We might want to switch to MMIO.
+
+#![deny(missing_docs)]
```
npmccallum:
```suggestion
#![deny(missing_docs)]
#![deny(clippy:all)]
```

vmsyscall/Cargo.toml
haraldh: At least for serde_cbor it works this way. No_std only has to be defined by the implementing serialize/deserialize crate...

vmsyscall/Cargo.toml
haraldh: Right

vmsyscall/Cargo.toml
haraldh: Default is "std"
```
[features]
default = ["std"]

# Provide derive(Serialize, Deserialize) macros.
derive = ["serde_derive"]

# Provide impls for common standard library types like Vec<T> and HashMap<K, V>.
# Requires a dependency on the Rust standard library.
std = []
```

vmsyscall/src/lib.rs
haraldh: done

vmsyscall/Cargo.toml
haraldh: removed

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// int madvise(void *addr, size_t length, int advice);
+    Madvise {
+        /// see madvise(2)
+        addr: usize,
+        /// see madvise(2)
+        length: usize,
+        /// see madvise(2)
+        advice: i32,
+    },
+    /// void *mmap(void *addr, size_t length, int prot, int flags, …);
```
steveeJ: What does `…` hint at?

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// int mprotect(void *addr, size_t len, int prot);
+    Mprotect {
+        /// see mprotect(2)
+        addr: usize,
+        /// see mprotect(2)
+        length: usize,
+        /// see mprotect(2)
+        prot: i32,
+    },
+    // Todo: extend with needed hypervisor proxy syscalls
```
steveeJ: Who do we find out which syscalls are needed? Are these the ones required by the runtime which is added in #139?

vmsyscall/src/lib.rs
haraldh: Left out unneeded parameters

vmsyscall/src/lib.rs
haraldh: For the sev kernel case we only proxy those, which the kernel can't handle itsself.. like network related syscalls like socket(2). And if the kernel does not want to do extensive memory handling, we want to proxy the mmap family, too.

haraldh: currently this is just a skeleton crate, which will be extended and stripped a lot in the future.

vmsyscall/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+//!
+//! Currently it uses a hard coded page and an I/O trigger.
+//! We might want to switch to MMIO.
+
+#![deny(missing_docs)]
+#![deny(clippy::all)]
+#![no_std]
+
+/// The IO trigger port for vm syscall proxy
+pub const TRIGGER_PORT: u16 = 0xFF;
```
npmccallum: Sorry for one more question. `TRIGGER_PORT` seems to me a VM specific detail which we won't use in process-based TEEs. Is this crate the right place for it?

haraldh: oh .. right..

haraldh: removed

PR #41, opened 2019-12-12
Add basic functionality for RSA key in Hasher struct

lkatalin:
This is a WIP. It is modeled on the `Builder`. So far a `Hasher` includes an RSA key and has some basic functionality to return properties of that key. TODO: getting the hash of SECS.


lkatalin:
Abandoned in favor of #49

PR #49, opened 2019-12-16
Add basic hashing functionality for enclave Hasher

lkatalin:
Co-authored by: Nathaniel McCallum <npmccallum@redhat.com>


sgx-hasher/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+authors = ["Lily Sturmann <lsturman@redhat.com>", "Nathaniel McCallum <npmccallum@redhat.com>" ]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+sgx-types = { path = "../sgx-types" }
+sgx-traits = { path = "../sgx-traits" }
+sgx-sys = { path = "../sgx-sys" }
```
npmccallum: This dependency does not appear to be needed.

sgx-hasher/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+        &mut self,
+        offset: usize,
+        data: &[u8],
+        flags: sgx_traits::Flags,
+        secinfo: page::SecInfo,
+    ) {
+        const EEXTEND: u64 = 0x00444E4554584545;
+        const EADD: u64 = 0x0000000044444145;
+
+        let si = secinfo.as_ref();
```
npmccallum: I don't think we need a separate variable for this. Let's do `&secinfo.as_ref()[..48]` below.

sgx-hasher/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+                self.0.update(&EEXTEND.to_le_bytes());
+                self.0.update(&eoff.to_le_bytes());
+                self.0.update(&[0u8; 48]);
+
+                self.0.update(&data[poff..][..256]);
+            }
+        }
+    }
+
+    // Returns SHA256 hash of MRENCLAVE
```
npmccallum: This text is confusing. It seems to imply that `MRENCLAVE` is being hashed. Rather, this function is producing `MRENCLAVE`.

sgx-traits/src/lib.rs
```diff
@@ -16,7 +16,12 @@ use std::io::Result;
 
 use bitflags::bitflags;
 use sgx_types::page::{Flags as PageFlags, SecInfo};
-use sgx_types::{secs::Secs, sig::Signature, tcs::Tcs, Offset};
+use sgx_types::{
+    secs::Secs,
+    sig::Signature,
+    tcs::Tcs,
+    Offset,
+};
```
npmccallum: What is the purpose of this change?

sgx-hasher/src/lib.rs
lkatalin: That's a good point.

sgx-traits/src/lib.rs
lkatalin: It's `cargo fmt`. In general I'm still not sure if `cargo fmt` changes should be their own PR... that seems sort of cluttered for the `git log`? But otherwise it ends up getting folded in, like this. Open to suggestions.

npmccallum: The question is why is `cargo fmt` making this change? The previous commits had `cargo fmt` run on them. So we shouldn't be seeing this.

npmccallum: I'll merge this PR and figure it out afterwards.

lkatalin: I just ran it again and it went away, so I'm confused about it too. Let me fix it and I'll re-push.

npmccallum:
@lkatalin Tests are failing. Also, lots of warnings are being introduced. Please don't add new warnings.

lkatalin:
@npmccallum Odd, I'll check on these and fix them.

PR #53, opened 2019-12-16
Sign MRENCLAVE hash to create SIGSTRUCT

lkatalin:
Co-authored by: Nathaniel McCallum <npmccallum@redhat.com>


sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
...
+    array
+}
+
+pub fn sign(hash: &[u8; 32], key: &rsa::Rsa<Private>) -> Result<sig::Signature> {
+    let author = sig::Author::new(sig::Vendor::default(), 0u32, 0u32);
+
+    let contents = sig::Contents::new(
+        misc::MiscSelect::EXINFO,
+        attr::Attributes::new(attr::Flags::MODE_64_BIT, attr::Xfrm::all()),
+        *hash,
```
npmccallum: Instead of deref'ing here, just change the function signature.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
...
+        0,
+        0,
+    );
+
+    let rsa_key = PKey::from_rsa(key.to_owned())?;
```
npmccallum: Instead of cloning the key here, change the function signature. The caller can decide whether to clone or move.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
     }
 }
 
+pub fn to_array(bn: bn::BigNum) -> [u8; 384] {
+    let mut array = [0u8; 384];
+    array.copy_from_slice(&bn.to_vec());
+    array
+}
+
+pub fn sign(hash: &[u8; 32], key: &rsa::Rsa<Private>) -> Result<sig::Signature> {
```
npmccallum: We probably want to take `author` and `contents` as arguments here.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
...
+
+    let rsa_key = PKey::from_rsa(key.to_owned())?;
+    let md = hash::MessageDigest::sha256();
+    let mut signer = sign::Signer::new(md, &rsa_key)?;
+    signer.update(author.as_ref())?;
+    signer.update(contents.as_ref())?;
+    let signature = signer.sign_to_vec()?;
+
+    let e = hex::encode(key.e().to_vec());
+    let e = u32::from_str_radix(&e[..], 16).unwrap();
```
npmccallum: Don't use `hex` for this. We might want to get upstream `openssl` to `impl TryFrom<BigNum> for (u8, u16, u32, u64, u128, usize)`.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
...
+    let mut sigmodq1 = bn::BigNum::new()?;
+
+    s_2.sqr(&s, &mut ctx)?; // sig^2
+    q1.div_rem(&mut qr, &s_2, &m, &mut ctx)?; // floor(sig^2 / modulus)
+
+    s_3.checked_mul(&s, &s_2, &mut ctx)?; // sig^3
+    sigmod.checked_mul(&s, &m, &mut ctx)?; // sig * modulus
+    sigmodq1.checked_mul(&q1, &sigmod, &mut ctx)?; // sig * modulus * q1
+    s_3minus.checked_sub(&s_3, &sigmodq1)?; // sig^3 - (q1 * sig * modulus)
+    q2.div_rem(&mut qr, &s_3minus, &m, &mut ctx)?; // floor((sig^3 - q1 * sig * modulus) / modulus)
```
npmccallum: I don't think this algorithm is correct. See: https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/sgxsign.c#L362

Also, I'd really prefer to use the Rust ops. The `openssl` crate appears to implement them. But I couldn't get them to work. That's probably something to pursue with upstream.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +81,68 @@ impl Hasher {
     }
 }
 
+pub fn to_array(bn: bn::BigNum) -> [u8; 384] {
```
npmccallum: This function probably shouldn't be `pub`. It also appears to ignore endianness issues. The array is little endian but openssl is big endian.

sgx-hasher/src/lib.rs
lkatalin: I'm trying to implement this the way it is implemented here https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/sgxsign.c#L362. But their algorithm seems to say `s * qr == s^3 - q1 * s * m`, and I don't understand why this must be true. 

By Rust ops do you mean the regular `*` and `/` operators? Should I extract `u32` from each `bignum` to do the calculations? Or did you want me to do an upstream fix with `openssl` as a prereq for this PR?

lkatalin: I will just adapt it from what you had, although I'm still not sure about the algorithm and I would like to understand it before pushing it.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +82,64 @@ impl Hasher {
...
+    let mut qr = bn::BigNum::new()?;
+
+    let sig = u32::try_from(s.as_ref()).unwrap();
+    let s2 = bn::BigNum::from_u32(sig * sig)?;
+    q1.div_rem(&mut qr, &(s2), &m, &mut ctx)?;
+
+    let rem = u32::try_from(qr.as_ref()).unwrap();
+    let modl = u32::try_from(m).unwrap();
+    let q2 = (sig * rem) / modl;
+    let q2 = bn::BigNum::from_u32(q2).unwrap();
```
npmccallum: We definitely should not convert to `u32` for this math. This block of code will fail at runtime because the `BigNum` will be too large to fit in `u32`.

sgx-hasher/src/lib.rs
```diff
@@ -77,6 +82,64 @@ impl Hasher {
...
+        author,
+        contents,
+        u32::try_from(key.e()).unwrap(),
+        m_array,
+        s_array,
+        q1_array,
+        q2_array,
+    ))
+}
+
```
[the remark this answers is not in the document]
lkatalin: I get a `the trait 'std::convert::From<()>' is not implemented for 'std::io::Error' error, but I can write an implementation for it.

sgx-hasher/src/lib.rs
lkatalin: Good point, however `binary operation '*' cannot be applied to type 'openssl::bn::BigNum'` ... so should I go back to using the `BigNum` ops? Or is there another way I am overlooking?

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +80,52 @@ impl Hasher {
     }
 }
 
+pub fn sign(
+    author: sig::Author,
+    contents: sig::Contents,
+    key: rsa::Rsa<pkey::Private>,
+) -> std::io::Result<sig::Signature> {
```
npmccallum: I'm wondering if it is better to use the `openssl` `ErrorStack` error type here...

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +80,52 @@ impl Hasher {
...
+    let md = hash::MessageDigest::sha256();
+    let mut signer = sign::Signer::new(md, &rsa_key)?;
+    signer.update(author.as_ref())?;
+    signer.update(contents.as_ref())?;
+    let signature = signer.sign_to_vec()?;
+
+    // Generates q1, q2 values for RSA signature verification
+    let s = bn::BigNum::from_slice(&signature)?;
+    let m = key.n();
+    let e = u32::try_from(key.e()).expect("Could not convert BigNum to u32");
```
npmccallum: `expect()` is not correct. I'm also worried that upstream rust-openssl isn't going to merge my `TryFrom` patches. We should probably write our own converter.

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +80,52 @@ impl Hasher {
...
+    let mut ctx = bn::BigNumContext::new()?;
+    let mut q1 = bn::BigNum::new()?;
+    let mut qr = bn::BigNum::new()?;
+
+    q1.div_rem(&mut qr, &(&s * &s), &m, &mut ctx)?;
+    let q2 = &(&s * &qr) / m;
+
+    // Returns modulus, signature, q1, and q2 as [u8; 384]
+    let mut m_array = [0; 384];
+    m.to_slice(&mut m_array).unwrap_or(());
```
npmccallum: `.unwrap_or(())` just swallows the error. Silent failures aren't good.

sgx-hasher/src/lib.rs
npmccallum: @lkatalin This still needs to be fixed.

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +80,52 @@ impl Hasher {
     }
 }
 
+pub fn sign(
+    author: sig::Author,
+    contents: sig::Contents,
+    key: rsa::Rsa<pkey::Private>,
+) -> std::io::Result<sig::Signature> {
+    // Generates signature on Signature author and contents
+    let rsa_key = pkey::PKey::from_rsa(key.clone())?;
```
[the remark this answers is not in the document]
npmccallum: @lkatalin This still needs to be fixed.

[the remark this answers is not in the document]
lkatalin: @npmccallum `q1.div_rem(&mut qr, &(&s * &s), &m, &mut ctx)?;` borrows the `key` through `&m`, so `borrow of moved value` errors are produced here for every permutation of ordering `m = key.n()`, `q1 = ...`, and `rsa_key = ...`. What am I missing?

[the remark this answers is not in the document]
lkatalin: `q1` depends on both `m` and the `signature`, and the `signature` depends on `rsa_key = ...`, so I can't move deriving the signature to the bottom.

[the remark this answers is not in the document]
npmccallum: Move `let m = ...` and `let e = ...` to the top of the function.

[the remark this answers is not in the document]
lkatalin: When I do that, I get:
```
   --> sgx-hasher/src/lib.rs:106:40
    |
103 |     let m = key.n();
    |             --- borrow of `key` occurs here
...
106 |     let rsa_key = pkey::PKey::from_rsa(key)?;
    |                                        ^^^ move out of `key` occurs here
...
120 |     q1.div_rem(&mut qr, &(&s * &s), &m, &mut ctx)?;
    |                                     -- borrow later used here

```

lkatalin:
I had to leave `openssl` pinned to your PR because otherwise I don't have access to the `to_slice` method.

sgx-hasher/src/lib.rs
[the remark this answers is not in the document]
npmccallum: Uncle. Let the `clone()` be retained.

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +79,55 @@ impl Hasher {
     }
 }
 
+fn bn_to_rsanum_inner(bn: bn::BigNum) -> Result<[u8; 384], openssl::error::ErrorStack> {
+    let mut array = [0u8; 384];
+    array.copy_from_slice(&bn.to_vec());
+    array.reverse();
+    Ok(array)
+}
```
npmccallum: This function never appears to return an error. Also, `copy_from_slice()` panics if the array size isn't correct. We definitely don't want to panic on user input. So you need a manual test to see if the `vec` is the right length and return an error if it isn't.

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +79,55 @@ impl Hasher {
...
+
+fn bn_to_u32(value: &bn::BigNum) -> Result<u32, openssl::error::ErrorStack> {
+    let mut buf = u32::default().to_ne_bytes(); // Zero
+    value
+        .to_slice(&mut buf)
+        .or(Err(openssl::error::ErrorStack::get()))?;
```
npmccallum: 1. I don't think that `ErrorStack::get()` works in this context because no openssl error was actually generated during `to_slice()`. So we need something different here. Maybe `std::io::Error` is the right error code (I know I told you to change it)?

2. Change `to_slice()` to `to_vec()` and copy the bytes around. Hopefully upstream will accept a version of `to_slice()` in the future.

sgx-hasher/src/lib.rs
```diff
@@ -79,6 +79,55 @@ impl Hasher {
...
+    Ok(u32::from_be_bytes(buf))
+}
+
+pub fn sign(
+    author: sig::Author,
+    contents: sig::Contents,
+    key: rsa::Rsa<pkey::Private>,
+) -> Result<sig::Signature, openssl::error::ErrorStack> {
```
npmccallum: I think `std::io::Error` might have been correct here since we need to also return something like `std::io::ErrorKind::InvalidInput` when the integer conversion fails. Sorry for misleading you.

sgx-hasher/src/lib.rs
lkatalin: I put in a manual check here.

lkatalin:
>  use std::io::Result when the error type is std::io::Error
Got it, thanks!

PR #56, opened 2019-12-17
Add ioctl

npmccallum:


sgx-types/src/attr.rs
```diff
@@ -37,6 +37,12 @@ pub struct Attributes {
     xfrm: Xfrm,
 }
 
+impl Attributes {
+    pub const fn new(flags: Flags, xfrm: Xfrm) -> Self {
```
lkatalin: Just curious why this is `const`.

sgx-types/src/secs.rs
```diff
@@ -55,7 +55,9 @@ testaso! {
 }
 
 impl Secs {
-    pub fn new(size: u64, base: u64, ssa: u32, contents: &Contents) -> Self {
+    pub const SIZE_MAX: u64 = 0x1_000_000_000;
```
lkatalin: This is a good addition.

sgx-traits/src/lib.rs
```diff
@@ -15,40 +15,30 @@
...
+pub trait Builder: Sized {
+    type Enclave: Enclave;
 
-    fn new(secs: Secs) -> Result<Self>;
+    fn new(secs: secs::Secs) -> Result<Self>;
 
-    unsafe fn add_tcs(&mut self, tcs: Tcs, offset: usize) -> Result<Offset<Tcs>>;
-
-    unsafe fn add_struct<T>(
+    unsafe fn add(
```
lkatalin: This seems to replace the functionality of `add_tcs`, `add_struct`, and `add_slice`, is that right?

sgx-sys/src/lib.rs
```diff
@@ -114,8 +82,51 @@ mod test {
...
+        let mut off = 0;
+        for f in &flags {
+            #[repr(C, align(4096))]
+            struct Page([u8; 4096]);
+            let page = Page([0u8; 4096]);
+
+            // TCS pages MUST NOT specify permissions.
+            let si = page::SecInfo::new(page::Flags::empty(), page::Class::Tcs);
+            eprintln!("{:?} {:?} {:?}", off, *f, si);
+            assert!(unsafe { builder.add(off, &page.0, *f, si).is_ok() });
```
lkatalin: Is this asserting that the builder was able to add a TCS page with the particular flag (`empty` for TCS) and secinfo combination at an offset that advances by one page each time?

sgx-types/src/attr.rs
npmccallum: `const` functions can be evaluated at compile time. For example: https://doc.rust-lang.org/unstable-book/language-features/const-fn.html

sgx-traits/src/lib.rs
npmccallum: Correct.

sgx-sys/src/lib.rs
npmccallum: No. `TCS` flags have no permissions. Therefore, the set of flags is empty.

jarkkojs (approved):

PR #59, opened 2019-12-18
Add tests to the hasher

npmccallum:
First we made the hasher produce the same output as the Linux kernel SGX
selftest. Then we made a bunch of tests to verify the algorithm doesn't
accidentally change. These tests have not yet been tested against the Intel
CPU, so they might still be incorrect in some corner cases. However, the
algorithm should be fairly correct at this point.


sgx-hasher/src/lib.rs
```diff
@@ -38,46 +36,185 @@ impl Hasher {
...
+
+        let contents = sig::Contents::new(
+            misc::MiscSelect::default(),
+            attr::Attributes::default(),
+            [0u8; 32],
+            0,
+            0,
+        );
+
+        let secs = secs::Secs::new(0, size.next_power_of_two(), 1, &contents);
```
lkatalin: So the `Secs` has a base of `0`, a size that is larger than the input size by a power of two, an `ssa` size of `1`, and the default `contents` from above. It may be helpful to include this as a comment.

sgx-types/src/page.rs
```diff
@@ -67,8 +65,20 @@ impl AsRef<[u8]> for SecInfo {
...
-        Self { flags, class }
+    pub fn reg(flags: Flags) -> Self {
+        Self {
+            flags,
+            class: Class::Reg,
+            reserved: Padding::default(),
+        }
+    }
+
+    pub fn tcs() -> Self {
```
lkatalin: This makes sense.

sgx-hasher/src/lib.rs
```diff
@@ -38,46 +36,185 @@ impl Hasher {
...
+    //
+    // The canonical source of correctness for this algorithm is, of
+    // course, the Intel SGX CPU/microcode/ME. If you can demonstrate a
+    // a case where we don't match this, we will happily change our ANSWERs.
+
+    const DATA: [u8; 4096] = [123u8; 4096];
+    const LONG: [u8; 8192] = [123u8; 8192];
+
+    fn hash(input: &[(&[u8], SecInfo)]) -> [u8; 32] {
+        let size = input.iter().fold(0, |c, x| c + x.0.len() as u64);
```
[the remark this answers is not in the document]
npmccallum: I added a comment in the latest push.

sgx-hasher/src/lib.rs
npmccallum: I added a comment in the latest push.

sgx-types/src/sig.rs
```diff
@@ -1,6 +1,22 @@
 use super::{attr::Attributes, misc::MiscSelect, utils::Padding};
 
 #[derive(Copy, Clone)]
+struct Number([u8; 384]);
```
lkatalin: This is a good idea since we're using this type of array multiple times, but could use a more descriptive name. What sort of Number is it? Why is a Number 384 bytes? Etc.

sgx-sys/src/lib.rs
```diff
@@ -76,6 +76,8 @@ mod test {
     fn builder_new() {
         let contents = sig::Contents::new(
```
lkatalin: This could possibly use a comment to say the second `misc` and `attr` are the masks.

sgx-types/src/sig.rs
npmccallum: Fixed.

sgx-sys/src/lib.rs
npmccallum: I actually haven't fully understood this part of SGX yet. I didn't take the masks in the constructor before because I was being lazy. But the `selftest` code required me to differentiate between the two.

PR #65, opened 2020-01-02
Add iocuddle library

npmccallum:
The iocuddle library is a library that enables one to move ioctl unsafe
blocks from the call site to the definition site. This aids in reducing the
number of unsafe blocks in the code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,10 @@ version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "iocuddle"
+version = "0.1.0"
+
+[[package]]
 name = "lazy_static"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+    "iocuddle",
     "sgx-types",
     "sgx-traits",
     "sgx-sys",
--- a/iocuddle/Cargo.toml
+++ b/iocuddle/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "iocuddle"
+version = "0.1.0"
+authors = ["Nathaniel McCallum <npmccallum@redhat.com>"]
+edition = "2018"
+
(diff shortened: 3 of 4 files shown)


npmccallum:
@steveeJ My first attempt used traits. The resulting use wasn't very ergonomic. You had to import a bunch of things just to make the interfaces work.

`Write`, `Read`, and `WriteRead` closely match the semantics of the kernel interface.

So basically, you have a syntax like this:

```rust
const FOO: Ioctl<Write, &Bar>;
```

That declares an ioctl named `FOO` that passes an reference to a `Bar` instance from userspace to the kernel (i.e. `Write`).

If the ioctl is a "classic" ioctl, it is defined this way, where the ioctl request number is allocated by the kernel (as a kernel constant):

```rust
const FOO: Ioctl<Write, &Bar> = unsafe { Ioctl::classic(0x0000) };
```

If the ioctl is a "modern" ioctl, the request number is composed of multiple inputs: a `group` identifier, an `nr` identifier, a `direction` constant and the size of the argument type. So the constructor looks like this:

```rust
const BAZ: Group = Group::new(0x00)
const FOO: Ioctl<Write, &Bar> = unsafe { BAZ.write(0x00) };
```

nishant0078:
WASM as in Windows Assembly?

nishant0078:
WASM as in Windows Assembly?

axelsimon:
No, we are referring to Web Assembly, the target binary format we are using for Enarx keeps.
Besides the information on Wikipedia, the Mozilla Developer Network has a good summary of WASM: https://developer.mozilla.org/en-US/docs/WebAssembly

Another useful piece of info: while WASM was initially thought up for browsers, it has since expanded beyond and we are more specifically targeting the WASI specification (Web Assembly System Interface). More (great) info here: https://hacks.mozilla.org/2019/03/standardizing-wasi-a-webassembly-system-interface/

Hope this helps!

On 5 January 2020 11:11:07 UTC, Nishant Mishra <notifications@github.com> wrote:
>WASM as in Windows Assembly?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/enarx/enarx/pull/65#issuecomment-570901117

-- 
axel simon
mail/jabber/matrix: axelsimon@axelsimon.net
twitter: @axelsimon
tel: +33 6 08040144

PR #64, opened 2020-01-02
Update for the Kernel v24 interfaces

npmccallum:

PR #67, opened 2020-01-02
Migrate sgx-sys to use iocuddle

npmccallum:
Only the last commit (5171a5f) is new material. The rest are from PRs #64 and #65.


iocuddle/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+pub struct Group(u8);
+
+impl Group {
+    /// Create a new group for related `ioctl`s from its allocated number
+    pub const fn new(value: u8) -> Self {
+        Self(value)
+    }
+
+    const unsafe fn make<'a, D, T>(self, nr: u8, dir: c_ulong) -> Ioctl<D, &'a T> {
+        const SIZE_BITS: c_ulong = 14;
```
steveeJ: what's this number?

iocuddle/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+
+        r.try_into().or_else(|_| Err(Error::last_os_error()))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    const KVMIO: Group = Group::new(0xAE);
```
steveeJ: Could we have an enum for all possible Groups, or is it not possible to maintain an exhaustive list?

PR #70, opened 2020-01-03
Remove use of sgx-traits from sgx-hasher

npmccallum:


npmccallum:
@steveeJ Not for the hasher. We also need to avoid a circular dependency.

PR #97, opened 2020-01-08
Add sgx-show

npmccallum:
This program exists to gather and display data relating to SGX hardware, SGX BIOS enablement, SGX kernel drivers and potentially SGX-related software for use in debugging.


sgx-show/src/main.rs
```diff
@@ -0,0 +1,176 @@
+//! This program exists to gather and display data relating to SGX hardware,
+//! SGX BIOS enablement, SGX kernel drivers and potentially SGX-related
+//! software for use in debugging.
+//!
+//! It works by running a hierarchical tree of tests and displaying the
+//! information gathered therefrom. If a parent test fails to validate,
+//! the child test are still called to produce their output, but do not
```
steveeJ:
```suggestion
//! the child tests are still called to produce their output, but do not
```

sgx-show/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+mod exec;
+mod sink;
+
+use exec::*;
+use sgx_types::{attr, misc};
+
+use data::cpuid::{CpuId, CpuInfo, ExtFeat, SgxAttr, SgxCaps};
+use sink::{bit::Bit, debug::Debug, mask::Mask};
+
+pub trait Convert<T>: Sized {
```
steveeJ: Could we use the existing https://doc.rust-lang.org/std/convert/trait.TryInto.html instead here?

sgx-show/src/data/cpuid.rs
```diff
@@ -0,0 +1,81 @@
...
+pub trait Leaf: Display {
+    const LEAF: u32;
+    const SUBL: u32;
+}
+
+pub struct CpuId<T: Leaf>(pub u128, PhantomData<T>);
+
+impl<T: Leaf> CpuId<T> {
+    pub fn eax(&self) -> u32 {
+        (self.0 >> 96) as u32
```
steveeJ:
```suggestion
        (self.0 >> 3 * 32) as u32
```

sgx-show/src/data/cpuid.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    pub fn ebx(&self) -> u32 {
+        (self.0 >> 64) as u32
```
steveeJ:
```suggestion
        (self.0 >> 2 * 32) as u32
```

sgx-show/src/data/cpuid.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    pub fn ecx(&self) -> u32 {
+        (self.0 >> 32) as u32
```
steveeJ:
```suggestion
        (self.0 >> 1 * 32) as u32
```

sgx-show/src/data/cpuid.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    pub fn edx(&self) -> u32 {
+        (self.0 >> 00) as u32
```
steveeJ:
```suggestion
        (self.0 >> 0 * 32) as u32
```

sgx-show/src/main.rs
npmccallum: Fixed.

sgx-show/src/main.rs
npmccallum: Originally, no; which is why I created `Convert`. But because of other changes that is now possible. I have moved this to `TryFrom`. So: fixed.

sgx-show/src/data/cpuid.rs
npmccallum: Fixed.

sgx-show/src/data/cpuid.rs
npmccallum: Fixed.

sgx-show/src/data/cpuid.rs
npmccallum: Fixed.

sgx-show/src/data/cpuid.rs
npmccallum: Fixed.

npmccallum:
@steveeJ I've addressed all your issues. Is it possible for you to remove the "changes requested" state without approving the whole PR? (Honest question.)

sgx-show/src/sink/bit.rs
```diff
@@ -0,0 +1,14 @@
...
+pub struct Bit<T: Leaf> {
+    pub reg: fn(&CpuId<T>) -> u32,
+    pub bit: u8,
+}
+
+impl<T: Leaf> super::Sink for Bit<T> {
+    type Type = CpuId<T>;
+
+    fn test(&self, data: &Self::Type) -> bool {
+        (self.reg)(data) & (1 << self.bit) != 0
```
lkatalin: I don't totally understand this line and `(self.reg)(data)`. `data` is a reference to a `CpuId<T>` and `reg` is a function to turn that into a `u32`? What do the parens do here?

sgx-show/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::convert::TryFrom;
+
+use exec::*;
+use sgx_types::{attr, misc};
+
+use data::cpuid::{CpuId, CpuInfo, ExtFeat, SgxAttr, SgxCaps};
+use sink::{bit::Bit, debug::Debug, mask::Mask};
+
+impl TryFrom<CpuId<SgxCaps>> for misc::MiscSelect {
+    type Error = ();
```
lkatalin: Why is the Error type `()` here? Shouldn't there be more information returned for error handling?

sgx-show/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+        Self(bytesize::ByteSize(1 << bits))
+    }
+}
+
+fn main() {
+    let test = Test {
+        name: "Intel CPU",
+        data: Box::new(CpuInfo),
+        sink: Box::new(Mask {
+            mask: 0x00000000ffffffffffffffffffffffffu128,
```
lkatalin: I assume these values come from Intel documentation. Is the value masked because only the first bits are relevant? Or is the mask for some other purpose? Will it potentially change?

sgx-show/src/sink/mod.rs
```diff
@@ -0,0 +1,15 @@
+pub mod bit;
+pub mod debug;
+pub mod mask;
+
+pub trait Sink {
+    type Type;
+
+    fn info(&self, _value: &Self::Type) -> Option<String> {
+        None
```
lkatalin: It seems like this will *always* return `None`; is there something I'm missing about how this is used?

sgx-show/src/sink/bit.rs
npmccallum: `self.reg` is a function pointer. So we are calling the function pointed to by `self.reg`. This is a function to extract one particular register from the CpuId struct. See, for example: https://github.com/enarx/enarx/pull/97/files#diff-187ebd96da7420d347a4eae0e18468b6R124

sgx-show/src/sink/mod.rs
npmccallum: This is a default function which will be overridden by specific implementations of `Sink`. The default outputs nothing.

sgx-show/src/main.rs
npmccallum: The error is always discarded: https://github.com/enarx/enarx/pull/97/files#diff-8c921bd2af53e89a5b8942f9d0f785d7R44

sgx-show/src/main.rs
npmccallum: See here for details: https://en.wikipedia.org/wiki/CPUID#EAX=0:_Highest_Function_Parameter_and_Manufacturer_ID

EAX contains the highest CPUID leaf. EBX, ECX and EDX contain the manufacturer identifier. We mask out EAX.

sgx-show/src/sink/mod.rs
lkatalin: Ahh, I see. Ex. here https://github.com/enarx/enarx/pull/97/files#diff-28457ac1732bda7b910ee565ab93f102R12

npmccallum: Exactly.

sgx-show/src/main.rs
lkatalin: Why do we want to discard the error by converting the `Result` into an `Option` there?

npmccallum: Because we don't report specific errors at this time. Basically, the test succeeds only if all of the conditions apply:
1. Data collection succeeds.
2. Data type conversion succeeds.
3. Data evaluation succeeds.

If all three conditions are true, we get a green check. Otherwise, we get a red 'x'.

PR #106, opened 2020-01-10
Add conference issue template

axelsimon:
Issue template for conference, automatically adds the 'conference' label.

### DIFF
--- a/.github/ISSUE_TEMPLATE/conferences.md
+++ b/.github/ISSUE_TEMPLATE/conferences.md
@@ -11,12 +11,12 @@ Conference / Event URL
 Any other relevant information
 
 **Dates**
-//Some of these are optional, enter what you can
-- CFP opens:     DD – DD Month YYYY
-- CFP closes:    DD – DD Month YYYY
-- CFP results:   DD – DD Month YYYY
-- Slides due:    DD – DD Month YYYY
-- Event:         DD – DD Month YYYY
+//Some of these are optional, enter what you can. If the event is over several days, use the YYYY-MM-DD/YYYY-MM-DD date format to represent the range.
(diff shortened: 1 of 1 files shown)


axelsimon:
This PR should be clean, thanks to @mbestavros's help. Let me know if anything needs to be changed.

npmccallum:
@axelsimon This mostly looks good. However, can we use ISO 8601 for dates?

axelsimon:
@mbestavros and i discussed using ISO format dates (which i usually use everywhere), but i argued against it for two reasons, in this case:
- readability: 2020-01-13 is harder on the eyes than 13 January 2020
- displaying a range

I've now just learned ISO 8601 has a solution for ranges :)

For the sake of being able to automate tagging and closing conference issues in the future by easily parsing dates, i will switch the dates to ISO 8601.

axelsimon:
@mbestavros and i discussed using ISO format dates (which i usually use everywhere), but i argued against it for two reasons, in this case:
- readability: 2020-01-13 is harder on the eyes than 13 January 2020
- displaying a range

I've now just learned ISO 8601 has a solution for ranges :)

For the sake of being able to automate tagging and closing conference issues in the future by easily parsing dates, i will switch the dates to ISO 8601.

npmccallum:
@axelsimon Any update on this?

axelsimon:
Yes, as the `master` branch has moved forward, working on updating my branch cleanly, without getting a merge commit.

axelsimon:
If all is right, i updated my branch to include a new commit that modifies the issue template to use the ISO 8601 format for dates, answering @npmccallum requested change.
Let me know if this is ok now :)

PR #121, opened 2020-01-13
Add documentation for SGX in sgx-crypto and sgx-types.

lkatalin:
Fixes #28 and fixes #58 for `sgx-types` and `sgx-crypto` crates (no other crates were modified). 

Added `#![deny(missing_docs)]` to these crates, so it is possible that the requirements are too verbose at this point, for example in `pub struct Gpr` in `sgx-types/src/ssa.rs` where all registers have to be labeled. 

Edit: Also added the licensing info to each file in these two crates.

PR #126, opened 2020-01-14
Create isv module for SVN and ProdId types

### LINKED ISSUE #122: Make data types for `ISV prod ID` and `ISV SVN`
They can be wrapped similar to `Vendor`.

lkatalin:
Fixes #122. Should be rebased on top of #121 when that is merged.

### DIFF
--- a/sgx-types/src/lib.rs
+++ b/sgx-types/src/lib.rs
@@ -85,6 +85,7 @@ macro_rules! testaso {
 }
 
 pub mod attr;
+pub mod isv;
 pub mod misc;
 pub mod page;
 pub mod secs;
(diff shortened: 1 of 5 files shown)


lkatalin:
@npmccallum I re-structured it the way you described and the build passes now.

sgx-types/src/lib.rs
```diff
@@ -85,6 +85,8 @@ macro_rules! testaso {
 }
 
 pub mod attr;
+/// Makes SIGSTRUCT's ISV SVN and ISV Prod ID types available (Table 38-19).
```
npmccallum: I think this comment is unneeded.

PR #139, opened 2020-01-16
wasm: initial libruntime and runtime

steveeJ:
TODO:
- [x] discuss and possibly rework approach for platform specific workload loading
-  [x] workload retrieval from fd
    - [x] impl
    - [x] test
    - [x] docstrings
- [x] workload run
    - [x] impl
    - [x] test
    - [x] docstrings
- [x] ~README~ will use docstrings instead
- [x] LICENSE


wasm-libruntime/src/workload.rs
```diff
@@ -0,0 +1,130 @@
...
+    impl FdWorkloadRetriever {
+        pub fn new(fd: Option<i32>) -> Self {
+            Self(fd.unwrap_or(DEFAULT_WORKLOAD_FD))
+        }
+    }
+
+    impl super::WorkloadRetriever for FdWorkloadRetriever {
+        fn retrieve(&self) -> Result<Workload> {
+            let mut buf = Vec::new();
+            unsafe { File::from_raw_fd(self.0) }
```
steveeJ: @haraldh will this work in the kernel you're working on?

haraldh: Ok, something has opened fd 3 for you. Maybe the parent pid, which started you.

But I guess in our case, the bytecode is first transferred over the net. Right?

So, something has to establish the connection and request that payload.

Now, if that something can ensure, that the (maybe ssl transparent) socket has fd 3, then we are good.

Otherwise, we might want to replace `FdWorkloadRetriever` with another `WorkloadRetriever`

@npmccallum ?

haraldh: Scratch my last comment.

haraldh: It will work.. yes.. no problem

haraldh: What I would like to see is a `ReadWorkloadRetriever`
so instead of:
```
let workload_retriever = fd_workload_retriever::FdWorkloadRetriever::new(Some(fd));
```
it will be
```
let workload_retriever = read_workload_retriever::ReadWorkloadRetriever::new(Some(file));
```

with 
pub struct FdWorkloadRetriever<S: Read>(read: S);

because `File` implements Read and &[u8] implements Read... So we can pass a predownloaded slice also.

steveeJ: Thanks for the input on the design, which is exactly the remaining item on the TODO list in this PRs description :smiley: I'll tinker with it a bit more!

wasm-runtime/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["Stefan Junker <mail@stefanjunker.de>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+wasm-libruntime = { path = "../wasm-libruntime" }
+
+env_logger = "0.7"
+log = "0.4"
```
npmccallum: We need a newline at the end of the file. Please adjust your editor to enforce this.

wasm-libruntime/src/workload.rs
```diff
@@ -0,0 +1,166 @@
...
+        let mut module_registry: HashMap<String, wasmtime::Instance> = HashMap::new();
+
+        // Instantiate WASI
+        let wasi = wasmtime_wasi::instantiate_wasi(
+            // preopened_dirs: &[(String, File)],
+            Default::default(),
+            // argv: &[String],
+            Default::default(),
+            // environ: &[(String, String)],
+            Default::default(),
```
npmccallum: Just a comment here for longer terms thoughts. When a tenant submits a workload, we probably want them to also supply deployment metadata. This might include things like:
  * argv
  * environ
  * a whitelist of envvars to accept from the host

wasm-libruntime/src/workload.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.buf.len()
+    }
+
+    /// Returns `true` if the workload buffer has a length 0
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Runs a WASM workload, invoking the given function with the given arguments, returning the result vector.
+    pub fn run(&self, function_name: &str, args: &[Val]) -> Result<Vec<Val>> {
```
npmccallum: Is there a standard `main()` function for WASM/WASI yet?

steveeJ: According to [wasmtime](https://github.com/bytecodealliance/wasmtime/blob/177af5357880c33863cbe3363754000dbd23bddb/src/commands/run.rs#L282-L291) there's a default command export and the fallback to `_start`.

wasm-libruntime/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# master on 2020-01-22
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", package = "wasmtime", rev = "9a88d3d8941ef5c03dfd156956dd045ab3d8ce2b" }
+wasmtime-jit = { git = "https://github.com/bytecodealliance/wasmtime", package = "wasmtime-jit",  rev = "9a88d3d8941ef5c03dfd156956dd045ab3d8ce2b" }
+wasmtime-interface-types = { git = "https://github.com/bytecodealliance/wasmtime", package = "wasmtime-interface-types",  rev = "9a88d3d8941ef5c03dfd156956dd045ab3d8ce2b" }
+wasmtime-wasi = { git = "https://github.com/bytecodealliance/wasmtime", rev ="9a88d3d8941ef5c03dfd156956dd045ab3d8ce2b" }
```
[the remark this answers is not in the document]
steveeJ: I wanted to include a recent change, specifically https://github.com/bytecodealliance/wasmtime/pull/789.

wasm-libruntime/src/workload.rs
steveeJ: Without having seen this comment, I did start designing a workload specification mechanism. Please see the latest revision of this PR. I'm looking forward to receiving feedback on it!

wasm-runtime/Cargo.toml
steveeJ: Done

keep-libruntime/fixtures/return_1.wkld
```diff
@@ -0,0 +1,8 @@
+{
+  "modules": {
+    "return_1": [0,97,115,109,1,0,0,0,1,133,128,128,128,0,1,96,0,1,127,3,130,128,128,128,0,1,0,4,132,128,128,128,0,1,112,0,0,5,131,128,128,128,0,1,0,1,6,129,128,128,128,0,0,7,145,128,128,128,0,2,6,109,101,109,111,114,121,2,0,4,109,97,105,110,0,0,10,138,128,128,128,0,1,132,128,128,128,0,0,65,1,11]
+  },
+  "function_module":"return_1",
```
haraldh: space after `:` .. my eyes! :smile:

keep-libruntime/src/workload.rs
```diff
@@ -0,0 +1,257 @@
...
+    ) -> Self {
+        Self {
+            modules,
+            function_module,
+            function_name,
+            function_args,
+        }
+    }
+
+    /// Retrieves a serliazed WASM workload from a Reader.
```
haraldh: typo: `serliazed`

keep-libruntime/fixtures/return_1.wkld
```diff
@@ -0,0 +1,8 @@
...
+  "function_name":"main",
```
haraldh: how about `main::return_1` ? instead of splitting it?
Or:
```json
"function" : {
   "module": "return_1",
   "name": "main",
   "args": []
}
```

keep-libruntime/src/workload.rs
```diff
@@ -0,0 +1,257 @@
...
+                        if let Some(export) = instance.get_export(field_name) {
+                            Ok(export.clone())
+                        } else {
+                            Err(format!(
+                                "Import {} was not found in module {}",
+                                field_name, module_name
+                            ))
+                        }
+                    } else {
+                        Err(format!("Import module {} was not found", module_name))
```
haraldh: String errors are not very good for libraries. Ok, there is not much action to do in the error case.. Just a thought

keep-libruntime/fixtures/return_1.wkld
steveeJ: I generated this and merely added newlines. I don't expect the tenant to write plain JSON. We can generate this from the command line even. I'm thinking of a CLI which takes an argument for all of the keys given here, i.e.
```console
`enxctl generate-workload --modules="./app.wasm,./somelib.wasm" --function_module="app" --function-name="main" --- "command_arg_arg-passed_to_function" --long-arg-passed="to_function".
```

In addition we can offer to parse a YAML or TOML for this purpose.

npmccallum:
@wgwoods This is the PR for the start of our WASM runtime.

keep-libruntime/fixtures/return_1.wkld
steveeJ: I'm not convinced that it's worth to split this out into a separate structure, but I'm also not opposed, so let's try it.

keep-libruntime/src/workload.rs
steveeJ: I wonder where this error will be propagated to. If a user is supplying the workload synchronously it's easy to display it, but if there's an orchestrator involved it's not that obvious.

npmccallum:
@steveeJ I'd really like to review this. Can you fix up the build failures and rebase on master? Thanks!

steveeJ:
> My only nitpick is that `wasm-workload` is in a directory named `keep-libruntime`.

I can see that, however finding a clear line between these hasn't been the primary goal at this point.

> However, I've thought a bit about the workload system here and I'm concerned we're trying to design something in code that should probably go through the RFC process. In the meantime, I think it would be better to just take the arguments on the command line. Something like:

I think you're describing a theoretical issue. I don't consider this stable whatsoever, and none of the crates version numbers suggests that either. I agree that we should discuss the workload handling via an RFC. The reason why I based this prototype on reading in via file-descriptors is that it's easier to demo as an embedded runtime in an environment with no filesystem or terminal, and therefore the following didn't, and still doesn't, make any sense in this context.

> `keep-runtime <wasm> <module_name> <function_name> [args...]`

In addition to the above thoughts this would be achievable using the vanilla wasmtime CLI today.

mbestavros:
Superseded by https://github.com/enarx/enarx/pull/502; closing.

PR #146, opened 2020-01-16
Add General Protection Fault case in sgx-types/src/ssa.rs

### LINKED ISSUE #135: Add GP faults to errcd in sgx-types/src/ssa.rs
It seems reasonable to add general protection faults to the `errcd`, as in the SGX documentation an `errcd` could refer to a page fault or a general protection fault. Currently, ours refers only to page faults with the `PageFault` struct.

lkatalin:
Attempts to fix #135.

### DIFF
--- a/sgx-types/src/ssa.rs
+++ b/sgx-types/src/ssa.rs
@@ -179,6 +179,7 @@ pub struct Gpr {
 }
 
 bitflags! {
+    /// Flags for a page fault;
     /// Section 38.9.2.2, Table 38-13
     pub struct PageFault: u32 {
         /// Same as non-SGX page fault exception P flag.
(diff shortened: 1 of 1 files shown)


lkatalin:
@npmccallum I think I still need some guidance on this. I wrote a test for the size of the `Fault` enum, and it passes (the size is always 8). I also tried the same test with other enums and with different sizes, and I got expected results, so I think the test works; maybe I'm not testing in the right way.

lkatalin:
This new version of the PR has the following improvements:
- `ExceptionInfo` is the correct size, with an `errcd` field that is a `u32`.
-  The `ExceptionInfo` struct has the `reserved` field added and its `testaso!` updated to reflect this.
- A small fix in `PageFault` for the `SGX` bitflag (should be 15 according to the docs).
- An enum encompassing GPFault and Pagefault variants, holding appropriate information.
- A `pub fault` method on `Footer` that converts from the `Footer`'s "inner" types that match the SGX documentation to the `enum Fault` type.

Some outstanding questions:
- Would it make sense to also return `ExitType` in the `enum Fault`? Will we want to know if it was a hardware or software exit?
- Would it make sense to add an error type that is returned by `some_footer.fault()` for an invalid fault type that shouldn't be contained in `errcd`?

@npmccallum Let me know what you think.

sgx-types/src/ssa.rs
```diff
@@ -254,6 +316,37 @@ impl Footer {
...
+
+            Exception::GeneralProtection => {
+                let gpflags = GenProtFault::from_bits(errcd).unwrap();
+                Some(Fault::GP { flags: gpflags })
+            }
+            _ => {
+                return None;
+            }
+        }
+    }
```
npmccallum: Try to reduce the number of temporary variables you are creating. Also, I think this code would benefit from this pattern:

```rust
match (self.gpr.exitinfo.exit_type(), self.gpr.exitinfo.exception()) {
    (Some(et), Some(Exception::PageFault)) => { ... }
    (Some(et), Some(Exception::GeneralProtection)) => { ... }
    _ => None,
}
```

Also,

sgx-types/src/ssa.rs
```diff
@@ -192,19 +193,80 @@ bitflags! {
...
 /// Section 38.9.2.1, Table 38-12
 #[derive(Debug)]
 #[repr(C)]
 pub struct ExceptionInfo {
     /// In case of a page fault, contains the linear address that caused the fault.
     pub maddr: u64,
     /// Exception error code for GP fault or page fault.
-    /// TODO: Should this have more options in case of general purpose fault?
-    pub errcd: PageFault,
+    pub errcd: u32,
```
npmccallum: I think we want this struct and all its members to be private.

sgx-types/src/ssa.rs
```diff
@@ -192,19 +193,80 @@ bitflags! {
...
+        /// The flags provide extra information on the fault.
+        flags: GenProtFault,
+    },
+    /// Page Fault with flags and memory address.
+    PF {
+        /// Contains the page fault linear address that caused a page fault.
+        addr: u64,
+        /// The flags provide extra information on the fault.
+        flags: PageFault,
+    },
```
npmccallum: Let's make these enum variants tuple structs. I think named fields is probably just adding clutter.

sgx-types/src/ssa.rs
lkatalin: You're right, that is much more concise.

sgx-types/src/ssa.rs
```diff
@@ -254,6 +308,20 @@ impl Footer {
             gpr,
         }
     }
+
+    /// Converts an internal fault type to a safe Rust enum describing the fault.
+    pub fn fault(&self) -> Option<Fault> {
+        match (self.gpr.exitinfo.exit_type(), self.gpr.exitinfo.exception()) {
+            (Some(_et), Some(Exception::PageFault)) => Some(Fault::PF(
```
npmccallum:
```suggestion
            (Some(_), Some(Exception::PageFault)) => Some(Fault::PF(
```

However, if we aren't going to use the exit type for anything, we should maybe just do this check up front. I think even just this should work...

```rust
self.gpr.exitinfo.exit_type()?;
```

Then it can be removed from the match statement to declutter it.

PR #149, opened 2020-01-20
Query CPUID to get enclave max size in sgx-types/src/secs.rs

### LINKED ISSUE #116: Write function using CPUID to obtain pub const SIZE_MAX in SECS
In `sgx-types/src/secs.rs`, the `pub const SIZE_MAX` value is currently hardcoded to `0x1_000_000_000`. We should instead write a function to obtain this value from `CPUID`; however, `CPUID` is not available from within an SGX enclave.

lkatalin:
I had to remove the `pub const SIZE_MAX`, since you can't assign the result of a function to a `const`. I checked the results of the function with what `sgx-show` produces for `SgxCaps` for 64bit and it was correct.

Fixes #116.

### DIFF
--- a/sgx-types/src/secs.rs
+++ b/sgx-types/src/secs.rs
@@ -20,6 +20,7 @@
 //! by the means of ENCLS(ECREATE) leaf.
 
 use super::{attr, isv, misc::MiscSelect, sig::Contents};
+use core::num::NonZeroU64;
 
 /// Section 38.7
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


sgx-types/src/secs.rs
```diff
@@ -66,8 +66,16 @@ testaso! {
 }
 
 impl Secs {
-    /// TODO: The max size of an enclave should come from CPUID, which is inaccessible from within SGX.
-    pub const SIZE_MAX: u64 = 0x1_000_000_000;
+    /// Returns the maximum enclave size for 64bit in bytes.
+    /// CPUID.(EAX=12H, ECX=0H) enumerates Intel SGX capability;
+    /// For more on CPUID enumeration leaves, see 37.7.2 and Table 37-4.
+    pub fn max_size_64bit() -> Option<u64> {
```
npmccallum: Name this `max_size()`. We don't support 32bit enclaves.

sgx-types/src/secs.rs
```diff
@@ -66,8 +66,16 @@ testaso! {
...
+        let leaf = 0x00000012;
+        let subleaf = 0x00000000;
+        let res = unsafe { core::arch::x86_64::__cpuid_count(leaf, subleaf) };
+        let max_size_64bit: u64 = 1 << (res.edx >> 8 as u8) as u64;
```
npmccallum: You can't just call CPUID on this leaf/subleaf. This will crash on CPUs that are not Intel or do not support sgx. So you need to test those leafs first.

lkatalin: That's a good point.

sgx-types/src/secs.rs
lkatalin: Fixed.

lkatalin:
- [x] Make max_size() unsafe for now. 
- [ ] Document the exception when CPUID isn't supported.
- [x] Drop the Intel CPU test.
- [x] Add a max leaf test (eax from leaf:0 subleaf:0).
- [x] Migrate leaf and subleaf to well-named constants.

I'm not sure how to go about documenting the exception without using `eflags`, because:
1. `__cpuid_count` doesn't have any type of error state; it always returns a `CpuidResult`.
2. Errors aren't available in `no_std` from what I can tell.

Any guidance appreciated for fixing this last part.

sgx-types/src/secs.rs
```diff
@@ -66,8 +66,35 @@ testaso! {
 }
 
 impl Secs {
-    /// TODO: The max size of an enclave should come from CPUID, which is inaccessible from within SGX.
-    pub const SIZE_MAX: u64 = 0x1_000_000_000;
+    /// Returns the maximum enclave size for 64bit in bytes.
+    /// CPUID.(EAX=12H, ECX=0H) enumerates Intel SGX capability;
+    /// For more on CPUID enumeration leaves, see 37.7.2 and Table 37-4.
```
npmccallum: To document why this function is unsafe, add something like this:

```
/// # Safety
/// This function is unsafe because it does not check if the `CPUID` instruction is available before issuing it.
```

In the future, we can always make this function safe.

sgx-types/src/secs.rs
```diff
@@ -66,8 +66,35 @@ testaso! {
...
+    pub unsafe fn max_size() -> Option<u64> {
```
npmccallum: I think we might actually want to return `Option<NonZeroU64>`.

sgx-types/src/secs.rs
```diff
@@ -66,8 +66,35 @@ testaso! {
...
+        // Test for max leaf size
+        let res = core::arch::x86_64::__get_cpuid_max(LEAF_MAX_PARAM);
+        let max_leaf = res.0;
+        if max_leaf < LEAF_SGX_SUPPORT || max_leaf < LEAF_MAX_ENCL_SIZE {
+            return None;
+        }
+
+        // Test for SGX support
+        let res = core::arch::x86_64::__cpuid_count(LEAF_SGX_SUPPORT, SUBLEAF_SGX_SUPPORT);
+        let sgx_support = res.ebx;
```
npmccallum: No need for this temporary variable.

sgx-types/src/secs.rs
lkatalin: Oh, you meant *literally* document! Sure thing.

PR #150, opened 2020-01-22
Implement Travis tests as git pre-push hooks

### LINKED ISSUE #43: Convert TravisCI tests to git pre-commit hooks
We currently run a bunch of sanity tests in TravisCI (things like checking code formatting, etc.). I'd like these converted to git pre-commit hooks.

We can move these to the `hooks/` directory and then use `git config core.hooksPath hooks` to ensure they are updated with the repo. Once this is done, we can configure TravisCI to run the pre-commit hooks instead.

This ensures that you hit the failures before submitting the PR.

mbestavros:
This PR moves all the tests we were previously running in Travis to shell scripts within the `hooks` folder. This allows us to run these tests before push time, ensuring that _every_ commit passes tests locally before being pushed remotely.

Since the tests now exist in their own file, Travis has also been updated to call that file (for organization). To add new tests, we should only need to add a new shell script to `hooks/pre-push.d`, and they will run both pre-push and on Travis.

Resolves #43.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,11 @@ matrix:
       rust: stable
       install:
         - cargo install --force cargo-audit
-      script:
-        - cargo audit
+      script: ./hooks/pre-push.d/cargo-audit --force
     - name: "sev"
       rust: stable
       script:
         - cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
 install:
   - rustup component add rustfmt
-script:
-  - cargo fmt -- --check
-  - cargo test
+script: ./hooks/pre-push
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -1,0 +1,2 @@
+#!/bin/bash -e
+for f in $0.d/*; do $f; done
--- a/hooks/pre-push.d/cargo-audit
+++ b/hooks/pre-push.d/cargo-audit
@@ -1,0 +1,3 @@
+#!/bin/bash -e
+[[ -n "$TRAVIS" && "$1" != "--force" ]] && exit 0
+cargo audit
--- a/hooks/pre-push.d/cargo-fmt
+++ b/hooks/pre-push.d/cargo-fmt
@@ -1,0 +1,2 @@
+#!/bin/bash -e
+cargo fmt -- --check
--- a/hooks/pre-push.d/cargo-test
+++ b/hooks/pre-push.d/cargo-test
@@ -1,0 +1,2 @@
+#!/bin/bash -e
+cargo test


mbestavros:
@npmccallum:

1. Yes, the tests run before the commit.
2. `cargo test` did abort the commit upon failure, but `cargo fmt` didn't. Stringing them together with `&&` seems to work.
3. Yes.

I've pushed a number of changes -- all tests are present in their own file now. I've also added `cargo audit` to the pre-commit hook; I kept it separate from the others so that CI can run it as a separate task. I've requested a re-review.

.travis.yml
```diff
@@ -27,14 +27,11 @@ matrix:
...
     - name: "sev"
       rust: stable
       script:
         - cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
 install:
   - rustup component add rustfmt
-script:
-  - cargo fmt -- --check
-  - cargo test
+script: ./hooks/pre-commit.d/sanity-checks
```
npmccallum:
```suggestion
script: ./hooks/pre-commit
```

.travis.yml
```diff
@@ -27,14 +27,11 @@ matrix:
       rust: stable
       install:
         - cargo install --force cargo-audit
-      script:
-        - cargo audit
+      script: ./hooks/pre-commit.d/cargo-audit
```
npmccallum:
```suggestion
      script: ./hooks/pre-commit.d/cargo-audit --force
```

hooks/pre-commit.d/sanity-checks
```diff
@@ -0,0 +1,2 @@
+if [ "`git rev-parse --show-cdup`" != "" ]; then cd `git rev-parse --show-cdup`; fi
+./hooks/pre-commit.d/cargo-fmt && ./hooks/pre-commit.d/cargo-test
```
npmccallum: Remove this whole file.

hooks/pre-commit.d/cargo-audit
```diff
@@ -0,0 +1 @@
+cargo audit
```
npmccallum: 1. Detect if we are running on Travis.
2. If on Travis, return success immediately unless the `--force` option is used.

hooks/pre-commit
```diff
@@ -0,0 +1,2 @@
+if [ "`git rev-parse --show-cdup`" != "" ]; then cd `git rev-parse --show-cdup`; fi
```
npmccallum: I don't understand what this check does.

hooks/pre-commit
```diff
@@ -0,0 +1,2 @@
...
+./hooks/pre-commit.d/sanity-checks && ./hooks/pre-commit.d/cargo-audit
```
npmccallum:
```suggestion
for f in $0.d/*; do $f; done
```

hooks/pre-commit
mbestavros: Sorry, I should have left a comment inline. All it does is change directories to the root of the Git working tree.

This is necessary because the pre-commit hook runs in whatever directory the user is committing from. So, in order to be able to refer to individual script files (the new ones we're adding in `hooks/pre-commit.d`), we need to be able to assume they're in the root of the repo. (Unless I'm missing something obvious.)

npmccallum: I **think** my use of `$0` below will solve that problem. Try it out.

mbestavros: You're right, it does. Good to know!

hooks/pre-commit.d/cargo-audit
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+if [[ ! -z $TRAVIS && ! $1 ]]; then
```
npmccallum:
```suggestion
[[ -z "$TRAVIS" || "$1" == "--force" ]] && cargo audit
```

mbestavros: Done. _Someday_ I'll make an elegant shell script on the first try...

mbestavros: @npmccallum Hm... There is one case where both conditions are "false" (where it's running on Travis, so `TRAVIS` exists, and `--force` is not enabled) and I think that's causing a failure.

I've amended the statement to look for the one case we want to exclude. It's similar to what I had before, but certainly less verbose -- and it works again.

hooks/pre-commit.d/cargo-audit
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+ [[ ! -z "$TRAVIS" && ! "$1" == "--force" ]] && exit 0
+ cargo audit
```
npmccallum: This file appears to have stray spaces at the beginning of the lines.

mbestavros: D'oh. Sorry -- fixed.

mbestavros:
OK... amended commit message and added a longer explanation of what we're doing here.

hooks/pre-commit.d/cargo-audit
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+[[ ! -z "$TRAVIS" && ! "$1" == "--force" ]] && exit 0
```
npmccallum:
```suggestion
[[ -n "$TRAVIS" && "$1" != "--force" ]] && exit 0
```

mbestavros:
@npmccallum Good changes, added.

mbestavros:
Rebased.

mbestavros:
@npmccallum Done. Commit message updated too.

(Though my fork's branch is now named wrong... :stuck_out_tongue: )

PR #152, opened 2020-01-22
add crt0stack crate

haraldh:


haraldh:
made it #[no_std] by default

npmccallum:
This is a resource for anyone reviewing this crate to understand what it is attempting to do.

https://lwn.net/Articles/631631/

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+use stack::Stack;
+
+/// The error type which is returned from serializing into a byte array.
+///
+/// This type does not support transmission of an error other than that an error
+/// occurred. Any extra information must be arranged to be transmitted through
+/// some other means.
+///
+/// An important thing to remember is that the type `Error` should not be
+/// confused with [`std::io::Error`] or [`std::error::Error`], which you may also
```
steveeJ: Would it make sense to diverge from the idiom and give it a more meaningful name?

crt0stack/examples/example.rs
```diff
@@ -0,0 +1,56 @@
...
+fn main() {
+    const ELF64_HDR_SIZE: u64 = 0x40;
+    const ELF64_PHDR_SIZE: u64 = 56;
+    let load_addr: u64 = 0x40_0000; // First elf PT_LOAD
+    let num_elf_program_sections: usize = 4;
+
+    let mut crt0sp = Crt0Stack::new();
+    crt0sp.envp.push("LANG=C".to_string());
+    crt0sp.argv.push("/init".to_string());
+    crt0sp.add_auxv_entry(AuxvEntry::ExecFilename("/init".to_string()));
```
steveeJ: The code block initializing the `Crt0Stack` looks like it would be a good fit for the builder pattern.

crt0stack/src/lib.rs
haraldh: It's the same as `alloc::fmt::Error`
https://doc.rust-lang.org/nightly/src/core/fmt/mod.rs.html#70-96

We might want to extend that..

crt0stack/examples/example.rs
haraldh: Well, Crt0Stack is the builder filling in a &[u8]

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+            sp.push_slice(s.as_bytes());
+            str_ptr.push(sp.ptr());
+        }
+        str_ptr.push(core::ptr::null());
+        for s in self.argv.iter().rev() {
+            sp.push_l(0u8);
+            sp.push_slice(s.as_bytes());
+            str_ptr.push(sp.ptr());
+        }
+
```
connorkuehl: Do we need to pad to a 16B boundary here?

Referring to this component that follows the args on the diagram for the lwn article that Nathaniel linked:

```
           /   4fd4: 'two\0'                                     |       | |     <----+
     args |    4fd0: 'one\0'                                     |       | |    <---+ |
           \_  4fcb: 'zero\0'                                    |       | |   <--+ | |
               3020: random gap padded to 16B boundary           |       | |      | | |
    - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -|       | |      | | |
               3019: 'x86_64\0'                        <-+       |       | |      | | 
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        match self.random {
+            None => {}
+            Some(x) => {
+                sp.align_down(core::mem::size_of::<AuxvType>());
+                sp.push_slice(&x);
+                rand_ptr = Some(sp.ptr());
+            }
+        }
```
connorkuehl: If we're not going to panic or anything when these are `None`, I think these can be compacted down with the [`if let` mechanism](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html), like this:

```diff
diff --git a/crt0stack/src/lib.rs b/crt0stack/src/lib.rs
index 5fbd02b..5bf9d99 100644
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -224,12 +224,9 @@ impl Crt0Stack {
         sp.push_l(0 as AuxvType);
         sp.push_l(AT_NULL);
 
-        match rand_ptr {
-            None => {}
-            Some(p) => {
+        if let Some(p) = rand_ptr {
                 sp.push_l(p as AuxvType);
                 sp.push_l(AT_RANDOM);
-            }
         }
 
         match platform_ptr {
```

npmccallum:
@haraldh In the SGX shim, we are not likely to have a heap. Is it possible to implement this such that it requires only the stack?

npmccallum:
I made a little experiment with a type that is stack allocated. This does work (verified with godbolt.org), though it is slightly dirty.

```rust
// NO Copy, Clone
pub struct AuxVec([u8; 4096]);

impl AuxVec {
    pub fn new() -> Self {
        Self([0u8; 4096])
    }

    /// Places the AuxVec onto the stack and calls the main function
    pub fn run(self, f: extern "C" fn() -> !) -> ! {
        // Internally, we need a little madness to keep the compiler from optimizing
        // out our stack allocation. We cast the function pointer which takes no
        // arguments into a function pointer which takes one argument (a pointer to
        // the AuxVec). Then we allocate the AuxVec on the stack and call the new
        // function pointer with a reference to the AuxVec on the stack.
        //
        // This only works on x86_64 (our target platform). Basically, Rust believes
        // it cannot optimize out the AuxVec allocation because the function pointer
        // "needs" a reference to it. But in reality the function ignores it. The
        // only invariant that I can tell that we have violated is that we are
        // leaking the reference to the AuxVec to the entry function in the %rdi
        // register. But this pointer is valid memory anyway. And the compiler for
        // the entry function presumes that %rdi contains invalid data anyway.
        //
        // So I *think* we're okay...
        let f: extern "C" fn(*const AuxVec) -> ! = unsafe { std::mem::transmute(f) };
        let av = self;
        f(&av)
    }
}
```

haraldh:
> I made a little experiment with a type that is stack allocated. This does work (verified with godbolt.org), though it is slightly dirty.
> 
> ```rust
> // NO Copy, Clone
> pub struct AuxVec([u8; 4096]);
> 
> impl AuxVec {
>     pub fn new() -> Self {
>         Self([0u8; 4096])
>     }
> 
>     /// Places the AuxVec onto the stack and calls the main function
>     pub fn run(self, f: extern "C" fn() -> !) -> ! {
>         // Internally, we need a little madness to keep the compiler from optimizing
>         // out our stack allocation. We cast the function pointer which takes no
>         // arguments into a function pointer which takes one argument (a pointer to
>         // the AuxVec). Then we allocate the AuxVec on the stack and call the new
>         // function pointer with a reference to the AuxVec on the stack.
>         //
>         // This only works on register ABIs (including x86_64). Basically, Rust believes
>         // it cannot optimize out the AuxVec allocation because the function pointer
>         // "needs" a reference to it. But in reality the function ignores it. The
>         // only invariant that I can tell that we have violated is that we are
>         // leaking the reference to the AuxVec to the entry function in the %rdi
>         // register. But this pointer is valid memory anyway. And the compiler for
>         // the entry function presumes that %rdi contains invalid data anyway.
>         //
>         // So I *think* we're okay... But this horribly breaks on any ABI which passes
>         // the first argument on the stack.
>         let f: extern "C" fn(*const AuxVec) -> ! = unsafe { std::mem::transmute(f) };
>         let av = self;
>         f(&av)
>     }
> }
> ```

Maybe `core::mem::ManuallyDrop` helps you also.

crt0stack/examples/example.rs
haraldh: Ok, implemented the owning builder pattern

crt0stack/src/lib.rs
haraldh: added padding

crt0stack/src/lib.rs
haraldh: done

haraldh:
Now without `alloc` and with the owning builder pattern. Please review again.

haraldh:
meh... the test failure is unrelated to my crate!

haraldh:
@npmccallum see https://github.com/enarx/enarx/issues/183

haraldh:
> I made a little experiment with a type that is stack allocated. This does work (verified with godbolt.org), though it is slightly dirty.
> 
> ```rust
> // NO Copy, Clone
> pub struct AuxVec([u8; 4096]);
> 
> impl AuxVec {
>     pub fn new() -> Self {
>         Self([0u8; 4096])
>     }
> 
>     /// Places the AuxVec onto the stack and calls the main function
>     pub fn run(self, f: extern "C" fn() -> !) -> ! {
>         // Internally, we need a little madness to keep the compiler from optimizing
>         // out our stack allocation. We cast the function pointer which takes no
>         // arguments into a function pointer which takes one argument (a pointer to
>         // the AuxVec). Then we allocate the AuxVec on the stack and call the new
>         // function pointer with a reference to the AuxVec on the stack.
>         //
>         // This only works on register ABIs (including x86_64). Basically, Rust believes
>         // it cannot optimize out the AuxVec allocation because the function pointer
>         // "needs" a reference to it. But in reality the function ignores it. The
>         // only invariant that I can tell that we have violated is that we are
>         // leaking the reference to the AuxVec to the entry function in the %rdi
>         // register. But this pointer is valid memory anyway. And the compiler for
>         // the entry function presumes that %rdi contains invalid data anyway.
>         //
>         // So I *think* we're okay... But this horribly breaks on any ABI which passes
>         // the first argument on the stack.
>         let f: extern "C" fn(*const AuxVec) -> ! = unsafe { std::mem::transmute(f) };
>         let av = self;
>         f(&av)
>     }
> }
> ```

@npmccallum for what do you need this?

npmccallum:
@haraldh Please rebase to master to fix the build.

haraldh:
> @haraldh Please rebase to master to fix the build.

done

haraldh:
> > @haraldh Please rebase to master to fix the build.
> 
> done

damn ... I love merge commits

haraldh:
rebased again

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+        let mut sp = Stack::new(slice);
+        let mut str_ptr: [*const u8; SIZE] = [core::ptr::null(); SIZE];
+        let mut str_len: usize = 0;
+        let mut rand_ptr: Option<*const u8> = None;
+        let mut platform_ptr: Option<*const u8> = None;
+        let mut base_platform_ptr: Option<*const u8> = None;
+        let mut exec_fn_ptr: Option<*const u8> = None;
+
+        str_ptr[str_len] = core::ptr::null();
+        str_len += 1;
```
connorkuehl: Would this change be equivalent and let us accomplish it when the variable is initialized since I think the way `str_ptr` is declared ensures it's populated with `core::ptr::null()` regardless.

```diff
diff --git a/crt0stack/src/lib.rs b/crt0stack/src/lib.rs
index 15398ed..564bb84 100644
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -240,15 +240,12 @@ impl<'a> Crt0Stack<'a> {
     pub fn serialize(self, slice: &mut [u8]) -> Result<usize, Error> {
         let mut sp = Stack::new(slice);
         let mut str_ptr: [*const u8; SIZE] = [core::ptr::null(); SIZE];
-        let mut str_len: usize = 0;
+        let mut str_len: usize = 1;
         let mut rand_ptr: Option<*const u8> = None;
         let mut platform_ptr: Option<*const u8> = None;
         let mut base_platform_ptr: Option<*const u8> = None;
         let mut exec_fn_ptr: Option<*const u8> = None;
 
-        str_ptr[str_len] = core::ptr::null();
-        str_len += 1;
-
         for s in self.envp.iter() {
             if let Some(s) = s {
                 sp.push_l(0u8);
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+            }
+        }
+
+        match exec_fn_ptr {
+            None => {}
+            Some(p) => {
+                sp.push_l::<AuxvType>(p as _);
+                sp.push_l::<AuxvKey>(AT_EXECFN);
+            }
+        }
```
connorkuehl: Just thinking out loud: I wonder if we should be panicking if any of these fields are missing. I don't think it has to be super elaborate, we can just [unwrap](https://doc.rust-lang.org/core/option/enum.Option.html#method.unwrap) the options here as it will panic for us if they are `None`. Like this:

```diff
-        match rand_ptr {
-            None => {}
-            Some(p) => {
-                sp.push_l::<AuxvType>(p as _);
-                sp.push_l::<AuxvKey>(AT_RANDOM);
-            }
-        }
+        sp.push_l::<AuxvType>(rand_ptr.unwrap() as _);
+        sp.push_l::<AuxvKey>(AT_RANDOM);
```

What do you think?

If we decide to not just unwrap, then I think the following can be adapted to `if let` to shrink the code a little bit.

```diff
diff --git a/crt0stack/src/lib.rs b/crt0stack/src/lib.rs
index 15398ed..e324e56 100644
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -301,37 +301,25 @@ impl<'a> Crt0Stack<'a> {
         sp.push_l::<AuxvType>(0 as AuxvType);
         sp.push_l::<AuxvKey>(AT_NULL);
 
-        match rand_ptr {
-            None => {}
-            Some(p) => {
-                sp.push_l::<AuxvType>(p as _);
-                sp.push_l::<AuxvKey>(AT_RANDOM);
-            }
-        }
+        if let Some(p) = rand_ptr {
+            sp.push_l::<AuxvType>(p as _);
+            sp.push_l::<AuxvKey>(AT_RANDOM);
+        };
 
-        match platform_ptr {
-            None => {}
-            Some(p) => {
-                sp.push_l::<AuxvType>(p as _);
-                sp.push_l::<AuxvKey>(AT_PLATFORM);
-            }
-        }
+        if let Some(p) = platform_ptr {
+            sp.push_l::<AuxvType>(p as _);
+            sp.push_l::<AuxvKey>(AT_PLATFORM);
+        };
 
-        match base_platform_ptr {
-            None => {}
-            Some(p) => {
-                sp.push_l::<AuxvType>(p as _);
-                sp.push_l::<AuxvKey>(AT_BASE_PLATFORM);
-            }
-        }
+        if let Some(p) = base_platform_ptr {
+            sp.push_l::<AuxvType>(p as _);
+            sp.push_l::<AuxvKey>(AT_BASE_PLATFORM);
+        };
 
-        match exec_fn_ptr {
-            None => {}
-            Some(p) => {
-                sp.push_l::<AuxvType>(p as _);
-                sp.push_l::<AuxvKey>(AT_EXECFN);
-            }
-        }
+        if let Some(p) = exec_fn_ptr {
+            sp.push_l::<AuxvType>(p as _);
+            sp.push_l::<AuxvKey>(AT_EXECFN);
+        };
 
         for s in self.auxv.iter() {
             if s.key == AT_NULL {
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+        assert_eq!(s, "x86_64");
+
+        let ap: AuxvPair = AuxvPair {
+            key: prep_stack.pop_l(),
+            value: prep_stack.pop_l(),
+        };
+        assert_eq!(ap.key, AT_NULL);
+        assert_eq!(ap.value, 0);
+    }
+}
```
connorkuehl: Nit:

```diff
diff --git a/crt0stack/src/lib.rs b/crt0stack/src/lib.rs
index 15398ed..57339b0 100644
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -449,10 +449,10 @@ mod tests {
     fn sp_head() {
         use std::ffi::CStr;
 
-        let prog = "/init".to_string();
+        let prog = "/init";
 
         let auxv = [
-            AuxvEntry::ExecFilename(prog.as_str()),
+            AuxvEntry::ExecFilename(prog),
             AuxvEntry::Platform("x86_64"),
             AuxvEntry::Uid(1000),
             AuxvEntry::Gid(1000),
@@ -460,7 +460,7 @@ mod tests {
 
         let mut stack = [1u8; 256];
         let spindex = Crt0Stack::new()
-            .add_arg(prog.as_str())
+            .add_arg(prog)
             .add_env("HOME=/root")
             .add_aux_entries(&auxv)
             .serialize(&mut stack)
@@ -476,7 +476,7 @@ mod tests {
         let arg: *const std::os::raw::c_char = prep_stack.pop_l();
         let cstr = unsafe { CStr::from_ptr(arg) };
         let s = cstr.to_string_lossy();
-        assert_eq!(s, prog.as_str());
+        assert_eq!(s, prog);
         let arg: *const std::os::raw::c_char = prep_stack.pop_l();
         assert_eq!(arg, core::ptr::null());
 
@@ -508,7 +508,7 @@ mod tests {
         assert_eq!(ap.key, AT_EXECFN);
         let cstr = unsafe { CStr::from_ptr(ap.value as _) };
         let s = cstr.to_string_lossy();
-        assert_eq!(s, prog.as_str());
+        assert_eq!(s, prog);
 
         let ap: AuxvPair = AuxvPair {
             key: prep_stack.pop_l(),
```

npmccallum:
> @npmccallum for what do you need this?

It was mostly just me thinking out loud. In SGX, the shim shares a stack with the static binary payload. So we have to place the crt0stack on **our** stack and then call the function.

npmccallum:
@haraldh I've been struggling over what I didn't like about this API. And I've finally realized what it is. Basically, this API attempts to manage a bunch of state and then serializes that state. But I suspect, this is unnecessary. What about an API like this?

```rust
pub struct OutOfSpace;
pub enum AuxVal {}

/// Creates a new crt0stack.
pub fn new(
    stack: &mut [u8],
    arg: impl IntoIterator<Item=&str>,
    env: impl IntoIterator<Item=(&str, &str)>,
    aux: impl IntoIterator<Item=AuxVal>,
) -> Result<(), OutOfSpace> {}


/// Parses the arguments from a crt0stack
pub fn arg(&[u8]) -> impl Iterator<Item=&str> {}

/// Parses the auxiliary vector from a crt0stack
pub fn aux(&[u8]) -> impl Iterator<Item=AuxVal> {}

/// Parses the environment variables from a crt0stack
pub fn env(&[u8]) -> impl Iterator<Item=(&str, &str)> {}
```

This lets the consumer of `crt0stack` build up state however they like (and using whatever facilities they have). Then, crt0stack just handles serialization and deserialization. Thoughts?

npmccallum:
Also, the stack contains pointers to within the stack. We might need to use `core::pin::Pin` to enforce that the stack doesn't move anywhere. Either that or we might need to make these functions unsafe considering they will attempt to read/write pointers to a movable buffer.

crt0stack/examples/example.rs
```diff
@@ -0,0 +1,59 @@
+use crt0stack::{AuxvEntry, Crt0Stack};
+
+fn main() {
+    const ELF64_HDR_SIZE: u64 = 0x40;
+    const ELF64_PHDR_SIZE: u64 = 56;
+    let load_addr: u64 = 0x40_0000; // First elf PT_LOAD
+    let num_elf_program_sections: usize = 4;
+
+    let prog = "/init".to_string();
```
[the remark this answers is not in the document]
haraldh: weil it was an example with String by purpose

crt0stack/src/lib.rs
haraldh: yes, but needs a comment

crt0stack/src/lib.rs
haraldh: panic!() should be appropriate

haraldh:
Here we go again..

crt0stack/src/auxv.rs
```diff
@@ -0,0 +1,148 @@
...
+
+//! Types and Constants to create an ELF AUXV
+
+/// The basic AuxvType able to hold a pointer
+pub type AuxvType = usize;
+
+/// AuxvEntry to be used with `Crt0Stack::add_auxv_entry()`
+pub enum AuxvEntry<'a> {
+    /// file descriptor of program
+    ExecFd(u64),
```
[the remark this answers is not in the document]
haraldh: I thought we are not going to support 32bit systems.

haraldh:
> In the process of thinking through how to improve the interior `Stack` type, I hit upon, I think a better external API that allows us to build up arguments dynamically. It also makes the `Stack` type redundant.
> ## Assumptions
> 
> The following proposal has the following assumptions, gleaned from [this document](https://lwn.net/Articles/631631/).
> 
>     1. The layout of the stack (from top to bottom) is data, random padding, items.
> 
>     2. The data section contains serialized information that can be referred to. It therefore, cannot be moved. It also cannot contain pointers. The ordering of the individual units of information in the data section is not significant.
> 
>     3. The items section contains serialized information or pointers to data items. However, its individual items cannot be referred to. As a result of this, the individual items can be moved so long as the relative ordering to other items is retained.
> 
>     4. The items section can be broken down into three subsections: `aux`, `env` and `arg`, which must appear in this order.
> 
> 
> ## Proposal
> 
> Something like this. The actual interfaces are just an idea, feel free to counter-propose. But I think this makes an even better public API than the single `serialize()` method. The only error should be if the provided buffer isn't large enough to serialize the data/items.
> 
> ```rust
> pub struct OutOfSpace;
> type Result<T> = Result<T, OutOfSpace>;
> 
> // Internal use only
> trait Serializable {
>     fn from(src: &[u8]) -> Result<(usize, Self)>;
>     fn into(&self, dst: &mut [u8]) -> Result<usize>;
> }
> 
> impl Serializable for str {}
> impl<T: Sized + Copy> Serializable for T {}
> impl<T: Sized + Copy> Serializable for [T] {}
> 
> // Externally unconstructable state markers
> pub struct Aux(());
> pub struct Env(());
> pub struct Arg(());
> 
> // Retains the immutability and immovability of the stack buffer
> pub struct Handle<'a>(&mut [u8]);
> 
> // Returns a reference to the top of the stack
> impl<'a> AsRef<()> for Handle<'a> {}
> 
> pub struct Builder<'a, T> {
>   stack: &mut [u8],
>   data: usize; // Index to the bottom of the data section
>   items: usize; // Index to the top of the items section
>   state: T,
> }
> 
> impl<'a, T> Builder<'a, T> {
>     // Private
>     // Serializes the input and saves it in the data section.
>     // Returns a reference (pointer?) to the serialized input within the data section.
>     fn save(&mut self, impl Serializable) -> Result<&()> { ... }
> }
> 
> impl<'a> Builder<'a, Aux> {
>   pub fn new(stack: &'a mut [u8]) -> Self {}
> 
>   // Adds the auxv::Entry to the items section,
>   // saving referred data to the data section as needed.
>   // Uses `copy_within()` to shift the items section.
>   pub fn push(&mut self, entry: auxv::Entry) -> Result<()> { ... }
> 
>   pub fn next(self) -> Result<Builder<'a, Env>> { ... }
> }
> 
> impl<'a> Builder<'a, Env> {
>   // Saves the env string to the data section and creates
>   // a reference to it in the items section.
>   // Uses `copy_within()` to shift the items section.
>   pub fn push(&mut self, env: &str) -> Result<()> { ... }
> 
>   pub fn next(self) -> Result<Builder<'a, Arg>> { ... }
> }
> 
> impl<'a> Builder<'a, Arg> {
>   // Saves the arg string to the data section and creates
>   // a reference to it in the items section.
>   // Uses `copy_within()` to shift the items section.
>   pub fn push(&mut self, env: &str) -> Result<()> { ... }
> 
>   pub fn done(self) -> Result<Handle> { ... }
> }
> ```
> 
> ## Discussion
> 
> While my suggestion to use `copy_within()` isn't my favorite thing, it does alleviate several problems in the existing code. It also means we don't have to pre-calculate the top of the items section.
> 
> The size of the random gap in the `crt0stack` can be controlled externally by passing in a randomly sized buffer.
> 
> Everything should be implementable without `std` or `alloc`.
> ## Safety
> 
> Since the written stack buffer will contain self-referential pointers, the stack buffer provided as input needs to be both immutable and immovable. We do this by taking a mutable reference to the stack which prevents all external actors from modifying or moving it. When we are done serializing the stack, we return the `Handle` type which retains the mutable reference. This ensures no one else can get a mutable reference or move the stack. The stack pointer is obtained from the `Handle` object, from which it shares a lifetime. This means that our code is safe until the `Handle` object is destroyed.

at what point of time can I say "over engineered" ? :stuck_out_tongue:

npmccallum:
> at what point of time can I say "over engineered" ? stuck_out_tongue

I certainly do toe the line. :stuck_out_tongue: 

My concerns are roughly as follows:

1. This crate is usable outside our project. We should provide an ergonomic, safe API.

2. This crate does fundamentally unsafe stuff and is thus a target for auditing. To aid in that auditing, it helps to divide each of the subtasks into a clearly defined chunk of code with well defined behavior and invariants.

3. Rust projects are routinely criticized for masking unsafe behavior. So we need to be as explicit as possible about this behavior.

I'm not above the charge of over-engineering. However, in the current case, my proposal is just a builder pattern and a three-phase state machine. These are well-worn primitives in idiomatic Rust.

I also understand this review has probably gone on too long. If you're feeling frustrated and would like to move on to something else, I can finish up the refactoring. I can point you to some pull requests I'm frustrated on if you'd like to trade. :)

crt0stack/src/auxv.rs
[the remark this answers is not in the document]
npmccallum: The enarx binary does not plan to support 32bit systems. Where such work is hard, we obviously won't do it. But I think one minor change to have support for 32bit isn't horrible.

`AsRawFd` requires `std`, so ignore that bad advice. Maybe just `usize` (or `isize`) is the right balance here? Alternatively, `Type` isn't a bad option either.

haraldh:
here we... first pass of review, before I pretty comment all of it

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        self.push_item(value)?;
+        Ok(())
+    }
+
+    pub fn done(mut self) -> Result<Handle<'a>> {
+        self.push_item(Key::default())?;
+        self.push_item(Type::default())?;
+
+        let start_ptr = {
+            // at the end, copy the items of the data section from the bottom to the top of the stack
```
haraldh: should be:
```
 // at the end, copy the items of the item section from the bottom to the top of the stack
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+// Internal use only
+trait Serializable {
+    fn from(src: &[u8]) -> Result<(usize, Self)>
+    where
+        Self: Sized;
+    fn into_buf(&self, dst: &mut [u8]) -> Result<usize>;
+}
+
+impl<T: Sized + Copy> Serializable for T {
```
npmccallum: As we discussed...

```suggestion
impl Serializable for usize {
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        let (_prefix, dst, suffix) = unsafe { dst.align_to_mut::<T>() };
+        if dst.is_empty() {
+            return Err(OutOfSpace);
+        }
+        dst[dst.len() - 1] = *self;
+        Ok(suffix.len() + core::mem::size_of::<T>())
+    }
+}
+
+impl<T: Sized + Copy> Serializable for [T] {
```
npmccallum: As we discussed...

```suggestion
impl<T: Sized + Copy> Serializable for &[T] {
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+    // a reference to it in the items section.
+    pub fn push(&mut self, env: &str) -> Result<()> {
+        self.save(0_u8)?; // c-str zero byte
+        let p = self.save_slice(env.as_bytes())?;
+        self.push_item(p)
+    }
+
+    pub fn next(mut self) -> Result<Builder<'a, Aux>> {
+        // last environ is NULL
+        self.push_item(0_usize)?;
```
npmccallum:
```suggestion
        self.push_item(0usize)?;
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        })
+    }
+}
+
+impl<'a> Builder<'a, Arg> {
+    pub fn new(stack: &'a mut [u8]) -> Self {
+        let len = stack.len();
+        Self {
+            stack,
+            data: len,
```
npmccallum:
```suggestion
            data: stack.len(),
```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        // Store argc at the beginning
+        let (prefix, dst, _suffix) = {
+            let start = 0;
+            let end = self.data;
+            unsafe { self.stack[start..end].align_to_mut::<Type>() }
+        };
+        if dst.is_empty() {
+            return Err(OutOfSpace);
+        }
+        dst[0] = ((self.items - prefix.len()) / core::mem::size_of::<Type>()) - 2; // argc
```
npmccallum: This block of code is really similar to the other methods. Can we deduplicate somehow?

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    pub struct Stack<'a> {
+        pub(crate) idx: usize,
+        pub(crate) stack: &'a mut [u8],
+    }
```
npmccallum: Do these need to be `pub`?

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    impl<'a> Stack<'a> {
+        #[inline(always)]
+        pub(crate) fn new(slice: &'a mut [u8]) -> Self {
+            Self {
+                idx: slice.len(),
+                stack: slice,
+            }
+        }
```
npmccallum:
```suggestion
        }

```

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+    state: PhantomData<T>,
+}
+
+impl<'a, T> Builder<'a, T> {
+    // Private
+    // Serializes the input and saves it in the data section.
+    // Returns a reference (pointer?) to the serialized input within the data section.
+    #[inline(always)]
+    fn save<S: Serializable>(&mut self, val: S) -> Result<&()> {
+        self.data -= Serializable::into_buf(&val, &mut self.stack[..self.data])?;
```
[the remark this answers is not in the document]
haraldh: no.. already written data is above, which we don't want to overwrite

crt0stack/src/auxv.rs
```diff
@@ -0,0 +1,154 @@
...
+    HWCap2(Type),
+
+    /// filename of program
+    ExecFilename(&'a str),
+}
+
+#[non_exhaustive]
+#[repr(usize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Key {
```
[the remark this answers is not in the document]
haraldh: well, it's fine to leave it there

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+//! Create the initial stack frame to start an ELF binary on Linux
+
+#![cfg_attr(not(test), no_std)]
+//#![deny(missing_docs)]
+#![deny(clippy::all)]
+
+pub mod auxv;
+pub use auxv::Entry;
+
+use crate::auxv::Key::*;
```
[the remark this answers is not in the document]
haraldh: ok

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+// Externally unconstructable state markers
+pub struct Aux(());
+pub struct Env(());
+pub struct Arg(());
+
+// Retains the immutability and immovability of the stack buffer
+pub struct Handle<'a>(&'a mut [u8], *const ());
+impl<'a> Handle<'a> {
+    // Returns a reference to the top of the prepared stack
+    pub unsafe fn get_start_ptr(&self) -> *const () {
```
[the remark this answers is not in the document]
haraldh: isn't it unsafe to use this pointer with the Handle dropped?

crt0stack/src/lib.rs
haraldh: Meh, I like this notation

crt0stack/src/lib.rs
haraldh: no.. leftover

crt0stack/src/lib.rs
haraldh: ?

npmccallum: Missing newline. :)

crt0stack/src/lib.rs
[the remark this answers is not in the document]
npmccallum: That's why I suggested `fn stack(&self) -> &()`. That way we communicate the lifetime. It is the responsibility of the caller to cast it appropriately and manage the lifetime of the `Handle`. That should probably be mentioned in a safety section in the function documentation.

crt0stack/src/lib.rs
haraldh: error[E0502]: cannot borrow `*stack` as immutable because it is also borrowed as mutable
   --> src/lib.rs:123:19
    |
118 |   impl<'a> Builder<'a, Arg> {
    |        -- lifetime `'a` defined here
...
121 | /         Self {
122 | |             stack,
    | |             ----- mutable borrow occurs here
123 | |             data: stack.len(),
    | |                   ^^^^^ immutable borrow occurs here
124 | |             items: core::mem::size_of::<usize>(),
125 | |             state: PhantomData,
126 | |         }
    | |_________- returning this value requires that `*stack` is borrowed for `'a`

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+impl<T: Sized + Copy> Serializable for [T] {
+    fn from(_src: &[u8]) -> Result<(usize, Self)>
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+
+    fn into_buf(&self, dst: &mut [u8]) -> Result<usize> {
```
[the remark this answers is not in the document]
haraldh: changed otherwise ... see updated PR

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        self.data -= Serializable::into_buf(&val, &mut self.stack[..self.data])?;
+        if self.data <= self.items {
+            Err(OutOfSpace)
+        } else {
+            Ok(unsafe { core::mem::transmute(&self.stack[self.data]) })
+        }
+    }
+
+    #[inline(always)]
+    fn save_slice<S: Sized + Copy>(&mut self, val: &[S]) -> Result<*const ()> {
```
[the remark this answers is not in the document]
haraldh: changed otherwise ... see updated PR

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+        self.data -= Serializable::into_buf(val, &mut self.stack[..self.data])?;
+        if self.data <= self.items {
+            Err(OutOfSpace)
+        } else {
+            Ok(unsafe { core::mem::transmute(&self.stack[self.data]) })
+        }
+    }
+
+    fn push_item<S: Sized + Copy>(&mut self, val: S) -> Result<()> {
```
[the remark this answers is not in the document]
haraldh: changed otherwise ... see updated PR

crt0stack/src/lib.rs
haraldh: not worth it

haraldh:
next round without full documentation

crt0stack/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+pub struct Env(());
+pub struct Arg(());
+
+// Retains the immutability and immovability of the stack buffer
+pub struct Handle<'a>(&'a mut [u8], usize);
+impl<'a> Handle<'a> {
+    // Returns a reference to the top of the prepared stack
+    #[inline]
+    pub fn start_ptr(&self) -> *const () {
+        &self.0[self.1] as *const u8 as *const ()
```
haraldh: or could be:
```rust
    pub fn start_ptr(&self) -> &'a () {
        unsafe { core::mem::transmute(&self.0[self.1]) }
    }
```

haraldh:
The checks fail, because `LICENSE` has a missing new line at the top. LICENSE was created by github, IIRC.

npmccallum:
> The checks fail, because `LICENSE` has a missing new line at the top. LICENSE was created by github, IIRC.

See #220

npmccallum:
@haraldh Can you rebase this for review?

haraldh:
done

haraldh:
and documented...

npmccallum:
:tada: :cake: :balloon:

Issue #6, opened 2019-12-04
Achieve static linking with rust and gcc


steveeJ:
There is [a thread on the rust users forum](https://users.rust-lang.org/t/statically-link-executable-with-glibc) which discusses this topic.
All examples I've seen online to link statically with rust rely on the musl libc.

haraldh:
hmm.. as mentioned in the thread:

> (to be clear: it is unsafe to ship these executables to systems that don't have exactly the same glibc, as glibc will be dynamically loading nss modules internally and there might be ABI conflicts, even silent crashy ones, although this likely only matters if you do DNS lookups)

haraldh:
Seems like this would require musl-gcc on Fedora/RHEL, too
There is no way around it, because openssl-sys compiles .c files

https://github.com/sfackler/rust-openssl/tree/master/openssl-sys/build

haraldh:
@npmccallum 

If we would choose glibc we would run into some problems:
```
$ cat hello.c
#include <stdio.h>
#include <stdlib.h>
#include <openssl/ssl.h>

int main(void)
{
        OpenSSL_add_ssl_algorithms();
        printf("Hello World!\n");
        return 0;
}
$ gcc -static hello.c -lssl -lcrypto -lz -ldl -Wl,--whole-archive -lpthread -Wl,--no-whole-archive
/usr/bin/ld: /usr/lib/gcc/x86_64-redhat-linux/9/../../../../lib64/libcrypto.a(fips.o): in function `verify_checksums':
(.text+0x516): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /usr/lib/gcc/x86_64-redhat-linux/9/../../../../lib64/libcrypto.a(b_addr.o): in function `BIO_lookup_ex':
(.text+0xd02): warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /usr/lib/gcc/x86_64-redhat-linux/9/../../../../lib64/libcrypto.a(b_sock.o): in function `BIO_gethostbyname':
(.text+0x75): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

* because of `pthread` we need special linking arguments because of the `weak` symbols.
`-Wl,--whole-archive -lpthread -Wl,--no-whole-archive`
[...]

PR #158, opened 2020-01-23
use musl as default build target via .cargo/config

haraldh:
for openssl, turn on the vendored feature, which propagated to
openssl-sys

Fixes https://github.com/enarx/enarx/issues/6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ rust:
   - stable
   - beta
   - nightly
+addons:
+  apt:
+    packages: libssl-dev musl-dev musl-tools
 matrix:
   allow_failures:
     - rust: nightly
@@ -34,4 +37,5 @@ matrix:
         - cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
 install:
   - rustup component add rustfmt
+  - rustup target add x86_64-unknown-linux-musl
 script: ./hooks/pre-push
--- a/iocuddle-sgx/Cargo.toml
+++ b/iocuddle-sgx/Cargo.toml
@@ -13,5 +13,5 @@ sgx-types = { path = "../sgx-types" }
 
 [dev-dependencies]
 sgx-crypto = { path = "../sgx-crypto" }
-openssl = "0.10.26"
+openssl = { version = "0.10.26", features = [ "vendored" ] }
 rstest = "0.5.2"
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[build]
+target = "x86_64-unknown-linux-musl"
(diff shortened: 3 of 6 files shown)


haraldh:
Seems like this would require musl-gcc on Fedora/RHEL, too
There is no way around it, because openssl-sys compiles .c files

https://github.com/sfackler/rust-openssl/tree/master/openssl-sys/build

.travis.yml
```diff
@@ -35,6 +38,8 @@ matrix:
         - cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
 install:
   - rustup component add rustfmt
+  - rustup target add x86_64-unknown-linux-musl
 script:
+  - export PKG_CONFIG_ALLOW_CROSS=1
```
npmccallum: Can you document in the `README.md` the steps required to set up the development environment? This probably needs to include which packages need to be installed, which targets to add and which environment variables to set.

I'm not particularly excited about having to set an environment variable. This means integrations like rls won't work out of the box. Is there any way we can do without this?

iocuddle-sgx/Cargo.toml
```diff
@@ -13,5 +13,5 @@ sgx-types = { path = "../sgx-types" }
 
 [dev-dependencies]
 sgx-crypto = { path = "../sgx-crypto" }
-openssl = "0.10.26"
+openssl = { version = "0.10.26", features = [ "vendored" ] }
```
npmccallum: I'd really like to avoid this if at all possible. Long term, we'd love to use the `openssl-static` build from RHEL. What stands in our way from doing that today?

haraldh: We can't link the static libs generated by gcc with glibc include headers with musl libc. So, the openssl-static on fedora doesn't help here. It would have to be recompiled with musl-gcc.

.travis.yml
haraldh: Difficult on fedora without musl-libc. There is https://copr.fedorainfracloud.org/coprs/taocris/musl/

haraldh: ok, works even without that env variable.. removed

haraldh:
Added instruction for development environment in README.md

iocuddle-sgx/Cargo.toml
npmccallum: Understood. So we have a longer term goal of static linking on glibc.

haraldh: > Understood. So we have a longer term goal of static linking on glibc.

see https://github.com/enarx/enarx/issues/6#issuecomment-578021892

npmccallum: Are we able to inject the vendored option only when building inside our workspace? I'm asking because it is possible for someone to use one of our crates as a dependency in a normal context. I don't want to force vendored on them.

npmccallum: I did some research on this and I think the answer is no. :(

PR #162, opened 2020-01-24
Major licensing reorganization + license testing

### LINKED ISSUE #136: Ensure all files have up-to-date Apache license headers
With the new year and all the new project files that come with it, we should ensure everything contains proper license headers.

mbestavros:
This PR represents a significant reorganization of our licensing strategy. It accomplishes a number of things (each of which is roughly in its own commit):

- Replaces all Apache license headers with [SPDX ID headers](https://spdx.org/ids-how) (and adds them where there was nothing before);
- Adds `license` fields to several `Cargo.toml` files that were missing them;
- Adds full Apache 2.0 license files to crates that were missing them;
- Implements tests to ensure all this licensing remains consistent going forward.

Resolves #136.


hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+for cargo_dir in $(find . -path ./target -prune -o -type f -name Cargo.toml -printf '%h\n'); do
+    LICENSE=$cargo_dir/LICENSE
+    if [ ! -f "$LICENSE" ]; then
+    echo "$cargo_dir does not have an Apache 2.0 license." && exit 1
+fi
```
npmccallum: Indent is wrong. Also this check tests for the existence of a license file but the error states the specific license even though this was not validated.

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+for cargo_dir in $(find . -path ./target -prune -o -type f -name Cargo.toml -printf '%h\n'); do
```
npmccallum: I think this is probably good enough and is easier to grok:

```suggestion
top=`git rev-parse --show-toplevel`
for cargo_dir in $(find "$top" -name Cargo.toml | sed 's|Cargo.toml|LICENSE|'); do
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,7 @@
...
+    LICENSE=$cargo_dir/LICENSE
```
npmccallum: This line isn't needed with my above find invocation.

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
+files_without_license=$(find . -path ./target -prune -o -type f -name "*.rs" -exec grep -L "// SPDX-License-Identifier: Apache-2.0" {} \;)
+[ -z "$files_without_license" ] && exit 0
+printf "SPDX license ID missing from the following .rs files:\n$files_without_license\n" && exit 1
```
npmccallum: 1. We want to enforce that the header is the first line of the file.
2. The error message is incorrect if the file contains a header but not an `Apache-2.0` license.
3. I think the find invocation can be cleaner.

hooks/pre-push.d/toml-license-field-test
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+files_without_license=$(find . -path ./target -prune -o -type f -name "Cargo.toml" -exec grep -l "\[package\]" {} \; | xargs grep -L "license = \"Apache-2.0\"")
+[ -z "$files_without_license" ] && exit 0
+printf "The following Cargo.toml files are missing Apache 2.0 license fields:\n$files_without_license\n" && exit 1
```
npmccallum: I think we need to properly parse the toml here. This test would pass for a toml that looks like this:

```toml
license = "Apache-2.0"
[package]
```

hooks/pre-push.d/crate-license-test
mbestavros: Do we want to verify that it is the specific license? Agreed otherwise, the message shouldn't state something untrue.

npmccallum: I think so. I wish we had a way to obviously warn if that is not the case. But then we'd have to live with those warnings forever or implement some license exception system. So I think we just test them as errors.

npmccallum: This also brings up the question of auditing the licenses of our dependencies to ensure we don't have any license incompatibility. Could you file an issue for that?

mbestavros:
I've made a few changes in my latest push:
- I've squashed the licensing changes into one commit instead of three, for a total of two new commits.
- I've made some changes to the `crate-license-test` based on Nathaniel's feedback and other things. 1) It now lists all noncompliant files intead of stopping at the first failure
2) It checks that the tested license matches the reference one found at [apache.org](https://www.apache.org/licenses/LICENSE-2.0.txt)

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+LICENSE_reference=$(curl https://www.apache.org/licenses/LICENSE-2.0.txt)
```
npmccallum:
```suggestion
apache=$(curl https://www.apache.org/licenses/LICENSE-2.0.txt)
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+SUCCESS=0
```
npmccallum:
```suggestion
status=0
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+for LICENSE_path in $(find "$top" -name Cargo.toml | sed 's|Cargo.toml|LICENSE|'); do
```
npmccallum:
```suggestion
for file in $(find "$top" -name Cargo.toml | sed 's|Cargo.toml|LICENSE|'); do
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+    if [ ! -f "$LICENSE_path" ]; then
```
npmccallum:
```suggestion
    if [ ! -f "$file" ]; then
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+        echo "$LICENSE_path does not exist. Please ensure an Apache 2.0 license is present." && SUCCESS=1
```
npmccallum:
```suggestion
        echo "$file does not exist. Please ensure an Apache 2.0 license is present."
        status=1
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+    else
+        LICENSE=`cat $LICENSE_path`
```
npmccallum: Drop this line.

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+        if [ "$LICENSE" != "$LICENSE_reference" ]; then
```
npmccallum:
```suggestion
        if [ "`cat $file`" != "$apache" ]; then
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+            printf "$LICENSE_path does not match the reference Apache 2.0 license found at\nhttps://www.apache.org/licenses/LICENSE-2.0.txt\n" && SUCCESS=1
```
npmccallum:
```suggestion
            echo "$file does not match the reference Apache 2.0 license found at:"
            echo "    https://www.apache.org/licenses/LICENSE-2.0.txt"
            status=1
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,15 @@
...
+        fi
+    fi
+done
+exit $SUCCESS
```
npmccallum:
```suggestion
exit $status
```

hooks/pre-push.d/rs-spdx-license-test
npmccallum: This still needs to be done.

hooks/pre-push.d/toml-license-field-test
npmccallum: This still needs to be done.

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+apache=$(curl https://www.apache.org/licenses/LICENSE-2.0.txt)
+status=0
+for file in $(find "$top" -name Cargo.toml | sed 's|Cargo.toml|LICENSE|'); do
+    if [ ! -f "$file" ]; then
+        echo "$file does not exist. Please ensure an Apache 2.0 license is present." && status=1
```
npmccallum:
```suggestion
        echo "$file does not exist. Please ensure an Apache 2.0 license is present."
        status=1
```

hooks/pre-push.d/crate-license-test
```diff
@@ -0,0 +1,16 @@
...
+    else
+        if [ "`cat $file`" != "$apache" ]; then
+            echo "$file does not match the reference Apache 2.0 license found at:"
+            echo "https://www.apache.org/licenses/LICENSE-2.0.txt"
```
npmccallum:
```suggestion
            echo "    https://www.apache.org/licenses/LICENSE-2.0.txt"
```

mbestavros:
My latest push fixes the SPDX license test. It now uses `sed` and checks the SPDX ID is actually on the first line of the file.

In the interest of getting the license changes and working tests into master, I've removed the `cargo.toml` license testing from this PR. I'll probably rewrite it some other way that will allow us to read TOML directly.

mbestavros:
Follow-up: rebased on top of upstream and added licenses to `enarx-keepmgr` -- which was missing them.

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -path "$top"/target -prune -o -name "*.rs")
+do
```
npmccallum: Don't waste a line for `do`.

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -path "$top"/target -prune -o -name "*.rs")
```
npmccallum:
```suggestion
for file in $(find "$top" -type f -name "*.rs")
```

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,12 @@
...
+do
+    if [ -f "$file" ] && [ ! "$(sed -n '/^\/\/ SPDX-License-Identifier: Apache-2.0/p;q' "$file")" ]
+    then
```
npmccallum: Don't waste a line for `then`.

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -path "$top"/target -prune -o -name "*.rs")
+do
+    if [ -f "$file" ] && [ ! "$(sed -n '/^\/\/ SPDX-License-Identifier: Apache-2.0/p;q' "$file")" ]
```
npmccallum:
```suggestion
    if ! awk '/\/\/ SPDX-License-Identifier: Apache-2.0$/ { exit 0 } { exit 1 }' $file; then
```

https://en.wikipedia.org/wiki/AWK should help explain why this works. Basically, awk will try to match the first line against the pattern. If it succeeds, `exit 0` otherwise `exit 1` will be issued because no pattern matches everything.

npmccallum:
@mbestavros Also, please put `Related: ####` in the commit message for the issue this is related to.

hooks/pre-push.d/rs-spdx-license-test
mbestavros: That line excludes the locally-generated `target` directory, which includes a lot of generated `.rs` files without a license. The test fails if it's removed, since those generated files get tested for license headers they don't have.

npmccallum: @mbestavros Thanks, I misunderstood that part. Yours is obviously better.

npmccallum: @mbestavros You still probably want `-type f` that will remove the need for your `[ -f $file ]` check later.

mbestavros: @npmccallum That's, again, related... This `find` invocation returns the right `.rs` files, but it also returns... the `target` directory itself, for some reason. I've already tried specifying `-type f`, but it doesn't seem to do anything -- the directory is still returned as a result.

Checking for `[ -f "$file" ]` later works. With my implementation it was necessary, since `sed` would fail when trying to read the `target` directory. Although, having tried your suggested `awk` line instead of `sed`, the test works without the `[ -f "$file" ]` check -- but it produces an error message.

If we can't fix the `target` directory in the `find` invocation, I prefer having `[ -f "$file" ]` to make the output clean.

mbestavros:
@npmccallum pointed me towards [toml-cli](https://crates.io/crates/toml-cli) as a way to test `.toml` files with shell. I've rewritten and re-added the `.toml` license field test using it, as well as incorporated some of Nathaniel's changes from the last round of feedback.

This is, however, another tool we'll need to guide new users to install. We should update the Rust installation guide on the How to Contribute wiki page with the needed install command.

hooks/pre-push.d/rs-spdx-license-test
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -path "$top"/target -prune -o -name "*.rs"); do
```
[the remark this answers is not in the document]
mbestavros: Aha, very good to know. I've fixed it in the latest push.

mbestavros:
Once this is merged, I will document the need to install `toml-cli` as part of Rust environment setup. (Needed for `cargo.toml` testing)

haraldh:
Just wanted to mention `cargo-about` and `cargo-deny` . Might already do what we want and way more, if configured.
https://crates.io/crates/cargo-deny
https://crates.io/crates/cargo-allow

npmccallum:
> Just wanted to mention `cargo-about` and `cargo-deny` . Might already do what we want and way more, if configured.
> https://crates.io/crates/cargo-deny
> https://crates.io/crates/cargo-allow

`cargo-deny` is an excellent find!

However, the link to `cargo-allow` is broken. Did you mean something different?

mbestavros:
@haraldh @npmccallum Issue filed: https://github.com/enarx/enarx/issues/211

hooks/pre-push.d/toml-license-field-test
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -name Cargo.toml); do
+    if [ "$(toml get "$file" package.license)" != "\"Apache-2.0\"" ]; then
+        echo "$file does not include Apache-2.0 in its license field. Please ensure the field is present and includes an Apache license."
```
npmccallum: Wrap output.

mbestavros: Fixed with two `echo` calls in latest commit revision.

PR #163, opened 2020-01-24
Test for newlines at end of all *.rs and *.toml files

### LINKED ISSUE #160: Test for newline at the end of every .rs and .toml file
We want consistency here to avoid diffing problems later in git.

mbestavros:
Resolves #160.

### DIFF
--- a/hooks/pre-push.d/eof-newline-test
+++ b/hooks/pre-push.d/eof-newline-test
@@ -1,0 +1,10 @@
+#!/bin/bash
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -path "$top"/target -prune -o -type f -name "*.rs" -print -o -type f -name "*.toml" -print); do
+    if [ ! $(tail -c 1 "$file") = $'\n' ]; then
+        echo "$file does not have a newline at EOF. Please ensure it has one."
+        status=1
+    fi
+done
+exit $status


mbestavros:
I've rewritten the test to better follow the format we've established for our other tests. It also correctly tests all files now. Ready for review.

hooks/pre-push.d/eof-newline-test
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find . -path ./target -prune -o -type f -name "*.rs" -o -name "*.toml"); do
```
npmccallum:
```suggestion
for file in $(find . -path ./target -prune -o -type f -name "*.rs" -o -type f -name "*.toml"); do
```

I think we need `-type f` again after the or.

hooks/pre-push.d/eof-newline-test
```diff
@@ -0,0 +1,16 @@
...
+    if [ -f "$file" ] && [ ! -z $(tail -c 1 "$file") ]; then
```
npmccallum:
```suggestion
    if [ ! -z $(tail -c 1 "$file") ]; then
```

Also, this tests for any whitespace, not specifically newline.

hooks/pre-push.d/eof-newline-test
```diff
@@ -0,0 +1,16 @@
...
+        status=1
+    fi
+done
+exit $status
+
+
+
+#  -print0 | xargs -0 -L1 bash -c 'test "$(tail -c 1 "$0")" && echo "$0"')
+# [ -z "$files_without_eof" ] && exit 0
+# printf "The following files are missing a newline at EOF:\n$files_without_eof\n" && exit 1
```
npmccallum: Clean this up.

hooks/pre-push.d/eof-newline-test
mbestavros: Good catch. Fixed in latest push.

hooks/pre-push.d/eof-newline-test
mbestavros: Right. Fixed in latest push.

hooks/pre-push.d/eof-newline-test
mbestavros: D'oh, sorry. Fixed in latest push.

mbestavros:
I've addressed the first round of feedback. I've also incorporated `-print` in the `find` invocation to suppress printing the `target` directory, as suggested by @npmccallum in #162.

PR #167, opened 2020-01-27
Add Targetinfo struct to sgx-types

### LINKED ISSUE #170: Add TargetInfo type to sgx-types
Add the `TargetInfo` type as described in the Intel documentation.

lkatalin:
Fixes #170.

### DIFF
--- a/sgx-types/src/ti.rs
+++ b/sgx-types/src/ti.rs
@@ -1,0 +1,58 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Report Target Info (Section 38.16)
+//! The Target Info is used to identify the target enclave that will be able to cryptographically
+//! verify the REPORT structure returned by the EREPORT leaf. Must be 512-byte aligned.
+
+use super::{attr::Attributes, misc::MiscSelect, secs::Secs};
+
+/// Table 38-22
+#[derive(Debug)]
+#[repr(C, align(512))]
+pub struct TargetInfo {
+    /// MRENCLAVE of the target enclave.
+    pub mrenclave: [u8; 32],
+    /// Attributes of the target enclave.
+    pub attributes: Attributes,
+    reserved0: u32,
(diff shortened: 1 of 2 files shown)


sgx-types/src/ti.rs
```diff
@@ -0,0 +1,66 @@
...
+impl AsRef<[u8]> for TargetInfo {
+    fn as_ref(&self) -> &[u8] {
+        unsafe {
+            core::slice::from_raw_parts(
+                self as *const Self as *const u8,
+                core::mem::size_of_val(self),
+            )
+        }
+    }
+}
```
npmccallum: Don't `impl AsRef<[u8]>` if it isn't used.

lkatalin: I thought we might use it to send `Targetinfo` between enclaves; but I can remove it if it is unnecessary.

sgx-types/src/ti.rs
```diff
@@ -0,0 +1,55 @@
...
+#[repr(C, align(512))]
+pub struct TargetInfo {
+    /// MRENCLAVE of the target enclave.
+    pub mrenclave: [u8; 32],
+    /// Attributes of the target enclave.
+    pub attributes: Attributes,
+    reserved0: u32,
+    /// MiscSelect of the target enclave.
+    pub misc: MiscSelect,
+    reserved1: [u8; 57],
```
npmccallum:
```suggestion
    reserved1: [u64; 32],
    reserved2: [u64; 25],
```

We want to keep array sizes to a max of 32 to get the implicit traits on arrays.

sgx-types/src/ti.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+testaso! {
+    struct TargetInfo: 512, 512 => {
+        mrenclave: 0,
+        attributes: 32,
+        reserved0: 48,
+        misc: 52,
+        reserved1: 56
```
npmccallum: Add a test for `reserved2` here.

sgx-types/src/ti.rs
npmccallum: I think we will want proper serialization (i.e. serde) for that in the future.

sgx-types/src/ti.rs
```diff
@@ -0,0 +1,59 @@
...
+    reserved0: u32,
+    /// MiscSelect of the target enclave.
+    pub misc: MiscSelect,
+    reserved1: [u64; 32],
+    reserved2: [u64; 25],
+}
+
+impl TargetInfo {
+    /// Creates new TargetInfo from an enclave's SECS.
+    pub const fn new(secs: Secs) -> Self {
```
npmccallum: Do we always create a `TargetInfo` from an `Secs`? I'm honestly not sure. Maybe it would be better to create it from the individual components and then `impl From<Secs> for TargetInfo`?

lkatalin: I am thinking this over. The components always have to be parts of the same `secs`, belonging to the target enclave; otherwise they won't refer to a valid enclave.

PR #168, opened 2020-01-27
Throw away the top-level workspace

npmccallum:
Cargo currently cannot handle our needs. It cannot build static binaries
with the standard entry point on glibc. Nor can it build cdylib or
static binaries with a custom entry point on musl. Nor can it build
different crates for different targets in the same workspace.

While we work to resolve these issues in upstream cargo, we need to
ditch the top-level workspace for now. We can revisit this at a later
time.

For more background, see:

  https://github.com/rust-lang/cargo/pull/7811
  https://github.com/rust-lang/cargo/pull/7804
  https://github.com/rust-lang/rfcs/pull/2735/files


npmccallum:
@mbestavros I'll wait for you to see this before I merge it. Please review and approve it when you're ready for me to merge it.

PR #174, opened 2020-01-28
Add fields to XSave struct

### LINKED ISSUE #171: Fill in XSave struct in sgx-types/ssa.rs
Add details to the XSave struct.

lkatalin:
Based on https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/encl_bootstrap.S#L83

Fixes #171 .

### DIFF
--- a/intel-types/src/lib.rs
+++ b/intel-types/src/lib.rs
@@ -9,7 +9,9 @@
 #![allow(clippy::identity_op)]
 #![deny(missing_docs)]
 
+use bitflags::bitflags;
 use core::{fmt::Debug, ops::BitAnd};
+use testing::testaso;
 
 /// Succinctly describes a masked type, e.g. masked Attributes or masked MiscSelect.
 /// A mask is applied to Attributes and MiscSelect structs in a Signature (SIGSTRUCT)
@@ -40,12 +42,313 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
     }
 }
 
-/// The size of XSAVE region in SSA is derived from the enclave’s support of the collection
-/// of processor extended states that would be managed by XSAVE. The enablement of those
-/// processor extended state components in conjunction with CPUID leaf 0DH information
-/// determines the XSAVE region size in SSA.
-///
-/// Section 38.9, Table 38-7
+/// This type represents an MMX register, used to perform operations on 64-bit packed integer data.
+/// It includes the 10-byte value as well as 6 bytes of padding.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct Mm(u128);
+
+/// This type represents an XMM 128-bit data register, used to operate on packed single-
+/// precision floating-point operands.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct Xmm(u128);
+
+bitflags! {
+    /// x87 Floating Point Unit (FPU) Control Word
+    /// Section 8.1.5
+    pub struct Fcw: u16 {
+        /// Invalid Operation
+        const INV_OP = 1 << 0;
+
+        /// Denormal Operand
+        const DENORM_OP = 1 << 1;
+
+        /// Zero Divide
+        const ZERO_DIV = 1 << 2;
+
+        /// Overflow
+        const OVERFLOW = 1 << 3;
+
+        /// Underflow
+        const UNDERFLOW = 1 << 4;
+
+        /// Precision
+        const PREC = 1 << 5;
+
+        /// Precision Control 0
+        const PREC_CTRL0 = 1 << 8;
+
+        /// Precision Control 1
+        const PREC_CTRL1 = 1 << 9;
+
+        /// Rounding Control 0
+        const ROUND_CTRL0 = 1 << 10;
+
+        /// Rounding Control 1
+        const ROUND_CTRL1 = 1 << 11;
+
+        /// Infinity Control
+        const INFINITY_CTRL = 1 << 12;
+    }
+}
+
+impl Default for Fcw {
+    /// The x87 state initial configuration for FCW masks all floating-point exceptions,
+    /// sets rounding to nearest, and sets the x87 FPU precision to 64 bits.
+    /// See Section 8.1.5.
+    fn default() -> Self {
+        Fcw::INV_OP
+            | Fcw::DENORM_OP
+            | Fcw::ZERO_DIV
+            | Fcw::OVERFLOW
+            | Fcw::UNDERFLOW
+            | Fcw::PREC
+            | Fcw::PREC_CTRL0
+            | Fcw::PREC_CTRL1
+    }
+}
+
+bitflags! {
+    /// 32-bit register providing status and control bits used in SIMD floating-point operations
+    pub struct MxCsr: u32 {
+        /// Invalid Operation Flag
+        const INV_OP = 1 << 0;
+
+        /// Denormal Flag
+        const DENORM = 1 << 1;
+
+        /// Divide by Zero Flag
+        const ZERO_DIV = 1 << 2;
+
+        /// Overflow Flag
+        const OVERFLOW = 1 << 3;
+
+        /// Underflow Flag
+        const UNDERFLOW = 1 << 4;
+
+        /// Precision Flag
+        const PREC = 1 << 5;
+
+        /// Denormals are Zeros
+        const DENORM_ARE_ZEROS = 1 << 6;
+
+        /// Invalid Operation Mask
+        const INV_OP_MASK = 1 << 7;
+
+        /// Denormal Operation Mask
+        const DENORM_MASK = 1 << 8;
+
+        /// Divide by Zero Mask
+        const ZERO_DIV_MASK = 1 << 9;
+
+        /// Overflow Mask
+        const OVERFLOW_MASK = 1 << 10;
+
+        /// Underflow Mask
+        const UNDERFLOW_MASK = 1 << 11;
+
+        /// Precision Mask
+        const PREC_MASK = 1 << 12;
+
+        /// Rounding Control 0
+        const ROUND_CTRL0 = 1 << 13;
+
+        /// Rounding Control 1
+        const ROUND_CTRL1 = 1 << 14;
+
+        /// Flush to Zero
+        const FLUSH_TO_ZERO = 1 << 15;
+    }
+}
+
+impl Default for MxCsr {
+    /// The initial state of MXCSR after power-up/reset or INIT; mask allows software
+    /// to identify any reserved bits in MXCSR (none are reserved here).
+    /// See Sections 11.6.4 and 11.6.6.
+    fn default() -> Self {
+        MxCsr::INV_OP_MASK
+            | MxCsr::DENORM_MASK
+            | MxCsr::ZERO_DIV_MASK
+            | MxCsr::OVERFLOW_MASK
+            | MxCsr::UNDERFLOW_MASK
+            | MxCsr::PREC_MASK
+    }
+}
+
+impl Default for Masked<MxCsr> {
+    fn default() -> Self {
+        Self {
+            data: MxCsr::default(),
+            mask: MxCsr::all(),
+        }
+    }
+}
+
+bitflags! {
+    /// Section 13.4.3
+    pub struct XcompBv: u64 {
+        /// Compacted form is used for the layout of the XSAVE EXtended Region
+        const COMPACT = 1 << 63;
+    }
+}
+
+impl Default for XcompBv {
+    /// XCOMP_BV[63] = 1, compaction mode for the XSave Extended Region. However, no state
+    /// components are included in the Extended Region as all other bits are 0. The size of
+    /// the XSaveXtd is therefore zero.
+    // TODO: Check this value, as the original appears to set bit 31 but claims to set bit 63
+    fn default() -> Self {
+        XcompBv::COMPACT
+    }
+}
+
+bitflags! {
+    /// The state component bitmap identifies the state components present in the XSAVE area.
+    /// Bits 62:10 are reserved. Bit 63 does not correspond to any state component.
+    /// Section 13.1
+    #[derive(Default)]
+    pub struct XstateBv: u64 {
+        /// x87 state (Section 13.5.1)
+        const X87 = 1 << 0;
+
+        /// SSE state (Section 13.5.2)
+        const SSE = 1 << 1;
+
+        /// AVX state (Section 13.5.3)
+        const AVX = 1 << 2;
+
+        /// MPX state: BND0-BND3 (BNDREGS state)
+        const BNDREGS = 1 << 3;
+
+        /// MPX state: BNDCFGU and BNDSTATUS (BNDCSR state)
+        const BNDCSR  = 1 << 4;
+
+        /// AVX-512 state: opmask state
+        const AVX512_OPMASK = 1 << 5;
+
+        /// AVX-512 state: ZMM_HI256 state
+        const AVX512_ZMM_HI256 = 1 << 6;
+
+        /// AVX-512 state: HI16_ZMM state
+        const AVX512_HI16_ZMM = 1 << 7;
+
+        /// Processor Trace MSRs
+        const PT = 1 << 8;
+
+        /// Protection key feature register (Section 13.5.7)
+        const PKRU = 1 << 9;
+    }
+}
+
+/// The legacy region of an XSAVE area comprises the 512 bytes starting at the area's base address.
+/// See Table 13-1. There is no alignment requirement.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct XSaveLegacy {
+    /// x87 Floating Point Unit (FPU) Control Word
+    pub fcw: Fcw,
+
+    /// x87 FPU Status Word
+    pub fsw: u16,
+
+    /// x87 FPU Tag Word
+    pub ftw: u8,
+
+    /// Reserved
+    pub reserved0: u8,
+
+    /// x87 FPU Opcode
+    pub fop: u16,
+
+    /// x87 FPU Instruction Pointer Offset
+    pub fip: u64,
+
+    /// x87 FPU Data Pointer Offset
+    pub fdp: u64,
+
+    /// 32-bit register providing status and control bits used in SIMD floating-point operations
+    pub mxcsr: Masked<MxCsr>,
+
+    /// Register used to perform operations on 64-bit packed integer data
+    pub mm: [Mm; 8],
+
+    /// 128-bit data register used to operate on packed single-precision floating-point operands
+    pub xmm: [Xmm; 16],
+
+    /// Padding: Size of XSaveLegacy must be 512.
+    pub padding0: [u64; 11],
+
+    /// Padding: Size of XSaveLegacy must be 512.
+    pub padding1: [u8; 7],
+}
+
+/// The XSAVE header of an XSAVE area comprises the 64 bytes starting at offset 512 from the
+/// area's base address. See Section 13.4.2. There is no alignment requirement.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct XSaveHeader {
+    /// State-component bitmap identifying the state components in the XSAVE area.
+    pub xstate_bv: XstateBv,
+
+    /// State-component bitmap indicating the format of the XSAVE extended region and whether
+    /// a component is in the XSAVE area.
+    pub xcomp_bv: XcompBv,
+
+    /// Reserved
+    pub reserved0: [u64; 6],
+}
+
+/// For our use case, the XSave Extended Region is in compacted format and currently holds nothing,
+/// as specifed in XCOMP_BV in
+/// https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/encl_bootstrap.S#L89.
+/// For more on this region, see Section 13.4.3. The alignment requirements are variable and specified in
+/// 13.4.3.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct XSaveExtend;
+
+/// For details on the fields included in XSave, see Section 13.4. Must be 64 byte aligned.
 #[derive(Debug)]
-#[repr(C, align(4096))]
-pub struct XSave([[u64; 32]; 16]);
+#[repr(C, align(64))]
+pub struct XSave {
+    /// Legacy region of the XSave area
+    pub legacy: XSaveLegacy,
+
+    /// XSave header
+    pub header: XSaveHeader,
+
+    /// XSave Extended Region (not used)
+    pub extend: XSaveExtend,
+}
+
+testaso! {
+    struct XSaveLegacy: 8, 512 => {
+        fcw: 0,
+        fsw: 2,
+        ftw: 4,
+        reserved0: 5,
+        fop: 6,
+        fip: 8,
+        fdp: 16,
+        mxcsr: 24,
+        mm: 32,
+        xmm: 160,
+        padding0: 416,
+        padding1: 504
+    }
+
+    struct XSaveHeader: 8, 64 => {
+        xstate_bv: 0,
+        xcomp_bv: 8,
+        reserved0: 16
+    }
+
+    struct XSaveExtend: 1, 0 => { }
+
+    struct XSave: 64, 576 => {
+        legacy: 0,
+        header: 512,
+        extend: 576
+    }
+}
--- a/intel-types/src/lib.rs
+++ b/intel-types/src/lib.rs
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Intel Documentation related to these types is available at the following link.
+//! Section references in further documentation refer to this document.
+//! https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf
+
+#![no_std]
+#![deny(clippy::all)]
+#![allow(clippy::identity_op)]
+#![deny(missing_docs)]
+
+use core::{fmt::Debug, ops::BitAnd};
+
+/// Succinctly describes a masked type, e.g. masked Attributes or masked MiscSelect.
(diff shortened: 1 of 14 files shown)


lkatalin:
@npmccallum I did my best to read the asm properly and turn it into a struct. For example, I read the line `.fill 1, 4, 0x037F # FCW` as `create a 4-byte field to hold FCW`. However, `FCW` itself is only 2 bytes, according to 8.1.5 in the [documentation](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf). In this case, I made the field 4 bytes in size anyway. Let me know if this was not the right way to do it.

npmccallum:
@lkatalin I would base this struct off of the section 13.4 in the same documentation. When he allocates 4 bytes for FCW, he's just taking a shortcut (I think). So I would split it up properly. But I would definitely follow the documentation.

lkatalin:
@npmccallum I changed the sizes of the `FCW` and `XCOMP_BV` fields according to the documentation.

sgx-types/src/ssa.rs
```diff
@@ -273,12 +273,39 @@ pub struct Miscellaneous {
...
+pub struct XSave {
+    fcw: u16,
+    reserved0: [u32; 5],
+    mxcsr: u32,
+    mxcsr_mask: u32,
+    reserved1: [[u32; 32]; 3],
+    reserved2: [u32; 27],
+    xcomp_bv: u64,
+    reserved3: [u32; 12],
+}
```
npmccallum: You appear to have filled in some of the XSave Legacy Region fields into the XSave struct. But:
1. You're missing many fields.
2. You've only covered the legacy region. You also need the XSAVE header and the extended area.
3. We're missing documentation (why isn't the docs error being thrown?).
4. The fields don't seem to correspond to the table 13-1.

lkatalin: @npmccallum The fields are missing because I only filled in the minimal set from https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/encl_bootstrap.S#L83, as suggested in #171. That's why it doesn't look like the table. 

However, I can add the rest of the fields from 13-1 now.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+#[repr(C)]
+pub struct XSaveLegacy {
+    fcw: u16,
+    fsw: u16,
+    ftw: u8,
+    reserved0: u8,
+    fop: u16,
+    fip31_0: u32,
+    fcs_or_fip47_32: u16,
+    fip63_48_or_reserved: u16,
```
npmccallum: Just do `fip: u64` here.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+    fdp31_0: u32,
+    fds_or_fdp47_32: u16,
+    fdp63_48_or_reserved: u16,
```
npmccallum: `fdp: u64,`

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+    mxcsr: u32,
+    mxcsr_mask: u32,
```
npmccallum: I think you can use the `Masked` type here.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+    mm0: [u8; 10],
+    mm1: [u8; 10],
+    mm2: [u8; 10],
+    mm3: [u8; 10],
+    mm4: [u8; 10],
+    mm5: [u8; 10],
+    mm6: [u8; 10],
+    mm7: [u8; 10],
```
npmccallum: I would make a `Mm` type and make this: `mm: [Mm; 8]`. Also, while the individual mm value is 10 bytes, the actual size of this segment is 16 bytes (10 + 6 padding).

You can read about MMX here: https://en.wikipedia.org/wiki/MMX_(instruction_set)

We might want to represent this (internally) as a `u128` with `From` implementations to `u64` that mask off the relevant bits. But we might just want to punt on the `From` implementations altogether.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+    xmm6: [u8; 16],
+    xmm7: [u8; 16],
+    xmm8: [u8; 16],
+    xmm9: [u8; 16],
+    xmm10: [u8; 16],
+    xmm11: [u8; 16],
+    xmm12: [u8; 16],
+    xmm13: [u8; 16],
+    xmm14: [u8; 16],
+    xmm15: [u8; 16],
```
npmccallum: `pub struct Xmm(u128);` then `xmm: [Xmm; 16],`

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+/// For more on this region, see 13.4.3 in
+/// https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf.
+#[derive(Debug)]
+#[repr(C)]
+pub struct XSaveXtd;
+
+/// This helper struct creates the correct amount of padding for XSave to be 4096 bytes
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct Padding([u8; Self::size()]);
```
npmccallum: Delete this helper type.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+
+impl Padding {
+    const fn size() -> usize {
```
npmccallum: Move `Padding::size()` to `StateSaveArea::padding()`.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +271,122 @@ pub struct Miscellaneous {
...
+
+/// When an AEX occurs while running in an enclave, the architectural state is saved
+/// in the thread’s current StateSaveArea (SSA Frame), which is pointed to by TCS.CSSA.
+///
+/// Section 38.9, Table 38-7
+//#[derive(Debug)]
+#[repr(C)]
+pub struct StateSaveArea {
+    xsave: XSave,
+    reserved: Padding,
```
npmccallum:
```suggestion
    reserved: [u8; Self::padding()],
```

sgx-types/src/ssa.rs
```diff
@@ -325,24 +403,8 @@ impl Footer {
...
-    /// Other information about the SSA
-    /// TODO: Be more specific about this
-    pub other: T,
-    /// Section 38.9, Table 38-7
-    pub footer: Footer,
-}
-
 testaso! {
-    struct Gpr: 8, 184 => {
+    struct Gpr: 8, 184 => { // TODO: the docs say this struct should be 176 bytes?
```
npmccallum: Honestly not sure. Please submit a separate patch for this.

sgx-types/src/ssa.rs
lkatalin: Makes sense; I just saw the padding this morning and added it in, but I'll make a type.

lkatalin: @npmccallum I've got `pub struct Mm(u128);` right now to encompass the padding and the values. Were you thinking the `From` would convert between the `u128` and *two* `u64`s to hold the 10 bytes of data?

npmccallum: I'd like to avoid the conversions for now. Basically, the size of the register is 80bit x87 floating point. However, I think the top 16 bits are always 1. So it is really a `u64` with some additional reserved space. But I'm not sure of the full implications of that, so I'd like to avoid making a conversion decision until we need to.

npmccallum:
@lkatalin Please rebase on the latest master to fix the build.

lkatalin:
@npmccallum Here is the new version. I had to list out default fields instead of using `..Default::default()` because apparently the latter gives me a recursion warning. Also, I had to remove the `Debug` trait from the `StateSaveArea` because the padding is too large. This I could change by adding a `Debug` implementation for the particular size of padding, if you think that is useful.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
-/// determines the XSAVE region size in SSA.
-///
-/// Section 38.9, Table 38-7
+/// This type represents an MMX register, used to perform operations on 64-bit packed integer data.
+/// It includes the 10-byte value as well as 6 bytes of padding.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct Mm(u128);
+
+/// This type represents an XMM 128-bit data register, used to operate on packed single-precision floating-point operands.
```
npmccallum: Wrap comment.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+/// The legacy region of an XSAVE area comprises the 512 bytes starting at the area’s base address.
+/// See Table 13-1 in
+/// https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf
 #[derive(Debug)]
-#[repr(C, align(4096))]
-pub struct XSave([[u64; 32]; 16]);
+#[repr(C)]
+pub struct XSaveLegacy {
+    /// x87 Floating Point Unit (FPU) Control Word
+    pub fcw: u16,
```
npmccallum: As a matter of style we want an empty line between documented fields.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
-///
-/// Section 38.9, Table 38-7
+impl Default for XSaveLegacy {
+    fn default() -> Self {
+        Self {
+            /// x87 state initial configuration for FCW. This masks all floating-point exceptions,
+            /// sets rounding to nearest, and sets the x87 FPU precision to 64 bits.
+            /// See Section 8.1.5 in
+            /// https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf
+            fcw: 0x037F,
```
npmccallum: Is this a bitfield? If so, we need a bitfield here rather than a `u16`.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+    padding1: [u8; 7],
+}
 
-/// The Footer combines the padding, Miscellaneous, and GPRSGX fields of the SSA Frame
-/// struct. It has no direct equivalent in the SGX documentation, but creates an SSA Frame
-/// in combination with the XSave struct.
-///
-/// Section 38.9, Table 38-7
+impl Default for XSaveLegacy {
+    fn default() -> Self {
```
npmccallum: It might be shorter to use `MaybeUninit::zeroed()` here and then fill in the fields you want.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+            ftw: Default::default(),
+            reserved0: Default::default(),
+            fop: Default::default(),
+            fip: Default::default(),
+            fdp: Default::default(),
+            /// Initial state of MXCSR after power-up/reset or INIT. The mask allows software
+            /// to identify any reserved bits in MXCSR (none are reserved here).
+            /// See Sections 11.6.4 and 11.6.6 in
+            /// https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf
+            mxcsr: Masked::<u32> {
```
npmccallum: Is mxcsr a bitfield? If so, we need it defined.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
         Self {
-            padding0: [[0; 32]; 15],
-            padding1: [0; 14],
-            misc,
-            gpr,
+            xstate_bv: Default::default(),
+            /// XCOMP_BV[63] = 1, compaction mode for the XSave Extended Region. However, no state
+            /// components are included in the Extended Region as all other bits are 0. The size of
+            /// the XSaveXtd is therefore zero.
+            xcomp_bv: 0x80000000, // TODO: This is setting bit 31. Check on which is correct.
```
npmccallum: I think it might make sense to have a helper type for `xcomp_bv`.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+            reserved0: Default::default(),
         }
     }
+}
 
+/// For our use case, the XSave Extended Region is in compacted format and currently holds nothing, as specifed in XCOMP_BV in
```
npmccallum: Wrap comment.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+/// https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf.
+#[derive(Debug, Default)]
+#[repr(C)]
+pub struct XSave {
+    /// Legacy region of the XSave area
+    pub legacy: XSaveLegacy,
+    /// XSave header
+    pub header: XSaveHeader,
+    /// XSave Extended Region (not used)
+    pub xtd: XSaveXtd,
```
npmccallum: `extend` is the same size as `legacy` and `header`.

sgx-types/src/ssa.rs
```diff
@@ -270,45 +272,143 @@ pub struct Miscellaneous {
...
+}
+
+/// When an AEX occurs while running in an enclave, the architectural state is saved
+/// in the thread’s current StateSaveArea (SSA Frame), which is pointed to by TCS.CSSA.
+///
+/// Section 38.9, Table 38-7
+#[repr(C)]
```
npmccallum: We probably do want an implementation of `Debug` here. Debugging the saved state will be common...

sgx-types/src/ssa.rs
lkatalin: Could you clarify what you mean by this?

npmccallum: `xtd` is an abbreviation. So is `extend`. But `extend` is the same length as `legacy` and `header`. It is merely an aesthetic point.

lkatalin: Ah okay, sure. : ) I thought you meant the struct size, so was pondering that.

sgx-types/src/ssa.rs
lkatalin: This is going to take a bit longer, so I'd prefer to submit it as a separate issue/PR.

lkatalin:
@npmccallum This new version has 2 commits:

1. Create `intel-types` crate. I moved all the `Xsave`-related structs here into `lib.rs` (for now, since we don't have much else in the crate). I also moved `Masked` and the `testaso!` macro here because we need these in both `intel-types` and `sgx-types` and I didn't want to be importing things back and forth between those two crates. This way, the only imports flow from `intel-types` => `sgx-types`.

2. The updates to `SSA` using the new `intel-types` crate.

npmccallum:
@lkatalin I hadn't realized we'd need `testaso` in the intel repo. Knowing that, I think our test macros need to end up into a separate crate used by both `intel-types` and `sgx-types`.

lkatalin:
@npmccallum I can add a separate `testing` crate as part of the first PR?

lkatalin:
@npmccallum This should be better organized now.

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+    pub struct XcompBv: u64 {
+
+        /// Compacted form is used for the layout of the XSAVE EXtended Region
+        const COMPACT = 1 << 63;
+    }
+}
+
+/// The legacy region of an XSAVE area comprises the 512 bytes starting at the area's base address.
+/// See Table 13-1.
+#[repr(C)]
```
npmccallum: Is there an alignment requirement?

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+        let mut xsl = MaybeUninit::<XSaveLegacy>::zeroed();
+        unsafe {
+            (*xsl.as_mut_ptr()).fcw = Fcw::from_bits(0x037F).unwrap();
+        }
+        unsafe {
+            (*xsl.as_mut_ptr()).mxcsr = Masked::<MxCsr> {
+                data: MxCsr::from_bits(0x1F80).unwrap(),
+                mask: MxCsr::from_bits(0xFFF).unwrap(),
+            }
+        }
```
npmccallum: This unsafe code isn't needed here. Figure out what the raw values mean in terms of flags and set those.

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+        let xsl_init = unsafe { xsl.assume_init() };
+        xsl_init
+    }
+}
+
+/// The XSAVE header of an XSAVE area comprises the 64 bytes starting at offset 512 from the
+/// area's base address. See Section 13.4.2.
+#[repr(C)]
```
npmccallum: Is there an alignment requirement?

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+pub struct XSaveHeader {
+    /// State-component bitmap identifying the state components in the XSAVE area.
+    pub xstate_bv: u64,
```
npmccallum: Is this supposed to be bitflags?

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+        }
+    }
+}
+
+/// For our use case, the XSave Extended Region is in compacted format and currently holds nothing,
+/// as specifed in XCOMP_BV in
+/// https://github.com/jsakkine-intel/linux-sgx/blob/master/tools/testing/selftests/x86/sgx/encl_bootstrap.S#L89.
+/// For more on this region, see Section 13.4.3.
+#[derive(Debug, Default)]
+#[repr(C)]
```
npmccallum: Is there an alignment requirement?

intel-types/src/lib.rs
```diff
@@ -52,12 +54,267 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+pub struct XSaveExtend;
+
+/// For details on the fields included in XSave, see Section 13.4.
+#[repr(C)]
```
npmccallum: Is there an alignment requirement?

sgx-types/src/ssa.rs
```diff
@@ -321,7 +321,6 @@ impl Footer {
 /// in the thread’s current StateSaveArea (SSA Frame), which is pointed to by TCS.CSSA.
 ///
 /// Section 38.9, Table 38-7
-#[derive(Debug)]
```
npmccallum: I don't think this belongs in this patch.

sgx-types/src/ssa.rs
```diff
@@ -315,18 +364,13 @@ impl Footer {
...
-pub struct StateSaveArea {
-    ///  Section 38.9, Table 38-7
-    pub xsave: XSave,
-    /// Section 38.9, Table 38-7
-    pub footer: Footer,
+    const fn padding() -> usize {
+        #[repr(C, align(4096))]
+        struct Unpadded(XSave, Miscellaneous, Gpr);
+
+        size_of::<Unpadded>() - size_of::<XSave>() - size_of::<Miscellaneous>() - size_of::<Gpr>()
```
npmccallum: I just realized that the following is probably more readable.

```rust
        #[repr(C, align(4096))]
        struct Aligned(XSave, Miscellaneous, Gpr);

        #[repr(C, packed)]
        struct Unaligned(XSave, Miscellaneous, Gpr);

        size_of::<Aligned>() - size_of::<Unaligned>()
```

sgx-types/src/ssa.rs
```diff
@@ -367,13 +411,10 @@ testaso! {
         exinfo: 0
     }
 
-    struct Footer: 4096, 4096 => {
-        misc: 4096 - core::mem::size_of::<Miscellaneous>() - core::mem::size_of::<Gpr>(),
-        gpr: 4096 - core::mem::size_of::<Gpr>()
-    }
-
-    struct StateSaveArea: 4096, 8192 => {
+    struct StateSaveArea: 8, 4096 => {
```
npmccallum: `StateSaveArea` definitely needs to be page aligned.

intel-types/src/lib.rs
lkatalin: The documentation says it takes up the first 512 bytes of the `XSave` region, starting from the `XSave` base address ([13.4.1](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf)). I understood this as a size requirement, but would this also be considered an alignment requirement? Other structs in the Intel documentation explicitly say "must be 512-byte aligned," etc.

npmccallum: No. It will say aligned if they want alignment.

intel-types/src/lib.rs
lkatalin: I don't believe so, as the size is variable based on [13.4.3](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf).

intel-types/src/lib.rs
lkatalin: "Each of the save and restore instructions takes a memory operand that specifies the 64-byte aligned base address of the XSAVE area on which it operates." Since the base address is 64-byte aligned, this means the entire XSAVE area is 64-byte aligned - is this the correct way to interpret this statement?

intel-types/src/lib.rs
lkatalin: Or, rather: The alignment requirements and size are both variable and based on the particular results of `CPU_ID`, so I think we were going to skip them.

intel-types/src/lib.rs
lkatalin: I might need some guidance here; did you want me to set the entire `XSaveLegacy` region as `bitflags!`? Other than this `unsafe` hack, I'm reading that it's not possible to create a `MaybeUninit` struct and then write to its fields: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field

sgx-types/src/ssa.rs
lkatalin: If `XSave` is 64-byte aligned (which I'm not sure; see comment above) then this won't work because `error[E0588]: packed type cannot transitively contain a `[repr(align)]` type`.

npmccallum: Okay then. Never mind. :)

intel-types/src/lib.rs
lkatalin: I've updated the flags; though as mentioned I don't know how to get rid of the `unsafe` block.

intel-types/src/lib.rs
lkatalin: There isn't here.

intel-types/src/lib.rs
```diff
@@ -40,12 +42,323 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+
+impl Default for XSaveHeader {
+    fn default() -> Self {
+        Self {
+            xstate_bv: Default::default(),
+
+            /// XCOMP_BV[63] = 1, compaction mode for the XSave Extended Region. However, no state
+            /// components are included in the Extended Region as all other bits are 0. The size of
+            /// the XSaveXtd is therefore zero.
+            xcomp_bv: XcompBv::COMPACT, // TODO: Check this value; the original (0x80000000) sets bit 31, not 63
```
npmccallum: Same as above. `impl Default for XcompBv` then derive default here.

intel-types/src/lib.rs
```diff
@@ -40,12 +42,315 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+    fn default() -> Self {
+        let fcw = Fcw::INV_OP
+            | Fcw::DENORM_OP
+            | Fcw::ZERO_DIV
+            | Fcw::OVERFLOW
+            | Fcw::UNDERFLOW
+            | Fcw::PREC
+            | Fcw::PREC_CTRL0
+            | Fcw::PREC_CTRL1;
+        fcw
```
npmccallum: Remove the `fcw` temporary variable.

intel-types/src/lib.rs
```diff
@@ -40,12 +42,315 @@ impl<T: Copy + Debug + PartialEq + BitAnd<Output = T>> PartialEq<T> for Masked<T
...
+    /// to identify any reserved bits in MXCSR (none are reserved here).
+    /// See Sections 11.6.4 and 11.6.6.
+    fn default() -> Self {
+        let mxcsr = MxCsr::INV_OP_MASK
+            | MxCsr::DENORM_MASK
+            | MxCsr::ZERO_DIV_MASK
+            | MxCsr::OVERFLOW_MASK
+            | MxCsr::UNDERFLOW_MASK
+            | MxCsr::PREC_MASK;
+        mxcsr
```
npmccallum: Remove temporary variable `mxcsr`.

PR #176, opened 2020-01-29
First pass on traits

npmccallum:
This is a stab at a high-level interface for bridging both SGX and SEV keeps. I know I'm missing stuff. And I'm sure I've gotten stuff wrong. But this is enough to start talking about.

In particular, this doesn't handle the multi-threaded case at all. Ideally, we will start a keep with a single thread and it can add more. We don't currently have a facility to handle multiple threads with this API.

Comments welcome!

### DIFF
--- a/enarx-keepmgr/src/drivers/mod.rs
+++ b/enarx-keepmgr/src/drivers/mod.rs
@@ -1,0 +1,111 @@
+//! # Overview
+//!
+//! The traits contained in this module constitute a state machine for
+//! launching and managing a keep across multiple technologies. Launching
+//! a keep goes through multiple phases, in this order:
+//!
+//!   1. Driver
+//!
+//!   2. Setup
+//!     a. Shim Setup
+//!     b. Keep Setup
+//!
+//!   3. Build
+//!     a. Shim Build
+//!     b. Keep Build
+//!
+//!   4. Enter
+//!
+//! # Details
+//! ## The Driver Phase
+//!
+//! First, the keep manager must find an appropriate driver for the system.
+//! Then, the keep manager must locate the keep runtime and the driver-supplied
+//! shim for use in the next phase. When the keep manager is ready to launch
+//! a keep, it uses `Driver::make()` to begin the setup phase.
+//!
+//! ## The Setup Phase
+//!
+//! The purpose of the setup phase is to give the driver a view of both the
+//! keep runtime and shim before creating the actual keep. This is useful for
+//! technologies like SGX where we need to hash and sign the binary before
+//! creating the keep.
+//!
+//! The keep manager will load the driver-supplied shim first, followed by
+//! keep runtime. Upon completion, we begin the build phase.
+//!
+//! ## The Build Phase
+//!
+//! The purpose of the building phase is to actually create a full keep
+//! with all the pages actually loaded. The workflow is similar to the
+//! setup phase.
+//!
+//! The keep manager will load the driver-supplied shim first, followed by
+//! keep runtime. Upon completion, we begin the enter phase.
+//!
+//! ## The Enter Phase
+//!
+//! The enter phase can be considered the main loop of the application.
+//! When the keep manager enters the keep, the keep will run until an event
+//! occurs. Upon an event, the keep manager will handle the event, potentially
+//! collecting the keep's required data, and re-entering the keep.
+
+pub mod debug;
+
+use crate::access::Access;
+use crate::span::Span;
+
+use std::io::Result;
+use std::path::Path;
+
+/// All possible keep events
+#[allow(non_camel_case_types)]
+pub enum Event {
+    /// Exit the process with the supplied exit value
+    exit(i32),
+
+    /// Get the user identifier of the process
+    getuid(Keep<libc::uid_t>),
+}
+
+/// A trait for supplying statically-linked pages into the keep
+pub trait Loader<T> {
+    /// Loads data into keep memory
+    fn load(&mut self, src: &[u8], dst: Span<u64>, access: Access) -> Result<()>;
+
+    /// Completes the loading process
+    fn done(self: Box<Self>, entry: u64) -> Result<T>;
+}
+
+/// A trait containing the function used to enter a keep
+pub trait Enterer<T> {
+    /// Enter the keep with the provided input until the next event
+    fn enter(self: Box<Self>, input: T) -> Result<Event>;
+}
+
+/// A driver for a particular type of keep
+pub trait Driver {
+    /// The name of this keep driver
+    fn name(&self) -> &str;
+
+    /// The path to the shim for this driver
+    fn shim(&self) -> Result<&Path>;
+
+    /// Begin the constructon process for a keep
+    fn make(&self) -> Result<ShimSetup>;
+}
+
+/// The shim half of the setup phase
+pub type ShimSetup = Box<dyn Loader<KeepSetup>>;
+
+/// The keep half of the setup phase
+pub type KeepSetup = Box<dyn Loader<ShimBuild>>;
+
+/// The shim half of the build phase
+pub type ShimBuild = Box<dyn Loader<KeepBuild>>;
+
+/// The keep half of the build phase
+pub type KeepBuild = Box<dyn Loader<Keep<()>>>;
+
+/// A keep which expects the input type `T` in order to resume execution
+pub type Keep<T> = Box<dyn Enterer<T>>;
--- a/enarx-keepmgr/Cargo.toml
+++ b/enarx-keepmgr/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "enarx-keepmgr"
+version = "0.1.0"
+authors = ["Nathaniel McCallum <npmccallum@redhat.com>"]
+edition = "2018"
+license = "Apache 2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+bitflags = "1.2.1"
+goblin = "0.2.0"
+libc = "0.2.66"
--- a/enarx-keepmgr/src/access.rs
+++ b/enarx-keepmgr/src/access.rs
@@ -1,0 +1,19 @@
+use bitflags::bitflags;
+
+bitflags! {
+    /// Memory access permissions
+    ///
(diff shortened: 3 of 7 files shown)


enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,147 @@
+//! The traits contained in this crate constitute a big state machine for
+//! lauching and managing a keep across multiple technologies. Launching
```
whitebrandy: Nit: typo in "launching"

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    None,
+}
+
+/// A trait for supplying statically linked pages into the keep
+pub trait Loader {
+    fn load(&mut self, pages: &[u8], offset: u64, access: Access) -> Result<()>;
+}
+
+/// Prepares to build the keep
+pub trait Preparer: Sized + Loader {
```
whitebrandy: IIUC, this means that the type must implement both Sized and Loader. I guess I kind of understand why Loader is required but why is Sized required ? What happens if we don't specify Sized here ?

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/// A keep reply
+pub enum Reply {
+    Halt,
+}
+
+/// A trait for supplying statically linked pages into the keep
+pub trait Loader {
+    fn load(&mut self, pages: &[u8], offset: u64, access: Access) -> Result<()>;
```
haraldh: What about .bss sections with `filesize = 0` and `memsize > 0`

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//! and sign the binary before creating the keep.
+//!
+//! A keep loader will start by creating a `Preparer` object by calling
+//! `Preparer::new()`. If the technology represented by this type is not
+//! supported by the system, it will return `None`.
+//!
+//! Once a `Preparer` is obtained, the keep manager will call `Loader::load()`
+//! to load all the pages of the static binary. Then finally, the manager
+//! will call `Preparer::next()` to specify the entry point and conclude the
+//! preparation state.
```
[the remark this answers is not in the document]
npmccallum: You could think of it that way. Basically, on SGX we need to 1. create a signature over the enclave bits 2. actually construct the enclave. One of these is preparation and one is building. How it is actually structured is an internal implementation detail.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//!
+//! ## Building
+//!
+//! The purpose of the building state is to actually create a full keep
+//! with all the pages actually loaded. The workflow is similar to the
+//! preparation state.
```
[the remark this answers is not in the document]
npmccallum: See above.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        const W = 1 << 1;
+
+        /// Read access
+        const R = 1 << 2;
+    }
+}
+
+/// A keep request
+pub enum Request {
+    Halt,
```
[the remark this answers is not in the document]
npmccallum: I'll explain below.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// This specifies the entry point for the binary and the socket on which
+    /// to perform attestation and encrypted communication. Upon success,
+    /// returns the new keep.
+    fn next(self, entry: u64, socket: impl Read + Write) -> Result<Self::Keep>;
+}
+
+/// Executes a keep
+pub trait Keep {
+    /// Runs the keep until a event occurs.
+    fn run(&mut self, reply: Reply) -> Result<Reply>;
```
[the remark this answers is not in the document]
npmccallum: This area still needs some thinking. But the basic idea is this. The keep runs until it needs something from the keep manager. This is a `Request`. The keep manager handles the `Request` and sends the `Reply` back to the keep.

I haven't solved the problem that the first time you call `run()` you don't have a `Reply` yet because there has been no request. Suggestions welcome.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Builds a keep
+pub trait Builder: Sized + Loader {
+    type Keep: Keep;
+
+    /// Completes the building process after loading all pages
+    ///
+    /// This specifies the entry point for the binary and the socket on which
+    /// to perform attestation and encrypted communication. Upon success,
+    /// returns the new keep.
+    fn next(self, entry: u64, socket: impl Read + Write) -> Result<Self::Keep>;
```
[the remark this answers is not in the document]
npmccallum: No. The launching of the keep happens with `Builder::next()`. Attestation happens out of band between the Keep and the tenant. The host merely proxies the information. We need better structure here. But this is what I have so far.

enarx-srv-traits/src/lib.rs
npmccallum: Good question. In SGX these have to be pre-expanded. That is, you allocate a temporary empty page and pass it in. But I suppose this could be handled internally for those technologies that require it. Is this a better API?

```suggestion
    fn load(&mut self, src: &[u8], dst: Range<u64>, access: Access) -> Result<()>;
```

Things like `src` alignment and intermediate temporary buffers need to be fixed up internally, I guess.

enarx-srv-traits/src/lib.rs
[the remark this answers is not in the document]
lkatalin: Ah okay, thanks for the explanation! It wouldn't work as an `Option<Reply>`?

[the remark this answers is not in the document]
npmccallum: I'll probably do that for now. But I'd like the compiler to enforce a `Reply` after the first `Request` is received.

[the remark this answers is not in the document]
connorkuehl: We could make a special `Reply` that is used only for the first run of the Keep. A kind of stub. In that case, we may need to keep some state to indicate we have seen it, as additional "first run" replies will probably be considered spurious and discarded since the Keep has already been running.

The vocabulary here is probably what we're getting caught up on for that first case. We could enclose a `Reply` or a `Request` in a top-level type called `Message`.

enarx-srv-traits/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "enarx-srv-traits"
+version = "0.1.0"
+authors = ["Nathaniel McCallum <npmccallum@redhat.com>"]
+edition = "2018"
+
```
connorkuehl: We may need a license field here, otherwise if #162 is merged first, we might get barked at.

enarx-srv-traits/src/lib.rs
npmccallum: I have done something similar in my latest push (forthcoming...).

enarx-srv-traits/src/lib.rs
npmccallum: @whitebrandy @lkatalin See the answer below.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/// Prepares to build the keep
+pub trait Preparer: Sized + Loader {
+    type Builder: Builder;
+
+    /// Creates a new Preparer
+    ///
+    /// Returns None when this preparer type is not supported.
+    fn new() -> Result<Option<Self>>;
```
npmccallum: @whitebrandy @lkatalin Because `new()` returns `Self`, `Self` must be sized. Otherwise, the caller doesn't know how to allocate enough memory.

enarx-srv-traits/src/lib.rs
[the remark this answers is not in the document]
npmccallum: I have a solution for this coming in a forthcoming push. My solution not only ensures the correct value as input for the first call, but it also ensures that the compiler enforces the correct `Reply` data corresponds to the `Request`.

enarx-srv-traits/src/lib.rs
npmccallum: Fixed. Thanks.

enarx-srv-traits/Cargo.toml
npmccallum: I fixed this in the forthcoming push.

enarx-srv-traits/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//! entity (usually the tenant). This state proxies packets back and forth
+//! between the keep and the tenant. Note that the contents of attestation
+//! packets is defined by the specific technology in question.
+//!
+//! ## Launching
+//!
+//! The purpose of the launching state is to proxy communication between
+//! the tenant and the keep after attestation. Note that the intent of this
+//! state is to proxy integrity protected ciphertext. The manager should
+//! have no visibility into its contents.
```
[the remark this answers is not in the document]
npmccallum: I think I've done this in the next version (forthcoming).

npmccallum:
@enarx/everyone I've pushed a rewrite of the traits. Please review.

enarx-keepmgr/src/drivers/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+//! The keep manager will load the driver-supplied shim first, followed by
+//! keep runtime. Upon completion, we begin the enter phase.
+//!
+//! ## The Enter Phase
+//!
+//! The enter phase can be considered the main loop of the application.
+//! When the keep manager enters the keep, the keep will run until an event
+//! occurs. Upon an event, the keep manager will handle the event, potentially
+//! collecting the keep's required data, and re-entering the keep.
+
```
lkatalin: Is it correct that attestation would happen after this? (The Keep needs to be running in order to issue a Report about itself, at least for SGX.)

npmccallum: For `SGX`, yes. For `SEV`, no.

npmccallum:
@enarx/everyone Upon further thought, I don't want to split `Span` off yet. Let's review this as it stands. I am now proposing it for merge.

Issue #153, opened 2020-01-22
Test which checks for lints


npmccallum:
All our crates should have the following lints turned on:

- `#![deny(missing_docs)]`
- `#![deny(clippy:all)]`

We should have a test which confirms this and fails if any required lints are missing.

Note, however, that we *will* allow selective disabling of specific lints on a case by case basis.

squidboylan:
It doesn't look like i can assign this to myself but I'm happy to work on this.

squidboylan:
https://github.com/enarx/enarx/pull/182 Addresses the `#![deny(clippy:all)]` part of this PR, however it doesn't address the `#![deny(missing_docs)]` piece as I don't have enough knowledge to add the missing docs.

npmccallum:
@squidboylan I would consider it acceptable to grandfather in old crates by doing this:

```rust
#![deny(missing_docs)]

// TODO: link to technology debt issue for documentation
#![allow(missing_docs)]
```

There is value in enforcing this lint on all new crates earlier rather than later.

PR #182, opened 2020-01-31
Add clippy hook

squidboylan:
This addresses one part of https://github.com/enarx/enarx/issues/153 . It doesn't turn on `#![deny(missing_docs)]` because I don't have the necessary context to add all the missing docs.

### DIFF
--- a/sev/src/lib.rs
+++ b/sev/src/lib.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![warn(clippy::all)]
+#![deny(clippy::all)]
 #![allow(unknown_lints)]
 #![allow(clippy::identity_op)]
 #![allow(clippy::unreadable_literal)]
--- a/sgx-show/src/data/cpuid.rs
+++ b/sgx-show/src/data/cpuid.rs
@@ -34,21 +34,23 @@ pub trait Leaf: std::fmt::Display {
 
 pub struct CpuId<T: Leaf>(pub u128, PhantomData<T>);
 
+#[allow(clippy::identity_op)]
+#[allow(clippy::erasing_op)]
 impl<T: Leaf> CpuId<T> {
     pub fn eax(&self) -> u32 {
-        (self.0 >> 3 * 32) as u32
+        (self.0 >> (3 * 32)) as u32
     }
 
     pub fn ebx(&self) -> u32 {
-        (self.0 >> 2 * 32) as u32
+        (self.0 >> (2 * 32)) as u32
     }
 
     pub fn ecx(&self) -> u32 {
-        (self.0 >> 1 * 32) as u32
+        (self.0 >> (1 * 32)) as u32
     }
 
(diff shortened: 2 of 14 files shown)


sev/src/lib.rs
```diff
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![warn(clippy::all)]
+#![deny(clippy::all)]
 #![allow(unknown_lints)]
 #![allow(clippy::identity_op)]
 #![allow(clippy::unreadable_literal)]
+#![allow(clippy::uninit_assumed_init)]
```
squidboylan: Clippy gets angry about this because it is potential UB. However I'm not sure how to fix this, it looks like the lack of initialization is intentional, but is also UB.

sgx-show/src/data/cpuid.rs
```diff
@@ -21,10 +21,10 @@ macro_rules! mkcpuid {
...
 
 mkcpuid! {
-    CpuInfo: 0x00000000, 0x00000000;
-    ExtFeat: 0x00000007, 0x00000000;
-    SgxCaps: 0x00000012, 0x00000000;
-    SgxAttr: 0x00000012, 0x00000001;
+    CpuInfo: 0x0000_0000, 0x0000_0000;
+    ExtFeat: 0x0000_0007, 0x0000_0000;
+    SgxCaps: 0x0000_0012, 0x0000_0000;
+    SgxAttr: 0x0000_0012, 0x0000_0001;
```
npmccallum: I think I prefer selectively disabling the unreadable literal lint.

sev/src/lib.rs
npmccallum: I would selectively add `#[allow(clippy::uninit_assumed_init)]` to the blocks in question rather than adding it here.

.gitignore
```diff
@@ -1,2 +1 @@
-Cargo.lock
```
npmccallum: The general strategy is to commit `Cargo.lock` for binaries but not for libraries. Currently we don't commit any of them because they cause lots of merge conflicts during the rapid development phase of the project. So please don't remove this.

squidboylan: Yeah i only removed this to test my theory of why travis was breaking, wasnt planning on leaving it in

sgx-show/src/data/cpuid.rs
```diff
@@ -36,19 +36,19 @@ pub struct CpuId<T: Leaf>(pub u128, PhantomData<T>);
...
     }
 
     pub fn ecx(&self) -> u32 {
-        (self.0 >> 1 * 32) as u32
+        (self.0 >> 32) as u32
     }
 
     pub fn edx(&self) -> u32 {
-        (self.0 >> 0 * 32) as u32
+        self.0 as u32
```
npmccallum: I think the code is more readable with the identity ops. And the compiler will optimize them out. So let's just disable the lint.

npmccallum:
Thanks @squidboylan ! Congrats on your first merge!

PR #190, opened 2020-02-03
SECS improvements

npmccallum:
This PR contains two patches that improve the `Secs` type.


npmccallum:
@lkatalin Please file that issue.

PR #208, opened 2020-02-05
Make Rust-related pre-push hooks conditional on code changes in commit

### LINKED ISSUE #206: Conditionally run tests based on files modified
As we flesh out our testing suite, we're noticing that our pre-push hooks are still running on commits with no code changes. This can slow down quick documentation changes, or even hold up an entire workflow if the build is broken for an unrelated reason.

We should make some of our tests conditional. Specifically, if no `.rs` or `.toml` files were modified, we should not run our Rust-related tests.

mbestavros:
Resolves #206. Thanks to @axelsimon for pointing this out.


hooks/pre-push
```diff
@@ -1,2 +1,11 @@
 #!/bin/bash -e
-for f in $0.d/*; do $f; done
+rust=0
+for file in $(git diff-index --name-only HEAD); do
+    if [[ -n "$TRAVIS" ]] || [[ $file == *.rs ]] || [[ $file == *.toml ]]; then
```
npmccallum: `[[ $file == *.toml ]]`

That's the first I've ever seen this syntax. I assume it checks that the file ends with `.toml`.

hooks/pre-push
```diff
@@ -1,2 +1,11 @@
...
+        rust=1
+        break
+    fi
+done
+for file in $0.d/*; do
+    if [[ $(echo $file | rev | cut -d'/' -f 1 | rev) != cargo-* ]] || [[ $rust == 1 ]]; then $file; fi
```
npmccallum: `
```suggestion
    if [[ $(basename $file) != cargo-* ]] || [[ $rust == 1 ]]; then
        $file
    fi
```

hooks/pre-push
mbestavros: Yes, it does. I suppose we could change it to explicitly check for `Cargo.toml` if we wanted.

hooks/pre-push
mbestavros: Good to know for the future. Amended.

mbestavros:
Amended the file name check to use `basename`.

This brings up a good question @npmccallum -- do we want a specialized way to identify Rust tests?Checking for `cargo-` leading the name works for now, but it my be awkward if we end up doing any Rust tests that don't involve Cargo.

My first revision of this PR renamed all the Rust tests to have `-rust` at the end of them, but I'm curious if there are opinions on a better way to do it.

npmccallum:
> Amended the file name check to use `basename`.
> 
> This brings up a good question @npmccallum -- do we want a specialized way to identify Rust tests?Checking for `cargo-` leading the name works for now, but it my be awkward if we end up doing any Rust tests that don't involve Cargo.
> 
> My first revision of this PR renamed all the Rust tests to have `-rust` at the end of them (kind of like a tag), but I'm curious if there are opinions on a better way to do it.

I would prefer a prefix. But there's really not another way.

hooks/pre-push
```diff
@@ -1,2 +1,13 @@
 #!/bin/bash -e
-for f in $0.d/*; do $f; done
+rust=0
+for file in $(git diff-index --name-only HEAD); do
```
npmccallum: I'm not 100% sure what this command does. But let me think aloud here for a moment. I can think of the following cases:

1. `git push origin foo` -- new branch `foo`

In this case we are creating a remote branch and pushing all commits to it. We have no way to determine which commits are "new". We run tests.

2. `git push origin foo` -- append to existing branch `foo`

We need a list of all files changed in the commits we are appending. We conditionally run tests if code files have changed.

3. `git push -f origin foo` -- overwrite existing branch

We don't know which commits are "new". Do we run tests? I'm not sure. It is certainly nice to have an option to disable tests (i.e. `-f` == don't run tests). However, when doing things like squashing commits we need to specify `-f` **and** run tests. So I think we have to run tests.

4. CI/CD

This case is similar to the appending case above, but rather than appending we are managing the difference between two forks. So we need to know which commits differ between two revisions.

I'm honestly wondering if this problem is too hard to solve right now... Or conversely, if trying to run tests pre-push was the wrong decision. Thoughts?

hooks/pre-push
```diff
@@ -1,2 +1,13 @@
...
+    if [[ -n "$TRAVIS" ]] || [[ $file == *.rs ]] || [[ $file == *.toml ]]; then
+        rust=1
+        break
+    fi
+done
+for file in $0.d/*; do
+    if [[ $(basename $file) != cargo-* ]] || [[ $rust == 1 ]]; then
```
npmccallum: The obvious way to filter here that I didn't think of before is to put Rust tests in a separate directory.

hooks/pre-push
mbestavros: To run through my logic when I wrote this:

The command (`git diff-index --name-only HEAD`) lists all filenames that differ from HEAD in the current branch. (I got the command [here](https://stackoverflow.com/questions/3882838/whats-an-easy-way-to-detect-modified-files-in-a-git-workspace).) The `for` loop iterates through all these files and checks if they meet some criteria that would indicate we need to run full Rust tests:

- a `.rs` or `.toml` file was edited
- we are running on CI (Travis, in this case)

If any of them are met, we flip a status bit. That's all we need to determine we need to run tests, so the loop breaks.

We then check, for every test in the `pre-push.d` directory, if they are a `cargo-` test. If they are, the status bit must also be enabled to run them. Otherwise, the test is run regardless.

I suppose a lot of my logic is predicated upon the `git diff-index` command accurately telling us which files have changed. I think if that assumption is sound, a lot of these cases are adequately addressed with the current logic. (That said, I'll need to debug Travis tests... doesn't seem like they're force-run yet.)

I will also say that I like having pre-push tests (or more generally, having tests run locally before pushing), as long as we can make them get out of the way when not needed. Is your thought more about the fact we should change where they're run, or whether we should have git hooks tests at all?

npmccallum: @mbestavros The problem is that you're only checking the last commit. But the user might be pushing more than one commit. So imagine a case like this:

Commit 1: change some code
Commit 2: change some docs

The existing code would look only at commit 2. Seeing no code changes, it will skip running tests. The problem is that both commits are new.

This leads to the four scenarios I listed above.

mbestavros: I see now.

I suppose we want to compare the current branch with `upstream/master`, which may be difficult if an upstream has not been set up... I think if we can reliably determine those things, we can make accurate judgement calls for all cases above.

So the question of the hour: is there a command that will do that?

hooks/pre-push
mbestavros: Yep, that also makes for much cleaner code. I've moved things around in the latest push.

hooks/pre-push
npmccallum: I don't think there's a ready-made way to do this, no. There are also other problems.

First, writing tests that have to execute in multiple special environments can be challenging.

Second, pre-push hooks cannot be enabled by default. They are opt-in.

Third, opting in to pre-push hooks means developers have to setup special environments just for testing. This isn't developer friendly.

Fourth, tests currently run in the working tree: https://github.com/enarx/enarx/issues/214 .

Hence, I'm wondering if this was a mistake (which I will completely own). Perhaps we should just focus on the CI/CD problem and migrate our tests there.

mbestavros:
@npmccallum Should we close this PR, given we're moving away from Travis and hooks now?

mbestavros:
We're moving away from pre-push hooks as a testing strategy generally, so we're abandoning this PR. Closing.

PR #209, opened 2020-02-05
Measure SSA frame size in 4k pages

### LINKED ISSUE #200: SECS.SSAFRAMESIZE is number of pages
All of our existing code presumes that `SECS.SSAFRAMESIZE` is in bytes. However, the specification says:

> Size of a frame in the State Save Area (SECS.SSAFRAMESIZE). Defines the number of 4K byte
> pages in a single frame in the State Save Area. Must be large enough to hold the GPR state, the
> XSAVE state, and the MISC state.

We should add a helper method `StateSaveArea::frame_size() -> NonZeroU32` that returns the correct value based on the `size_of::<StateSaveArea>()`. All existing code needs to be fixed.

lkatalin:
Fixes #200 

@npmccallum Let me know what you think of this or if further changes are needed. 

I added the 4096 alignment requirement to `SSA` because the documentation says that the end of the `GPR` section (the last section in `SSA`) must be aligned to the end of a page. For this reason, if the size of `SSA` is not a multiple of 4096, `SSA::frame_size()` returns `None`.

I wasn't sure whether we want a `secs::Spec::ssa_size()` function, so I added one that just calls the one from `SSA`.

I also didn't see any code that uses the SSA frame size and explicitly expects it to be in bytes; if I missed any of these that you know of, please let me know and I'll update them. [This line](https://github.com/enarx/enarx/blob/master/iocuddle-sgx/src/lib.rs#L140) in `iocuddle-sgx/src/lib.rs` might be one, but it's an arbitrary value that seems like it could refer to pages as easily as bytes.

### DIFF
--- a/sgx-types/src/secs.rs
+++ b/sgx-types/src/secs.rs
@@ -7,7 +7,7 @@
 //! page created for any enclave. It is moved from a temporary buffer to an EPC
 //! by the means of ENCLS(ECREATE) leaf.
 
-use super::{attr, isv, misc::MiscSelect, sig::Contents};
+use super::{attr, isv, misc::MiscSelect, sig::Contents, ssa::StateSaveArea};
 use core::num::{NonZeroU32, NonZeroU64};
 
 /// An enclave's size specification
@@ -66,6 +66,15 @@ impl Spec {
         let res = unsafe { __cpuid_count(LEAF_MAX_ENCL_SIZE, SUBLEAF_MAX_ENCL_SIZE) };
         let max_size: u64 = 1 << (res.edx >> 8 as u8) as u64;
         Some(NonZeroU64::new(max_size).unwrap())
+    }
+}
+
+impl Default for Spec {
+    fn default() -> Self {
+        Self {
+            enc_size: Spec::max_enc_size().unwrap(),
+            ssa_size: StateSaveArea::frame_size(),
+        }
     }
 }
 
--- a/sgx-types/src/ssa.rs
+++ b/sgx-types/src/ssa.rs
@@ -7,6 +7,10 @@
 #![allow(clippy::unreadable_literal)]
 
 use bitflags::bitflags;
+use core::{
+    mem::{align_of, size_of},
+    num::NonZeroU32,
+};
 
 /// Section 38.9.1.1, Table 38-10
 #[derive(Copy, Clone, Debug)]
@@ -318,12 +322,23 @@ impl Footer {
 ///
 /// Section 38.9, Table 38-7
 #[derive(Debug)]
-#[repr(C)]
+#[repr(C, align(4096))]
 pub struct StateSaveArea {
     ///  Section 38.9, Table 38-7
     pub xsave: XSave,
     /// Section 38.9, Table 38-7
     pub footer: Footer,
+}
+
+impl StateSaveArea {
+    /// Returns the size of the SSA in 4k pages, rather than bytes. According to
+    /// the documentation, SSAFrameSize (used in SECS) is referenced in pages.
+    /// See section 38.7 for page-size requirement for SECS and Table 38-7 for
+    /// requirement that the SSA is page aligned.
+    pub const fn frame_size() -> NonZeroU32 {
+        let pages = size_of::<StateSaveArea>() / align_of::<StateSaveArea>();
+        unsafe { NonZeroU32::new_unchecked(pages as u32) }
+    }
 }
 
 testaso! {


sgx-types/src/secs.rs
```diff
@@ -65,6 +65,12 @@ impl Spec {
         let max_size: u64 = 1 << (res.edx >> 8 as u8) as u64;
         Some(NonZeroU64::new(max_size).unwrap())
     }
+
+    /// Returns the size of the SSA in 4k pages, rather than bytes. For more, see documentation
+    /// on ssa::StateSaveArea::frame_size().
+    pub fn ssa_size() -> Option<NonZeroU32> {
+        StateSaveArea::frame_size()
+    }
```
npmccallum: It is possible that the consumer of this library doesn't want to use our `StateSaveArea` type. That makes the size of the SSA a property of the `StateSaveArea` type rather than the `Spec` type. Please move the method there.

npmccallum: However, I would take an `impl Default for Spec` where the defaults use the `StateSaveArea` and the maximum enclave size.

sgx-types/src/ssa.rs
```diff
@@ -330,14 +331,29 @@ impl Footer {
...
     pub footer: Footer,
 }
 
+impl StateSaveArea {
+    /// Returns the size of the SSA in 4k pages, rather than bytes. According to
+    /// the documentation, SSAFrameSize (used in SECS) is referenced in pages.
+    /// If the SSAFrameSize is not 4k page aligned, returns None. See section
+    /// 38.7 for page-size requirement for SECS and Table 38-7 for requirement
+    /// that the SSA is page aligned.
+    pub fn frame_size() -> Option<NonZeroU32> {
```
npmccallum: Don't use option here. We're defining a method on the `StateSaveArea` type. We'll just make sure it has the right alignment. I'd also implement the body as:

```rust
assert!(size_of::<StateSaveArea>() >= align_of::<StateSaveArea>());
let pages = size_of::<StateSaveArea>() / align_of::<StateSaveArea>();
unsafe { NonZeroU32::new_unchecked(pages) }
```

Then you can make the function `const`.

lkatalin: It looks like we have to choose between the `const` and the assertion:
```
error[E0723]: loops and conditional expressions are not stable in const fn
   --> src/ssa.rs:349:9
       |
349 |         assert!(size_of::<StateSaveArea>() >= align_of::<StateSaveArea>());
       |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

https://github.com/rust-lang/rust/issues/51999

npmccallum: @lkatalin Ditch the assertion. We control the size of the `StateSaveArea`. It is our problem.

sgx-types/src/secs.rs
lkatalin: Not sure why this isn't marked as `Outdated`, but I did make this change.

sgx-types/src/secs.rs
```diff
@@ -67,6 +67,15 @@ impl Spec {
     }
 }
 
+impl Default for Spec {
+    fn default() -> Self {
+        Self {
+            enc_size: unsafe { Spec::max_enc_size().unwrap() },
```
npmccallum: So it appears that the general practice within the Rust community is to unconditionally issue the `CPUID` instruction. So rather than an `unsafe` block here, let's just make `max_enc_size()` safe. However, please make sure there is a safety section in the documentation for that function that clearly outlines why we made that choice.

sgx-types/src/ssa.rs
```diff
@@ -330,14 +334,26 @@ impl Footer {
...
     /// Section 38.9, Table 38-7
     pub footer: Footer,
 }
 
+impl StateSaveArea {
+    /// Returns the size of the SSA in 4k pages, rather than bytes. According to
+    /// the documentation, SSAFrameSize (used in SECS) is referenced in pages.
+    /// If the SSAFrameSize is not 4k page aligned, returns None. See section
+    /// 38.7 for page-size requirement for SECS and Table 38-7 for requirement
+    /// that the SSA is page aligned.
```
npmccallum: You need to update the documentation for this function.

sgx-types/src/secs.rs
lkatalin: I'm not sure how they are doing that. 
```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
  --> src/secs.rs:52:19
     |
52 |         let res = core::arch::x86_64::__get_cpuid_max(LEAF_MAX_PARAM);
     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
     |

```

npmccallum: They immediately wrap that in an unsafe block.

npmccallum: It is probably not correct. But it is common practice. Which is why I suggested documenting it.

sgx-types/src/secs.rs
```diff
@@ -39,34 +39,47 @@ impl Spec {
...
+        let res = unsafe { core::arch::x86_64::__cpuid_count(LEAF_MAX_ENCL_SIZE, SUBLEAF_MAX_ENCL_SIZE) };
         let max_size: u64 = 1 << (res.edx >> 8 as u8) as u64;
         Some(NonZeroU64::new(max_size).unwrap())
     }
 }
 
+impl Default for Spec {
+    /// # Safety
+    /// This function involves calling the `CPUID` instruction, which is unsafe because it
+    /// does not check whether this instruction is available before issuing it.
```
npmccallum: This safety documentation needs to go on `max_enc_size()`. It should explain that although issuing `CPUID` will crash on unsupported CPUs, this is extremely rare and common practice is to just wrap `CPUID` in unsafe blocks and ignore the unsafety. You can also link to this issue for more details: https://github.com/rust-lang/rust/issues/60123

lkatalin: I had it on `max_enc_size()` also, but I've now expanded the documentation on `max_enc_size()` and now removed it completely from `Spec::default()`.

npmccallum: Please use this text:

```
This function is technically unsafe because it does not check if the `CPUID` instruction is available before issuing it. This could result in a crash on some very old CPUs. However, the only modern context where it could crash is environments like SGX or some virtualized CPUs. But even in these contexts it is common to trap and emulate the instruction.

Therefore, it is common practice to ignore the test to see if the `CPUID` instruction is available and just issue it anyway. For this reason, we are marking this function as safe. For more background to the state of `CPUID` in Rust, see: https://github.com/rust-lang/rust/issues/60123
```

Also, please mark `max_enc_size()` as unsafe in a separate commit. The commit message should contain the above text as well.

sgx-types/src/secs.rs
```diff
@@ -27,34 +27,48 @@ impl Spec {
...
+        let res = unsafe { core::arch::x86_64::__get_cpuid_max(LEAF_MAX_PARAM) };
         let max_leaf = res.0;
         if max_leaf < LEAF_SGX_SUPPORT || max_leaf < LEAF_MAX_ENCL_SIZE {
             return None;
         }
 
         // Test for SGX support
-        let res = core::arch::x86_64::__cpuid_count(LEAF_SGX_SUPPORT, SUBLEAF_SGX_SUPPORT);
+        let res =
+            unsafe { core::arch::x86_64::__cpuid_count(LEAF_SGX_SUPPORT, SUBLEAF_SGX_SUPPORT) };
```
npmccallum: Rather than wrapping the line here, do `use core::arch::x86_64::__cpuid_count;` at the top of the function.

npmccallum:
@lkatalin You need to rebase this on master. FYI, I'm blocked on needing this PR.

PR #210, opened 2020-02-05
Test all Cargo.toml files for edition = "2018"

### LINKED ISSUE #161: Test Cargo.toml for edition
All `Cargo.toml` files should have:

```toml
[package]
edition = "2018"
```

mbestavros:
Resolves #161.

An extension of the `Cargo.toml` license test from #162. If that changes upon merge, changes from it should be echoed here.

### DIFF
--- a/hooks/pre-push.d/cargo-edition-test
+++ b/hooks/pre-push.d/cargo-edition-test
@@ -1,0 +1,11 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -name Cargo.toml); do
+    if [ "$(toml get "$file" package.edition)" != '"2018"' ]; then
+        echo "$file is not using edition = "2018"."
+        echo "Please ensure the field is present and includes the correct Rust edition."        
+        status=1
+    fi
+done
+exit $status
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,4 +39,5 @@ install:
   - rustup component add rustfmt
   - rustup component add clippy
   - rustup target add x86_64-unknown-linux-musl
+  - cargo install toml-cli
 script: ./hooks/pre-push


mbestavros:
@connorkuehl Yes, good catch. That change is also present in #162 for its `Cargo.toml` test. I've replicated it here, but when one gets merged, the other shouldn't need it.

hooks/pre-push.d/rust-edition-test
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -name Cargo.toml); do
+    if [ "$(toml get "$file" package.edition)" != "\"2018\"" ]; then
```
npmccallum:
```suggestion
    if [ "$(toml get "$file" package.edition)" != '"2018"' ]; then
```

hooks/pre-push.d/rust-edition-test
```diff
@@ -0,0 +1,10 @@
...
+        echo "$file is not using edition = "2018". Please ensure the field is present and includes the correct Rust edition."
```
npmccallum: Please wrap this comment.

hooks/pre-push.d/rust-edition-test
mbestavros: Fixed in latest push.

hooks/pre-push.d/rust-edition-test
mbestavros: Fixed in latest push.

hooks/pre-push.d/cargo-edition-test
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
+top=`git rev-parse --show-toplevel`
+status=0
+for file in $(find "$top" -name Cargo.toml); do
+    if [ "$(toml get "$file" package.edition)" != '"2018"' ]; then
+        echo "$file is not using edition = "2018". Please ensure the field is\
+        present and includes the correct Rust edition."
```
npmccallum: Sorry. I didn't mean wrap the source code. I mean to wrap the output. Output should always be less than the default size of a terminal (80 chars). Being concise can also help with that.

mbestavros: I've split the output across two `echo` calls in my latest push, though the length of the filename could push it over in any case.

PR #216, opened 2020-02-07
Implement an interim sev-show

### LINKED ISSUE #205: Implement sev-show
There is a tool called `sgx-show` in the repo which checks that it is running on an appropriate environment for SGX-based workloads.

There should be one for SEV as well, likely named `sev-show`.

connorkuehl:
Closes #205.

Although it's likely this kind of platform discovery will be consolidated in the future, we should have something like `sgx-show` for SEV.

I believe I captured all of the SEV-related information described in the AMD Developer's Manual, but if I've missed some I will go back and add them.

Quick example:

```
✔ AMD CPU
  ✔ Processor is known to have microcode that supports SEV: AMD EPYC 7251 8-Core Processor                 
  ✔ SME support
    ✔ Physical address bit reduction: 5
    ✔ C-bit location in page table entry: 15
  ✔ SEV support
    ✔ Number of encrypted guests supported simultaneously: 15
    ✔ Minimum ASID value for SEV-enabled, SEV-ES disabled guest: 1
  ✔ SEV-ES support
  ✔ Page Flush MSR available
  ✔ /dev/sev device special file exists
    ✔ File is readable by user
    ✔ File is writable by user
  ✔ SEV is enabled for kernel
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "sgx-crypto",
 
     "sev",
+    "sev-show",
 ]
 
 [patch.crates-io]
--- a/sev-show/Cargo.toml
+++ b/sev-show/Cargo.toml
@@ -8,3 +8,4 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+colored = "1.9.1"
--- a/sev-show/Cargo.toml
+++ b/sev-show/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "sev-show"
+version = "0.1.0"
+authors = ["Connor Kuehl <ckuehl@redhat.com>"]
+license = "Apache-2.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
--- a/sev-show/LICENSE
+++ b/sev-show/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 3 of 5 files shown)


squidboylan:
I ran this on my desktop (ryzen 1700 cpu) and I got the following output:

```
squid@zerus:~/hacking/enarx/sev-show$ cargo run --release
    Finished release [optimized] target(s) in 0.01s
     Running `target/release/sev-show`
 ✔ AMD CPU
   ✔ SME support
     ✔ Physical address bit reduction: 5
     ✔ C-bit location in page table entry: 15
   ✔ SEV support
     ✔ Number of encrypted guests supported simultaneously: 15
     ✔ Minimum ASID value for SEV-enabled, SEV-ES disabled guest: 0
   ✗ SEV-ES support
   ✔ Page Flush MSR available
 ✗ /dev/sev device special file exists
 ✗ SEV is enabled for kernel
```

npmccallum:
@connorkuehl @squidboylan There should be at least one other thing we detect -- but I don't know what it is. Ryzen boxes don't (yet?) have firmware support for SEV. But from the output of the tool, it looks like only kernel configuration is missing. It would be good if we could detect this somehow. But, I haven't looked into it.

haraldh:
> @connorkuehl @squidboylan There should be at least one other thing we detect -- but I don't know what it is. Ryzen boxes don't (yet?) have firmware support for SEV. But from the output of the tool, it looks like only kernel configuration is missing. It would be good if we could detect this somehow. But, I haven't looked into it.

Ryzen will never have firmware support for it. Only Epyc will get it.
See:
https://github.com/AMDESE/AMDSEV/issues/1

connorkuehl:
This has been rebased and integrated with the workspace patch.

I've also added a test to see if the processor is an EPYC processor for now.

connorkuehl:
This is ready for review now.

connorkuehl:
>     1. The "errors can report strings" shouldn't be a separate commit. Add this functionality in the first patches and rebase on top of it. Then force push to GitHub.

Done!

>     2. I think testing for the existence of `/dev/sev` is rather limited. We really want to test if we can open it. So, I think you should just try to open the file. This will also report permissions problems.

I left the existence check, but added two sub-tests to that one that verify that the special file is readable and writable.

```
  ✔ /dev/sev device special file exists
    ✔ File is readable by user
    ✔ File is writable by user
```

PR #236, opened 2020-02-11
Add report module to sgx-types

lkatalin:
Here is the basic memory layout for Report. I can add functionality as needed.

### DIFF
--- a/sgx-types/src/report.rs
+++ b/sgx-types/src/report.rs
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Section 38.15
+//! The REPORT structure is the output of the EREPORT instruction, and must be 512-Byte aligned.
+
+use super::{attr::Attributes, isv, misc::MiscSelect};
+use testing::testaso;
+
+/// This struct is separated out from the Report to be usable by the Quote struct.
+/// Table 38-21
+#[derive(Default)]
+#[repr(C)]
+pub struct Body {
+    /// The security version number of the processor
+    pub cpusvn: [u8; 16],
+
+    /// Bit vector specifying which extended features are saved to the
+    /// MISC region of the SSA frame when an AEX occurs
+    pub miscselect: MiscSelect,
+
+    /// Reserved
+    reserved0: [u32; 7],
+
+    /// Attributes of the enclave (Section 38.7.1)
+    pub attributes: Attributes,
+
+    /// Value of SECS.MRENCLAVE
+    pub mrenclave: [u8; 32],
+
+    /// Reserved
+    reserved1: [u32; 8],
+
+    /// Value from SECS.MRSIGNER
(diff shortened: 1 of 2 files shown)


sgx-types/src/report.rs
```diff
@@ -0,0 +1,70 @@
...
+//! The REPORT structure is the output of the EREPORT instruction, and must be 512-Byte aligned.
+
+use super::{attr::Attributes, isv, misc::MiscSelect};
+use testing::testaso;
+
+/// This struct is separated out from the Report to be usable by the Quote struct.
+/// Table 38-21
+#[derive(Default)]
+#[repr(C)]
+pub struct ReportBody {
```
npmccallum:
```suggestion
pub struct Body {
```

sgx-types/src/report.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Data provided by the user and protected by the Report's MAC (Section 38.15.1)
+    pub reportdata: [u16; 32],
+}
+
+/// Table 38-21
+#[derive(Default)]
+#[repr(C, align(512))]
+pub struct Report {
+    /// The body of the Report
+    reportbody: ReportBody,
```
npmccallum:
```suggestion
    pub reportbody: ReportBody,
```

Is this supposed to be private?

sgx-types/src/report.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub mrsigner: [u8; 32],
+    reserved2: [u32; 24],
+    /// Product ID of the enclave
+    pub isvprodid: isv::ProdId,
+    /// Security version number of the enclave
+    pub isvsvn: isv::Svn,
+    reserved3: [u32; 15],
+    /// Data provided by the user and protected by the Report's MAC (Section 38.15.1)
+    pub reportdata: [u16; 32],
+}
```
npmccallum: I suspect we probably need padding at the end. This struct will likely be signed. So we need to make sure that any trailing bytes are properly zeroed.

sgx-types/src/report.rs
npmccallum: We don't prefix types by the module name.

sgx-types/src/report.rs
lkatalin: Very good point; I just thought of this during the standup, based on yesterday's PR.

sgx-types/src/report.rs
lkatalin: I don't think so. That was a mistake.

sgx-types/src/report.rs
npmccallum: I misread this code. It looks like `reportdata` ends at the end of the struct. So there's no padding needed here.

npmccallum:
@lkatalin Also, I personally try to avoid speaking about filesystem paths in commit messages. This can be misleading when files move around later. But this is just a nitpick.

PR #237, opened 2020-02-11
Prepraration for SGX keep merge

npmccallum:
This is a collection of changes required to merge my SGX keep. It is basically just a bunch of small improvements to existing libraries.

### DIFF
--- a/intel-types/src/lib.rs
+++ b/intel-types/src/lib.rs
@@ -10,7 +10,10 @@
 #![deny(missing_docs)]
 
 use bitflags::bitflags;
-use core::{fmt::Debug, ops::BitAnd};
+use core::{
+    fmt::Debug,
+    ops::{BitAnd, BitOr, Not},
+};
 use testing::testaso;
 
 /// Succinctly describes a masked type, e.g. masked Attributes or masked MiscSelect.
@@ -19,7 +22,7 @@ use testing::testaso;
 /// the struct and its mask for simplicity.
 #[repr(C)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct Masked<T: Copy + Debug + PartialEq + BitAnd<Output = T>> {
+pub struct Masked<T: BitAnd<Output = T>> {
     /// The data being masked, e.g. Attribute flags.
     pub data: T,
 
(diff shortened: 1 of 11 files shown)


intel-types/src/lib.rs
```diff
@@ -19,24 +22,48 @@ use testing::testaso;
...
+    T: BitAnd<Output = T>,
+    T: BitOr<Output = T>,
+    T: Not<Output = T>,
+    T: Copy,
+{
     fn from(value: T) -> Self {
         Self {
             data: value,
-            mask: value,
+            mask: value | !value,
```
lkatalin: This is bitwise or? So the mask is by default all 1's, right (nothing is masked out)?

sgx-types/src/lib.rs
```diff
@@ -9,16 +9,6 @@
...
-macro_rules! defflags {
-    ($name:ident $($value:ident)|*) => {
-        impl Default for $name {
-            fn default() -> Self {
-                $name::empty() $( | $name::$value )* | $name::empty()
-            }
-        }
-    };
-}
-
```
lkatalin: Using Defaults instead of this does seem to make sense in this crate.

sgx-types/src/tcs.rs
```diff
@@ -45,6 +45,8 @@ pub struct Tcs {
     fslimit: u32,
     /// Size to become a new GS-limit (only 32-bit enclaves).
     gslimit: u32,
+    reserved0: [u64; 23],
```
lkatalin: The "extra bytes" not being zeroed by the compiler were the bytes added to by `align(4096)` on `Tcs`? Is it expected to run into that problem? Should we always pad out these structs if there's an alignment requirement?

sgx-types/src/isv.rs
```diff
@@ -5,12 +5,12 @@
...
 /// ISV assigned Product ID.
 #[repr(transparent)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
 pub struct ProdId(u16);
 
 /// ISV assigned SVN (security version number).
 #[repr(transparent)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
```
lkatalin: I also just added this. : ) Great minds think alike....

sgx-types/src/ssa.rs
```diff
@@ -351,6 +351,17 @@ impl Default for StateSaveArea {
     }
 }
 
+impl AsRef<[u8]> for StateSaveArea {
```
lkatalin: Is there some way to guarantee in the code that we won't ever try to go in reverse?

sgx-types/src/secs.rs
```diff
@@ -16,7 +16,7 @@ pub struct Spec {
     /// The enclave size (in bytes; power of 2)
     pub enc_size: NonZeroU64,
 
-    /// The state save area frame size (in bytes)
+    /// The state save area frame size (in 4k pages)
```
lkatalin: Good catch.

sgx-crypto/src/lib.rs
```diff
@@ -40,7 +40,15 @@ impl From<&secs::Spec> for Hasher {
 
 impl Hasher {
     /// Mimics call to SGX_IOC_ENCLAVE_ADD_PAGES (EADD and EEXTEND).
-    pub fn add(&mut self, offset: u64, data: &[u8], measure: bool, secinfo: page::SecInfo) {
+    pub fn add<T: AsRef<[u8]> + ?Sized>(
```
lkatalin: I need to read about implicit bounds apparently...

intel-types/src/lib.rs
npmccallum: Basically, we need a way to say "all flags". However, `bitflags` doesn't implement any trait for this. So instead we calculate it by a bitwise or of all the flags in the value and all the flags not in the value.

sgx-types/src/tcs.rs
npmccallum: Basically, when a type is `align(4096)` the size of the type will always be a multiple of 4096. But the compiler will only ensure the actual fields are initialized, not the extra padding. The extra padding being initialized only cares when we are going to *measure* the whole size. Because, in this case, the padding's contents is undefined, the measurement result will be unpredictable.

sgx-types/src/ssa.rs
npmccallum: No. But all such methods are `unsafe`.

PR #238, opened 2020-02-11
Long live the workspace!

npmccallum:
I've come up with a way to link a binary with custom entry point on musl
libc. So I think I've solved our linking problems for now. This means we
can have the workspace back and cargo should work out of the box like
normal.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[build]
+target = "x86_64-unknown-linux-musl"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,18 @@
+[workspace]
+members = [
+    "iocuddle",
+    "iocuddle-sgx",
+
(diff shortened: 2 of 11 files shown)


npmccallum:
@connorkuehl @haraldh So my basic plan is this.

1. Make a small python script called `cc`. Store it in the root directory. This will wrap all linker invocations of `cc` and, when building the shim project, will remove all libraries linked in. We can also add or remove other linker flags as we see fit for other projects.

2. Set our custom linker via this `.cargo/config` snippit:

```toml
[build]
rustflags = [ "-C", "linker=./cc" ]
```

3. After this, build of the whole project should be a simple `cargo build`.

4. Some care will be needed for `cargo test` on binaries with custom entry points. I'm confident we can solve this problem.

npmccallum:
@connorkuehl Yes. It isn't needed until we merge a shim.

PR #248, opened 2020-02-12
Add quote module to sgx-types

### LINKED ISSUE #107: Add attestation-types crate
Types to add include:
- [ ] Quote (depends on #248 )
- [x] Report (depends on #236 )
- [x] TargetInfo ( depends on #170 )

lkatalin:
A few notes about the tricky parts of implementing the `Quote` struct:

- According to the documentation ([A.4, Table 8](https://download.01.org/intel-sgx/dcap-1.0/docs/SGX_ECDSA_QuoteGenReference_DCAP_API_Linux_1.0.pdf)),The `QEAuth` portion has 1) a `size` field and 2) a `data` [field](https://github.com/enarx/enarx/compare/master...lkatalin:quote-type?expand=1#diff-e7fe996c9d1d239623aac2a380efaa54R38) whose size is determined by the `size` field. I couldn't find a way to do this without making the size a `const`. Currently it's set to `0`, a valid value. 
   - Maybe the `data` field should be a vector, with a method that checks that the length of the vector matches the `size` field?
- It's clear that the `vendor ID` used in this `Quote` struct (`939A7233F79C4CA9940A0DB3957F0607` for Intel) is different from the `vendor ID` used in `SIGSTRUCT` (`0x8086` for Intel). However, it's not clear whether the "QE Identifier" required as part of the `user_data` field is the same as the `vendor ID`, though they are both 16 bytes. I have assumed they are the same and [copied it](https://github.com/enarx/enarx/compare/master...lkatalin:quote-type?expand=1#diff-e7fe996c9d1d239623aac2a380efaa54R191) into `user_data`.
- The endianness of integers is specified as LE; but the endianness of byte arrays is not specified. So copying the `vendor ID` into the `user_data` with `copy_from_slice` may not have the correct layout.
- The `QECert.certdata` [field](https://github.com/enarx/enarx/compare/master...lkatalin:quote-type?expand=1#diff-e7fe996c9d1d239623aac2a380efaa54R90) should be large enough to hold a concatenated PCK Cert Chain (PEM formatted): `PCK Leaf Cert||Intermediate CA Cert||Root CA Cert`. However, it's not clear that this chain will always be a predictable length. I've made the field `4096` bytes for the moment.

Fixes #107

### DIFF
--- a/attestation-types/Cargo.toml
+++ b/attestation-types/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "attestation-types"
+version = "0.1.0"
+authors = ["Nathaniel McCallum <npmccallum@redhat.com>", "Lily Sturmann <lsturman@redhat.com>"]
+license = "Apache-2.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+testing  = { path = "../testing" }
+sgx-types = { path = "../sgx-types" }
--- a/attestation-types/LICENSE
+++ b/attestation-types/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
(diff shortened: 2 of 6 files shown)


sgx-types/src/quote.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// contain a QE identifier used to link a PCK Cert to an Enc(PPID). This
+    /// identifier is consistent for every quote generated with this QE on this
+    /// platform.
+    pub user_data: [u8; 20],
+}
+
+impl Default for QuoteHeader {
+    fn default() -> Self {
+        let mut user_data = [0u8; 20];
+        user_data.copy_from_slice(&INTELVID); // TODO: Endianness of byte arrays?
```
connorkuehl: I don't think we need to worry about endianness here. Individual bytes (and I think arrays of) are endian-independent. I also think that all of Intel's chips are little endian.

lkatalin: The thing I'm concerned about is whether the bytes themselves need to be ordered as BE or LE, because I didn't see that documented anywhere. I just wouldn't want to record the number backwards. Then again, I think `user_data` is the most flexible field here, so maybe it is fine. It depends on whether this needs to match anything that we don't control, and I can't remember right now whether that's the case.

npmccallum:
> A few notes about the tricky parts of implementing the `Quote` struct:
> 
> * According to the documentation ([A.4, Table 8](https://download.01.org/intel-sgx/dcap-1.0/docs/SGX_ECDSA_QuoteGenReference_DCAP_API_Linux_1.0.pdf)),The `QEAuth` portion has 1) a `size` field and 2) a `data` [field](https://github.com/enarx/enarx/compare/master...lkatalin:quote-type?expand=1#diff-e7fe996c9d1d239623aac2a380efaa54R38) whose size is determined by the `size` field. I couldn't find a way to do this without making the size a `const`. Currently it's set to `0`, a valid value.

First, let me give you some Rust supported options: https://doc.rust-lang.org/nomicon/exotic-sizes.html

However, I find the use of `QEAuth` to be very strange. In C, this idiom is called flexible arrays. You basically have an unsized array at the end of the struct. Then you can vary the size of the struct at run time. But `QEAuth` is used in the middle of a struct here: https://github.com/enarx/enarx/pull/248/files#diff-e7fe996c9d1d239623aac2a380efaa54R137 . That's not even legal in C! So first let me ask, are you sure this is correct? Because if they are actually doing a flexible array, then we can do something similar in Rust.

If the code is correct as you've written, we're probably better off to parse this into a `Vec`.

sgx-types/src/quote.rs
npmccallum: @lkatalin I'm with @connorkuehl here. I don't think we need to worry about it. If we find something wrong, we'll discover it later.

lkatalin: @connorkuehl @npmccallum I was thinking of something like this: https://doc.rust-lang.org/std/primitive.u32.html#method.to_le_bytes But I am more than happy to drop it and not worry about it.

lkatalin:
> However, I find the use of `QEAuth` to be very strange.... So first let me ask, are you sure this is correct?

@npmccallum I agree, the `QEAuth` and `QECert` data types are deeply strange. They are both of variable size, as documented in A.4, Table 4 on page 22 in the [doc](https://download.01.org/intel-sgx/dcap-1.0/docs/SGX_ECDSA_QuoteGenReference_DCAP_API_Linux_1.0.pdf). So basically it's unclear where `QECert` should begin. For now, I can change `QECert` to an enum and a `Vec`, and `QEAuth` to just a `Vec` of currently zero size.

lkatalin:
@npmccallum I pushed a new version here. There is a commit to move attestation types to their own crate, and then another commit to add the Quote to these types. I thought about making the `const` values into `structs` with a `default`, but in the end wasn't sure which would be better.

PR #249, opened 2020-02-13
Add some basic docs for AMD and IBM attestation

npmccallum:
This is incomplete, but is necessary for some vendor conversations. We
expect to fill these in and improve them.


MikeCamel:
Spoke to IBM - they are working on this, and hope to have an update soon.

PR #261, opened 2020-02-17
Add certificate chain  and process diagram to Intel SGX documentation

lkatalin:
Re: the SGX certificate chain diagram. 

- Do we want to include the TCB Signing Cert? AFAIK we don't use this (currently).
- The key images are rather large, in order to fit the text. I could shorten to abbreviations, ex. "PCK" instead of "Provisioning Certification Key" if that would be preferable.

Re: the SGX process diagram:
I tried to make it relevant to Enarx without making it Enarx-specific, ex. it doesn't have "Enarx host agent" or "Enarx client agent" components. There are many ways to do attestation with SGX, so I picked a way that closely resembles what we are doing. It's more or less a draft, so I can add more detail where warranted.

Fixes #261


docs/attestation/intel/sgx/sgx_process.msc
```diff
@@ -0,0 +1,31 @@
+# Commits which modify this file MUST generate the new .png!
+# .png files can be generated at https://mscgen.js.org/
```
connorkuehl, line 2: Our automation will enforce this rule if appropriate entries are added to [docs/Makefile](https://github.com/enarx/enarx/blob/master/docs/Makefile#L1).

The automated script depends on this relationship between the markup and the diagram filename:

PNG: `docs/attestation/intel/sgx/sgx_certchain.msc.png`
MSC: `docs/attestation/intel/sgx/sgx_certchain.msc`

so for it to work for these diagrams out of the box you'll need to change the file extension of the png you checked in to end with `.msc.png` instead of just `.png`

lkatalin: Thanks for pointing this out, @connorkuehl . I changed the file names and included them in the doc Makefile.

lkatalin: Actually, the old ones are still here for some reason even though I did `git rm`. Will fix before I ping for re-review.

docs/Makefile
```diff
@@ -3,7 +3,9 @@ DIAGRAMS := \
 	attestation/amd/sev/certchain.dot.png \
 	attestation/amd/sev/process.msc.png \
 	attestation/ibm/pef/certchain.dot.png \
-	attestation/ibm/pef/process.msc.png
+	attestation/ibm/pef/process.msc.png \
+	attestation/intel/sgx/sgx_certchain.dot.png \
+	attestation/intel/sgx/sgx_process.msc.png
```
connorkuehl, line 8: This Makefile was replaced by a cargo-make Makefile.

These entries now belong in `docs/Makefile.toml` and the notable difference here is that now the final rendered image format extension is omitted from the initial list.

haraldh (approved):

connorkuehl (dismissed):

PR #266, opened 2020-02-18
Add a walkthrough of AMD SEV attestation to the documentation

### LINKED ISSUE #253: Improve SEV attestation docs
Currently, there is just two flow charts in PR #249. We need text describing the process.

connorkuehl:
Closes #253.


connorkuehl:
@npmccallum and I met to take a first pass at this to help shape it a bit more.

The following topics came up that I'm aiming to fix in the next revision

- [ ] This document conflates the term "tenant" and "guest" and refers to it as one entity. This is incorrect. A tenant is an entity that wants to create a guest. A guest is the virtual machine.
- [ ] SEV-ES and SEV-SNP are to be documented at a later date, so a note should be left saying such.
- [ ] Factor out bits of information that are essential pre-requisite knowledge for understanding the attestation process and post this in something like `prerequisites.md` and point to it at the beginning of the document.
- [ ] Add a brief overview that explains the 4 parts of attestation in simple language
- [ ] Add tables to the descriptions of each packet that is transmitted between the tenant and the guest that denotes what is plaintext and what is ciphertext. What is integrity-protected and what is not.
- [ ] Mention that a "self-owned" OCA keypair is the default state.
- [ ] Amend the paragraph that explains the PDH key to indicate that it is only used to create a master key derived from a shared secret between the tenant and the platform.
- [ ] Clarify the relationship between the PSP and the firmware
- [ ] Note a host can *always* launch a guest. A tenant will not provide secrets to that guest unless attestation completes.
- [ ] The PSP only has part of the certificate chain. The CEK is initially unsigned. Host is expected to download the public key components of the ARK, ASK, and it will make a POST request to an AMD-owned service to have the CEK signed.
- [ ] ~~Produces~~ -> use "calculates" or "derives" for a cryptographic component that is emitted from a function; use "generates" for data that is created from a secure source of entropy.
- [ ] Add the labels to certain derivations, for example, the master key is derived from the shared secret, the nonce, *and the string "master-secret"*.
- [ ] There are some minor grammatical errors that will be corrected throughout.

npmccallum:
Fantastic docs! Great work! :tada:

PR #278, opened 2020-02-21
Ensure diagrams are always up-to-date

### LINKED ISSUE #256: Ensure that changed process flow diagrams always include an updated .png
@npmccallum merged some neat new process flow diagrams in #249. These diagrams, described as `.msc` and `.dot` files, generate a corresponding `.png`.

We should implement a test that checks if a commit has changes to `.dot` or `.msc` files and enforces that such changes must also contain an updated `.png` to ensure these are always up to date.

connorkuehl:
Closes #256 

This pull request adds a test to our GitHub Actions suite to confirm that the diagrams in a pull request are always up-to-date.

This ended up being a little more complex than I initially thought, due to how certain approaches made the test quite brittle.

**Idea:** just compare the (c/m)times of the files
**Why doesn't it work:** Git doesn't necessarily track this ([and after Googling why](https://stackoverflow.com/questions/1964470/whats-the-equivalent-of-use-commit-times-for-git/1964508#1964508), I tend to agree with that decision). So, when our CI clones the repo to check out the files, the (c/m)times are simply the time of the clone, and if all the times are the same, then this test wouldn't make sense.

**Idea:** force `make` to regenerate the diagrams, and if the binary output changes, then obviously the diagram sources have changed and the contributor forgot to commit them!
**Why doesn't it work:** It kind of worked! However, when `dot` generates the .PNGs, the binary output changes each time even if the source remains the same. It must be putting a timestamp somewhere, so it would fail every time. When I switched the file types to .svg, this approach _did_ work; however, that is not the form this pull request has taken, as I do not want to tightly couple the test to the image format of our diagrams.

Finally, here's the approach I _did_ take for this PR:

**Idea:** Query the Makefile for the diagrams that _should_ be tested (this allows us to avoid globbing for file extensions or duplicating the list of files somewhere else). If the commit date of the diagram is older than its source, then the test fails. The commit date is converted to UTC and is the seconds past the epoch.

### DIFF
--- a/.github/workflows/miscellaneous.yml
+++ b/.github/workflows/miscellaneous.yml
@@ -26,3 +26,12 @@ jobs:
     # Ensure all .rs files have an Apache 2.0 SPDX license ID.
     - name: All .rs files have SPDX IDs
       run: .tests/misc-licenses-rs-spdx
+
+  misc-diagrams:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    # Ensure all diagrams are up-to-date.
+    - name: All diagrams are up-to-date
+      run: .tests/misc-diagrams
--- a/.tests/misc-diagrams
+++ b/.tests/misc-diagrams
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+diagram_dir="./docs/attestation/"
+diagrams=$(make --quiet -C "$diagram_dir" diagrams)
+
+status=0
+for diagram in $diagrams ; do
+	# Truncate the file extension (.png, .svg, etc)
+	source=${diagram%????}
+
+	source_date=$(git log -1 --date=unix --format="%cd" -- "$diagram_dir$source")
+	diagram_date=$(git log -1 --date=unix --format="%cd" -- "$diagram_dir$diagram")
+
+	if [[ $diagram_date -lt $source_date ]]; then
+		echo "This diagram is out-of-date: $diagram_dir$diagram. Please generate an"`
+		      `" updated diagram and commit it to include it with your pull request."
+		status=1
+	fi
+done
+exit $status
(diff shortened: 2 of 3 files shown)


.github/workflows/miscellaneous.yml
```diff
@@ -26,3 +26,12 @@ jobs:
...
       run: .tests/misc-licenses-rs-spdx
+
+  misc-diagrams:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    # Ensure all diagrams are up-to-date.
+    - name: All diagrams are up-to-date
+      run: .tests/check-diagrams
```
mbestavros: Rule of thumb we've settled on for tests: the filename should match the test name in the workflow. `check-diagrams` should be renamed to `misc-diagrams`.

If you anticipate that we'll add more diagram tests, you can name them more specifically (see how I subcategorized our licensing tests above this).

.tests/check-diagrams
```diff
@@ -0,0 +1,19 @@
...
+status=0
+for diagram in $diagrams ; do
+	# Truncate the file extension (.png, .svg, etc)
+	source=${diagram%????}
+
+	source_date=$(git log -1 --date=unix --format="%cd" -- "$diagram_dir$source")
+	diagram_date=$(git log -1 --date=unix --format="%cd" -- "$diagram_dir$diagram")
+
+	if [[ $diagram_date -lt $source_date ]]; then
+		echo "$diagram_dir$diagram"
```
mbestavros: Near as I can tell, this just emits the file path of any offending diagrams. Could you add a brief explanation of the problem and what needs to be done to fix it? That would be more helpful when a developer encounters a test failure.

[Example](https://github.com/enarx/enarx/blob/fbaa92856f4b6b275ee72e492902221f92f5ae02/.tests/misc-eof#L8) of what I mean.

PR #287, opened 2020-02-26
Test for SPDX ID headers in assembly files

### LINKED ISSUE #276: Test for SPDX ID headers in assembly files
This should be a fairly straightforward adaptation of the current SPDX ID test; it just needs to check for headers in `.s` and `.S` files in addition to whatever it's doing now.

mbestavros:
Resolves #276.

### DIFF
--- a/.tests/misc-licenses-asm-spdx
+++ b/.tests/misc-licenses-asm-spdx
@@ -1,0 +1,12 @@
+#!/bin/bash -e
+set -o pipefail
+
+top=$(git rev-parse --show-toplevel)
+status=0
+while read -r file; do
+    if ! awk '# SPDX-License-Identifier: Apache-2.0$/ { exit 0 } { exit 1 }' "$file"; then
+        echo "$file does not have an Apache 2.0 SPDX ID on the first line of the file."
+        status=1
+    fi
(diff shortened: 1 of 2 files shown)


npmccallum:
@mbestavros The earlier statement of `.c` files was an error. We need `.s` **and** `.S`. `.S` files are the same as `.s` except that the compiler runs the C pre-processor on them.

.tests/misc-licenses-asm-spdx
```diff
@@ -0,0 +1,12 @@
...
+set -o pipefail
+
+top=$(git rev-parse --show-toplevel)
+status=0
+while read -r file; do
+    if ! awk '# SPDX-License-Identifier: Apache-2.0$/ { exit 0 } { exit 1 }' "$file"; then
+        echo "$file does not have an Apache 2.0 SPDX ID on the first line of the file."
+        status=1
+    fi
+done < <(find "$top" -regex '.*/target' -prune -o -type f -regex '.*\.s' -print)
```
npmccallum: We need `.S` files too.

mbestavros: Done in latest push.

PR #290, opened 2020-02-28
sev: Porting over the sev functionality from ketuvim

### LINKED ISSUE #180: Port sev functionality from ketuvim to enarx/sev
There are plans for api common code in ketuvim to be merged with rust-vmm. The sev functionality needs to be moved and ported to the SEV crate in enarx.

whitebrandy:
1. vm and fd in struct Launch are changed to AsRawFd type

2. I changed the ioctl interface to something we have in sgx-sys/src/ioctl/mod.rs
   I moved enum Code and struct Command from sev.rs to ioctl/run.rs
   Then I created a new struct EncryptOp which takes struct Command and casts
   it as u64.

3. I changed the implementation of Launch to match the new format where
   vm and fw are set to AsRawFd

Related: #180 

Signed-off-by: Bandan Das <bsd@redhat.com>

Resolves #180.

### DIFF
--- a/sev/src/firmware/linux/ioctl/wrapper.rs
+++ b/sev/src/firmware/linux/ioctl/wrapper.rs
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::convert::TryInto;
+use std::io::Error;
+use std::marker::PhantomData;
+use std::os::raw::{c_int, c_uint, c_ulong, c_void};
+use std::os::unix::io::AsRawFd;
+use std::ptr::null;
+
+extern "C" {
+    pub fn ioctl(fd: c_int, request: c_ulong, ...) -> c_int;
+}
+
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum Code {
+    Init = 0,
+    EsInit,
+
+    LaunchStart,
+    LaunchUpdateData,
+    LaunchUpdateVmsa,
+    LaunchSecret,
+    LaunchMeasure,
+    LaunchFinish,
+
+    SendStart,
+    SendUpdateData,
+    SendUpdateVmsa,
+    SendFinish,
+
+    ReceiveStart,
+    ReceiveUpdateData,
+    ReceiveUpdateVmsa,
+    ReceiveFinish,
+
+    GuestStatus,
+    DebugDecrypt,
+    DebugEncrypt,
+    CertExport,
+}
+
+#[repr(C)]
+pub struct Command {
+    pub code: Code,
+    pub data: u64,
+    pub error: u32,
+    pub fd: u32,
+}
+
+pub trait Ioctl {
+    const REQUEST: c_ulong;
+
+    fn ioctl(&self, fd: &impl AsRawFd) -> std::io::Result<c_uint> {
+        let r = unsafe {
+            ioctl(
+                fd.as_raw_fd(),
+                Self::REQUEST,
+                self as *const _,
+                null::<c_void>(),
+            )
+        };
+
+        r.try_into().or_else(|_| Err(Error::last_os_error()))
+    }
+}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct EncryptOp<'a>(u64, PhantomData<&'a ()>);
+
+impl<'a> Ioctl for EncryptOp<'a> {
+    const REQUEST: c_ulong = 3221794490;
+}
+
+impl<'a> EncryptOp<'a> {
+    pub fn new(cmd: &'a Command) -> Self {
+        EncryptOp(cmd as *const _ as _, PhantomData)
+    }
+}
--- a/sev/src/firmware/linux/mod.rs
+++ b/sev/src/firmware/linux/mod.rs
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub mod ioctl;
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum State {
+    Uninitialized,
+    Initialized,
+    Working,
+}
--- a/sev/src/firmware/linux/ioctl.rs
+++ b/sev/src/firmware/linux/ioctl.rs
@@ -5,8 +5,8 @@ use std::mem::{size_of_val, MaybeUninit};
 use std::os::raw::{c_int, c_ulong};
 use std::os::unix::io::AsRawFd;
 
-use super::*;
 use crate::certs::sev::Certificate;
+use crate::firmware::*;
 
 #[repr(u32)]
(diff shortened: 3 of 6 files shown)


sev/src/sev.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    pub fn measure(self) -> Result<Launch<'a, Measured, F, V>> {
+        #[repr(C)]
+        struct Data {
+            addr: u64,
+            size: u32,
+        }
+
+        //let mut measurement: launch::Measurement = unsafe { MaybeUninit::uninit().assume_init() };
```
connorkuehl: Let's delete this comment.

whitebrandy:
> Thanks, @whitebrandy! Congrats on your first Rust PR (if I remember correctly from standup this morning) tada
...
> I also think that the commands that are in `sev/src/sev.rs` should be moved to `sev/src/firmware/linux.rs`.
IIUC, the right place for the sev impl blocks is actually sev/src/firmware/linux.rs ? Is it ok if I make that change in a separate task and do the other suggested changes in this PR ?

connorkuehl:
@whitebrandy I'm really sorry for being the blocker on this for so long. Since I've had to adjust my priorities, it no longer makes sense for you to have to wait on me to finish the `iocuddle-sev` PR (#202) and base this PR on that.

I think the best way forward is to adapt this PR to something similar to what is done in #202 and I will rebase my PR against yours.

This means (and you can probably just cargo-cult copy whatever you find helpful from #202):

1. Create a crate called `iocuddle-sev`. This crate will provide type-safe ioctls for the SEV platform, much like `iocuddle-sgx` does.
1. Move the command buffer types used by your ioctls to `sev-types/src/command/mod.rs`
1. Add a [generic `Command` struct](https://github.com/enarx/enarx/pull/202/commits/1efb36d4586620eb2a4e68ca5b40ea780a0a9033#diff-2f5cab92b5c6bf8ba4ce63a9ef142fceR79) that encloses the `ioctl` commands.
1. Finally, [declare your new type-safe ioctls](https://github.com/enarx/enarx/pull/202/commits/1efb36d4586620eb2a4e68ca5b40ea780a0a9033#diff-2f5cab92b5c6bf8ba4ce63a9ef142fceR27) as public constants so that other crates can safely invoke your new ioctls.

There is a key acceptance criteria that is not currently met by #202 and that has to do with pointer lifetimes, so I will do my best to explain what we want here as I have no example to point you to. Because the SEV ioctls basically provide a pointer to the data buffer, we want to make sure that the memory address used has the same lifetime as the command buffer structs we are defining.

For example, [this struct](https://github.com/enarx/enarx/pull/202/commits/7c3ef5d2d1c2390591777d361239bbbaafc6c3f7#diff-e5104af3982342f87cd5e3dfbec1a9f9R83). The prescribed approach is to make its constructor function _not_ take the memory address outright as a `u64` but to instead take a reference to the data buffer with lifetime `&'a` and perform the cast to the `u64` in the constructor function and store the memory address `u64` in the struct there. Right now, my code takes the already-cast `u64` when instead it should be taking a `&'a Cert` and casting it to an address inside the constructor. This would signal to the compiler that the command buffer _must_ enclose the same lifetime as the `Cert` object and therefore we guarantee that the address is valid for the entire lifetime of the ioctl call.

Like I said, #202 does _not_ do that right now but you're not on your own, please don't hesitate to ask questions on any of the above.

sev/src/firmware/linux/ioctl/wrapper.rs
```diff
@@ -0,0 +1,80 @@
...
+
+extern "C" {
+    pub fn ioctl(fd: c_int, request: c_ulong, ...) -> c_int;
+}
+
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum Code {
+    Init = 0,
+    EsInit,
```
connorkuehl, line 18: ~~EsInit~~ -> InitEx

sev/src/firmware/linux/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub mod ioctl;
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum State {
+    Uninitialized,
+    Initialized,
+    Working,
+}
```
connorkuehl, line 11: Why not use the definition of this enum in `sev/src/firmware/mod.rs`?

sev/src/firmware/linux/ioctl/wrapper.rs
whitebrandy: Since these are ioctls to /dev/kvm, it might be a good idea for the names to be similar. For example, this is KVM_SEV_ES_INIT in kvm.h. What do you think ?

whitebrandy:
> Finally, my last note would be to break this PR down into smaller PRs and land just 1 of these ioctl

@connorkuehl  Sorry, a lot of this is already what you have pointed out before! :) I thought getting this in would be the first step to breaking it down and adding the other changes. Anyway, this looks like a plan too. Any suggestions for how to to break this down for the smaller PRs ?

sev/src/firmware/linux/mod.rs
whitebrandy: Sounds good, thanks for pointing out!

sev/src/firmware/linux/ioctl/wrapper.rs
connorkuehl: Ahh I'm torn now. On one hand I prefer `InitEx` because it aligns with the SEV API spec and I don't want to deviate from it just because the kernel does...

On the other hand, leaving it as is is technically more consistent with the interface we're using (the one exported by the kernel).

Let's leave it as `EsInit`, but please leave a comment saying `/* AKA InitEx in the SEV spec */`

connorkuehl:
> Any suggestions for how to to break this down for the smaller PRs ?

@whitebrandy The strategy I'm following is to [open a PR with just 1 of the ioctls](https://github.com/enarx/enarx/pull/516) (in fact, that's the PR I've been linking to) just so I know that I understand how to implement the ioctl as an iocuddle-based ioctl. That way with just 1 ioctl you  can get the "recipe" down but have a quicker feedback loop without having to update dozens of ioctls in between each revision :-)

npmccallum:
@connorkuehl What is the state of this regarding our decision to re-merge the `sev` crate?

connorkuehl:
> What is the state of this regarding our decision to re-merge the `sev` crate?

@npmccallum the final revision of this PR will exist as a module somewhere under `firmware/`, likely as a submodule to the `linux` module.

connorkuehl:
I spoke with @whitebrandy and we agreed that I'll carry this work forward because his primary responsibilities are taking up a lot of his bandwidth. I've saved the patches from this PR and I'm closing this PR until I have something ready for review on a branch that I can push to for the review process.

PR #303, opened 2020-03-08
Various small fixups

npmccallum:


npmccallum:
@mbestavros Why is Mergeable blocking my merge?

PR #305, opened 2020-03-09
Fix guidance comment formatting in conferences template

axelsimon:
And turn more things into bullet points.

### DIFF
--- a/.github/ISSUE_TEMPLATE/conferences.md
+++ b/.github/ISSUE_TEMPLATE/conferences.md
@@ -7,11 +7,11 @@ assignees: ''
 
 ---
 **Info**
-Conference / Event URL
-Any other relevant information
+- Conference / Event URL
+- Any other relevant information
 
 **Dates**
-//Some of these are optional, enter what you can. If the event is over several days, use the YYYY-MM-DD/YYYY-MM-DD date format to represent the range.
+<!-- Some of these are optional, enter what you can. If the event is over several days, use the YYYY-MM-DD/YYYY-MM-DD date format to represent the range. -->
 - CFP opens:     YYYY-MM-DD
 - CFP closes:    YYYY-MM-DD
 - CFP results:   YYYY-MM-DD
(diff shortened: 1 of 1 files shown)


axelsimon:
Further to this, as we've already mentioned before, it doesn't really make sense to run most (all?) of these tests on things such as issue templates.
It's probably a safe bet to exclude `.github` or at least `.github/ISSUE_TEMPLATE` from the tests.

I'll try to force-push to see if it goes through this time.

mbestavros:
> Further to this, as we've already mentioned before, it doesn't really make sense to run most (all?) of these tests on things such as issue templates.
> It's probably a safe bet to exclude `.github` or at least `.github/ISSUE_TEMPLATE` from the tests.

I think this was more of an issue when we were running tests pre-push, and tests needed to pass in order to even get your stuff up to the cloud -- which wasted time.

While I don't disagree with your comment @axelsimon, I think it's much less of an issue now (barring freak outages like this).

axelsimon:
I'm also considering this from a frugality perspective: there is little point in using CPU time and by extension energy for tests that are unrelated to the changes at play. Automating testing is a great thing for ensuring quality over time, but it should be balanced with the need to spin up resources all the time.

Two more things. I mistakenly created a branch on the Enarx repo and not in my fork. So i'd rather not (force) push any further to the Enarx repo. But the good news is that there's simply a [rerun jobs](https://github.com/enarx/enarx/pull/305/checks?check_run_id=495563604) button on the "checks" page, which makes force pushing unnecessary. The bad news is that `misc-diagrams` is still failing. There are no diagrams here, i suggest we rebase and merge and be done with it.
Do you agree?

mbestavros:
I just realized "rerun jobs" exists. Neat!

Having done that, it looks like `misc-diagrams` passes now. So I think we're good to go.

And yes -- agreed again. We shouldn't be spinning up resources we don't need to be. It's a question of how we prioritize it.

PR #306, opened 2020-03-09
SGX now handles exceptions!

npmccallum:


npmccallum:
Big thanks to @MWShan for walking me through the exception flow.

enarx-keep-sgx/src/component.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        // Parse the file.
+        let elf = Elf::parse(data.as_ref()).unwrap();
+
+        // Validate identity assumptions.
+        assert_eq!(elf.header.e_ident[EI_CLASS], ELFCLASS64);
+        assert_eq!(elf.header.e_ident[EI_DATA], ELFDATA2LSB);
+        assert_eq!(elf.header.e_ident[EI_VERSION], EV_CURRENT);
+        //assert_eq!(elf.header.e_ident[EI_OSABI], ELFOSABI_NONE);
+        //assert_eq!(elf.header.e_ident[EI_ABIVERSION], 0);
```
connorkuehl: Let's drop these comments and if they're still important to us let's file a `debt` issue and link to it in a comment here.

sgx-crypto/src/lib.rs
```diff
@@ -205,7 +196,7 @@ mod test {
             off += i.0.len();
         }
 
-        hasher.finish()
+        hasher.finish(None).mrenclave()
```
lkatalin:
```suggestion
        // Use default signature parameters
        hasher.finish(None).mrenclave()
```

enarx-keep-sgx/src/component.rs
npmccallum: https://github.com/enarx/enarx/issues/308

npmccallum: @connorkuehl Done.

sgx-types/src/sig.rs
```diff
@@ -61,58 +61,88 @@ pub struct Author {
...
-        attr: Masked<Attributes>,
-        mrenclave: [u8; 32],
-        isv_prod_id: isv::ProdId,
-        isv_svn: isv::Svn,
-    ) -> Self {
-        Self {
-            misc,
+impl Parameters {
+    /// Combines the parameters and a hash of the enclave to produce a `Measurement`
+    pub const fn measure(&self, mrenclave: [u8; 32]) -> Measurement {
```
lkatalin: Just a comment: calling this `measure(...)` is slightly confusing to me since the actual measurement `mrenclave` is an input and so the function is not really "measuring". I would have expected `measure(...)` to call something that produces `mrenclave` as output. Probably can ignore this for now, though.

sgx-types/src/ssa.rs
```diff
@@ -8,62 +8,36 @@
...
 
 /// Section 38.9.1.1, Table 38-9
-#[derive(Copy, Clone, Debug)]
-pub enum ExitType {
-    /// Hardware exit.
-    Hardware,
-
-    /// Software exit.
-    Software,
+#[repr(C, align(4))]
```
lkatalin: Could use some documentation for this struct.

npmccallum: I'm not sure what you're looking for here. This isn't a new struct. It is just a rework of the existing one. I would like better documentation here, but I'm still not 100% clear on how it is used yet. However, we're now at the place in the SGX code where we can start actually handling these errors. So we'll know shortly.

sgx-types/src/sig.rs
npmccallum: Honestly, I had the same thoughts. But I couldn't come up with a better name. Any suggestions?

sgx-types/src/ssa.rs
lkatalin: I was just thinking about it because I had to consult the documentation about the `valid` field. But I can create an issue for this and do it as a separate PR.

sgx-types/src/sig.rs
lkatalin: `measurement_wrapper`? `measurement_constructor`? `measurement_interface`? It's hard to think of something succinct. `include_measurement`? `add_mrenclave`?

lkatalin: Though, we don't need to get stuck on this if all of those are unappealing.

crt0stack/Cargo.toml
```diff
@@ -23,3 +23,4 @@ keywords = [
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+enumerate = { path = "../enumerate" }
```
haraldh: Which means, it's not a standalone crate anymore, which can be published independently. :disappointed:

enarx-keep-sgx/src/enclave.s
```diff
@@ -0,0 +1,66 @@
...
+
+	push	%r13
+	.cfi_adjust_cfa_offset	8
+	.cfi_rel_offset		%r13, 0
+
+	push	%r12
+	.cfi_adjust_cfa_offset	8
+	.cfi_rel_offset		%r12, 0
+
+    mov     0x48(%rsp), %eax    # Leaf
```
haraldh: spaces and tabs mixed in the file?

sgx-types/src/sig.rs
npmccallum: @lkatalin How about we leave it as it is for now and file a debt issue?

crt0stack/Cargo.toml
npmccallum: Why can't we just publish `enumerate`?

enarx-keep-sgx/src/enclave.s
npmccallum: @haraldh Argh! I'll fix that... :(

sgx-crypto/src/lib.rs
npmccallum: Fixed.

sgx-types/src/sig.rs
npmccallum: Never mind. Fixed.

enarx-keep-sgx/src/enclave.s
npmccallum: Fixed.

npmccallum:
@connorkuehl @lkatalin @haraldh I believe all issues have been addressed. All tests pass.

crt0stack/Cargo.toml
haraldh: yay! Do it! I can use it in a lot of places :)

PR #310, opened 2020-03-11
Create sev-types crate

connorkuehl:
We have a lot of SEV work in flight that would benefit from a centrally shared, iteratively developed package full of type-safe SEV types.

Let's add one to avoid duplicating types across SEV work items.

Just for inspiration of how things can be arranged as people add to this: I'm currently sitting on a much larger branch full of other types to merge, but my module structure is organized like this (this can be subject to change as types are added):

```
src/
├── command
│   ├── buffer.rs
│   └── mod.rs
├── guest.rs
├── lib.rs
└── platform
    ├── mod.rs
    └── x86.rs
```

### DIFF
--- a/sev-types/src/lib.rs
+++ b/sev-types/src/lib.rs
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `sev-types` crate contains types that are useful with interacting with
+//! a SEV-enabled environment.
+//!
+//! These types are described in the ["Secure Encrypted Virtualization API"](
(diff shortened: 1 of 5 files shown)


sev-types/src/lib.rs
```diff
@@ -17,3 +17,5 @@
 //! date; the link to the SEV API specification.
 
 #![no_std]
```
npmccallum:
```suggestion
#![no_std]

#![deny(missing_docs)]
#![deny(clippy::all)]
```

PR #314, opened 2020-03-11
Add SEV API command codes (AKA command IDs)

connorkuehl:
These are embedded in the command struct that the SEV API expects so that the firmware can know what action to take and how it should interpret the contents of the command buffer.

`src/command/mod.rs` is basically empty but will contain types and other submodules in a future patch set.

These will be particularly useful for a future PR that I want to base `iocuddle-sev` on.

### DIFF
--- a/sev-types/Cargo.toml
+++ b/sev-types/Cargo.toml
@@ -8,3 +8,5 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+enumerate = { path = "../enumerate" }
+
--- a/sev-types/src/command/code.rs
+++ b/sev-types/src/command/code.rs
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `code` module contains convenient symbolic descriptions
+//! of the possible commands that may be made to the SEV platform.
+//!
+//! Chapter 4.3, "Command Identifiers"
+
+enumerate::enumerate! {
+    /// Symbolic representation of the command codes exported by the
+    /// SEV API. These enums map directly to an integral value that the
+    /// SEV API calls an "ID."
+    #[derive(Copy, Clone)]
+    pub enum Code: u32 {
+        /// Initialize the platform.
+        Init = 0x001,
+
+        /// Initialize the platform with extended parameters.
+        InitEx = 0x00d,
+
+        /// Shut the platform down.
+        Shutdown = 0x002,
+
+        /// Delete the persistent platform state.
+        PlatformReset = 0x003,
+
+        /// Query the platform status.
+        PlatformStatus = 0x004,
+
+        /// Generate a new PEK.
+        PekGen = 0x005,
+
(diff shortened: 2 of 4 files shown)


connorkuehl:
> Using the `enumerate` crate (about to merge) would save a lot of code here.

@npmccallum Done!

sev-types/src/command/code.rs
```diff
@@ -0,0 +1,123 @@
...
+
+enumerate! {
+    /// Symbolic representation of the command codes exported by the
+    /// SEV API. These enums map directly to an integral value that the
+    /// SEV API calls an "ID."
+    #[derive(Copy, Clone)]
+    pub enum Code: u32 {
+
+    /// Initialize the platform.
+    Init = 0x001,
```
npmccallum: No empty line before the first variant. Also, indentation is wrong for all variants.

sev-types/Cargo.toml
```diff
@@ -8,3 +8,4 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+enumerate = { path = "../enumerate" }
```
npmccallum: Needs newline. @mbestavros I thought we had a test for this. Am I wrong?

npmccallum:
@connorkuehl Also, I haven't verified all the enumeration variant values. So I'm trusting you. :)

sev-types/src/command/code.rs
connorkuehl: I agree, but `cargo fmt` does not.

sev-types/Cargo.toml
connorkuehl: I ran into an issue with this as well, CI will complain that `cargo fmt` fails. `cargo fmt` does not want a new line.

mbestavros: We do, but... it's somehow not triggering?

We use `tail -c 1` and fail if we see anything, and this file behaves the same way as ones that _did_ fail correctly. Very strange.

sev-types/src/command/code.rs
npmccallum: Hrm... `cargo fmt` doesn't yell at me for this in `intel-types::Exception`...

connorkuehl:
@mbestavros @npmccallum I've updated the PR to uphold the letter of the team coding style to help us troubleshoot what the CI is/is not missing.

sev-types/src/command/code.rs
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `code` module contains convenient symbolic descriptions
+//! of the possible commands that may be made to the SEV platform.
+//!
+//! Chapter 4.3, "Command Identifiers"
+
+use enumerate::enumerate;
+
+enumerate! {
```
npmccallum:
```suggestion
enumerate::enumerate! {
```

Then remove the `use` statement above.

npmccallum:
@mbestavros Why are tests failing?

PR #316, opened 2020-03-11
Test that Rust lints are enabled

### LINKED ISSUE #153: Test which checks for lints

mbestavros:
Resolves #153.

### DIFF
--- a/.tests/misc-lints-missing-docs
+++ b/.tests/misc-lints-missing-docs
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 top=$(git rev-parse --show-toplevel)
 echo "The following files do not #![deny(missing_docs)]:"
-! grep -LF '#![deny(missing_docs)]' $top/*/src/main.rs $top/*/src/lib.rs
+! git grep -LF '#![deny(missing_docs)]' $top/*/src/main.rs $top/*/src/lib.rs
--- a/.tests/misc-lints-missing-docs
+++ b/.tests/misc-lints-missing-docs
@@ -1,0 +1,4 @@
+#!/bin/bash -e
+top=$(git rev-parse --show-toplevel)
+echo "The following files do not #![deny(missing_docs)]:"
+! grep -LF '#![deny(missing_docs)]' $top/*/src/main.rs $top/*/src/lib.rs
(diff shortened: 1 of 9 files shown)


npmccallum:
In order to make tests pass, my suggestion is to do this:

1. Add the lints to all crates that are missing them.
2. Any crates that currently fail those lints should be fixed if the fix is trivial.
3. Cases where fixes aren't trivial (usually docs), use the below grandfather technique.
4. File a `debt` issue for every grandfathered crate.

## Grandfather Technique

```rust
#![deny(missing_docs)]
#![allow(missing_docs)]
```

npmccallum:
@mbestavros You also need to create a debt issue for each grandfathered crate.

mbestavros:
Debt issues filed for all missing lints here:

https://github.com/enarx/enarx/issues/343
https://github.com/enarx/enarx/issues/344
https://github.com/enarx/enarx/issues/345
https://github.com/enarx/enarx/issues/346
https://github.com/enarx/enarx/issues/347
https://github.com/enarx/enarx/issues/348
https://github.com/enarx/enarx/issues/349

.tests/misc-lints-missing-docs
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+top=$(git rev-parse --show-toplevel)
+status=0
+for f in $top/*/src/main.rs $top/*/src/lib.rs; do
+    if [[ -f "$f" && ! $(grep -F '#![deny(missing_docs)]' "$f") ]]; then
+        echo "$f is missing lint: '#![deny(missing_docs)]'"
+        status=1
+    fi
+done
+exit $status
```
npmccallum:
```
top=$(git rev-parse --show-toplevel)
echo "The following files do not #![deny(missing_docs)]:"
! grep -LF '#![deny(missing_docs)]' $top/*/src/main.rs $top/*/src/lib.rs
```

npmccallum: The clippy one can be expressed similarly.

npmccallum:
@haraldh Any idea why these tests fail?

haraldh:
> @haraldh Any idea why these tests fail?

@npmccallum  ... no idea, except a broken ubuntu version of grep

mbestavros:
Using `git grep` seems to have fixed it. @haraldh was right; Ubuntu's standard `grep` is broken.

PR #322, opened 2020-03-12
Discussion Prompt

npmccallum:
This is an idea of how we might do syscall proxying. It isn't intended to be an implementation, but a discussion starter.


npmccallum:
I am closing this because we've chosen to pass the syscalls in registers directly for now.

PR #324, opened 2020-03-13
Validate that PRs include a closing keyword with Mergeable

mbestavros:
This should fail, since I have not yet created a corresponding issue. :stuck_out_tongue:

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -1,3 +1,9 @@
 version: 2
 mergeable:
   - when: pull_request.*
+    validate:
+      - do: description
+        must_include:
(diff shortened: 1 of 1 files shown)


mbestavros:
@npmccallum I'm not necessarily proposing we merge this yet, but it is a possible use case. I'm looking into more.

npmccallum:
@mbestavros If I remember correctly, the consensus was that we didn't want to force every PR to close an issue. If that is correct, please close this.

mbestavros:
@npmccallum Yes, that was our consensus. We don't want to enforce this, so I'm closing it.

PR #325, opened 2020-03-13
Validate that infrastructure PRs are correctly labeled

mbestavros:
This is WIP, but could be a cool use case if we can flesh it out.

When filing this PR, I only assigned it a project, and not a label. If this works, it should automatically get the label.

Edited to re-trigger Mergeable, again, and again.

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -1,3 +1,16 @@
 version: 2
 mergeable:
   - when: pull_request.*
+    validate:
+      - do: project
+        must_include:
+          regex: 'Infrastructure'
+          message: 'PR is in the Infrastructure project.'
+      - do: label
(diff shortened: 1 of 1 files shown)


mbestavros:
Current issue I'm having with this is the Projects validator not correctly detecting that this PR is, in fact, in the Infrastructure project. I've asked upstream for help.

mbestavros:
This seems to be an upstream bug with Mergeable. It's being tracked here: https://github.com/mergeability/mergeable/issues/257

mbestavros:
We've decided to implement our own bot. the first draft of that is open here: https://github.com/enarx/enarx/pull/360

Since that PR accomplishes what we were trying to do here and then some, we're closing this issue.

PR #326, opened 2020-03-16
Implement syscall handlers for exit() and getuid()

npmccallum:
This incrementally enables the output and input paths for syscalls. The
exit() syscall handler tests that we can make an outgoing request. It is
also, obviously, necessary for us to terminate the application. The
getuid() syscall handler tests that we can additionally return data to
the syscall call site. Implementation of other syscalls should be fairly
straightforward from here.


npmccallum:
I added six additional issues that cover either debt from this PR or next steps.

npmccallum:
@haraldh I just added a documentation commit. It isn't complete, but it is probably enough to get started.

npmccallum:
@lkatalin I added a stack layout table because I realized that this code was impossible to understand without it.

npmccallum:
FYI, docs are rendered here: https://github.com/npmccallum/enarx/blob/sgx/docs/keep/sgx/README.md

enarx-keep-sgx-shim/src/main.rs
```diff
@@ -16,10 +16,73 @@ fn main() {}
...
+            [0x0f, 0x05] => {
+                aex.gpr.rip += 2;
+                aex.gpr.rax = match aex.gpr.rax {
+                    // SYS_exit
+                    rax @ 60 => syscall(aex.gpr.rdi, 0, 0, aex, 0, 0, 0, rax, ctx),
+
+                    // SYS_getuid
+                    rax @ 102 => syscall(0, 0, 0, aex, 0, 0, 0, rax, ctx),
+
+                    _ => syscall(8, 0, 0, aex, 0, 0, 0, 60, ctx),
```
connorkuehl: Same `linux-syscall` comment as above.

enarx-keep-sgx-shim/src/main.rs
```diff
@@ -16,10 +16,73 @@ fn main() {}
...
+                };
+            }
+
+            _ => {
+                syscall(9, 0, 0, aex, 0, 0, 0, 60, ctx);
```
connorkuehl: Same `linux-syscall` comment as above.

enarx-keep-sgx-shim/src/main.rs
```diff
@@ -16,10 +16,73 @@ fn main() {}
...
+                rsi: u64,
+                rdx: u64,
+                r10: u64,
+                r8: u64,
+                r9: u64,
+                rax: u64,
+            ) -> u64;
+        }
+
+        let uid = do_syscall(0, 0, 0, 0, 0, 0, 102); // SYS_getuid
```
connorkuehl: Same `linux-syscall` comment as above.

enarx-keep-sgx-shim/src/main.rs
```diff
@@ -16,10 +16,73 @@ fn main() {}
...
+        do_syscall(uid / 100, 0, 0, 0, 0, 0, 60); // SYS_exit
```
connorkuehl: Same `linux-syscall` comment as above.

npmccallum:
This PR now depends on #342.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -56,26 +117,112 @@ shp:                            # The bottom of the secure heap
...
+    sub     %r11,                   %r10            # Subtract random offset
+
+    xor     %r11,                   %r11            # %r11 = NULL
+    xor     %rax,                   %rax            # %rax = NULL
+    jmp     .Lstack
+
+# CSSA != 0
+.Levent:
+    dec     %rax                                    # %rax = CSSA - 1
+    shl     $12,                    %rax            # %rax = (CSSA - 1) * 4096
```
jovanbulck: hm, not entirely sure, but I think you're silently assuming here that SSA frames are 1page (4KiB) in size(?) AFAIK this is _not_ guaranteed in the SGX architecture and you'd have to retrieve that from the SECS.SSAFRAMESIZE attribute:

> SSAFRAMESIZE 16 4 Size of one SSA frame in pages, including XSAVE, pad, GPR, and MISC (if
CPUID.(EAX=12H, ECX=0):.EBX != 0).

In practice, it might be that the way you build enclaves in Enarx they always have an SSA frame size of 1page (for now), but then again I think you should better explicitly document this assumption and encode it in a constant?

enarx-keep-sgx-shim/src/start.S
```diff
@@ -5,10 +5,67 @@
...
+.endm
+
+# Clear all temporary registers
+.macro  zerot
+    xor     %r10,                   %r10
+    xor     %r11,                   %r11
+.endm
+
+# Clear CPU flags using the supplied register (which MUST contain zero!)
+.macro  zerof reg
```
jovanbulck: Note that this macro currently does _not_ clear all EFLAGS bits. For instance one bit that remains under attacker control afais is the EFLAGS.AC bit which can be abused for side-channel purposes. As it should be fairly trivial to clear this bit, I strongly feel it is the resp of the trusted runtime to do so.

Also I'd advise to check out our Taleof2Worlds paper to try and sanitize the ABI systematically:

https://people.cs.kuleuven.be/~jo.vanbulck/ccs19-tale.pdf

enarx-keep-sgx-shim/src/start.S
```diff
@@ -56,26 +117,112 @@ shp:                            # The bottom of the secure heap
...
+    .type syscall, @function
+syscall:
+    savep                                           # Save preserved registers
+    mov     %rsp,                   SRSP(%rcx)      # Save restoration stack pointer
+
+    mov     0x38(%rsp),             %r10            # Export %r10
+    mov     0x40(%rsp),             %r11            # Export %rax in %r11
+    mov     0x48(%rsp),             %rsp            # Get the exit context
+
+    xor     %rcx,                   %rcx            # Clear %rcx
```
jovanbulck: It seems better to jump to some label `clear_regs_before_eexit` here to avoid having to duplicate this critical code?

enarx-keep-sgx-shim/src/start.S
```diff
@@ -56,26 +117,112 @@ shp:                            # The bottom of the secure heap
...
+
+    zerop                                           # Clear preserved registers
+    zerot                                           # Clear temporary registers
+    zerof   %r11                                    # Clear CPU flags
+
+    # void enter(rdi, rsi, rdx, tcs, r8, r9, &aex[CSSA-1], ctx);
+    call    enter                                   # Call enter()
+    add     $16,                    %rsp            # Remove parameters from stack
+
+    # Prepare CPU context for exit
```
jovanbulck: It seems this is not sufficient? You don't clear all general purpose registers like R15 etc? AFAIK no ABI will guarantee they don't contain secrets and you should generally clear _all_ CPU state before doing EEXIT..

Also, crucially, your entry/exit code does _not_ consider CPU extended state (including XMM/YMM/etc registers used by for instance AES-NI)? You should use `XSAVE` and friends for this -- see eg Intel SGX SDK entry/exit code.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -56,26 +117,112 @@ shp:                            # The bottom of the secure heap
...
-    pop     %rbx                                    # Restore next non-enclave addr
-    pop     %rsp                                    # Restore old stack
-    mov     $4,                     %rax            # EEXIT
-    enclu                                           # ENCLU[EEXIT]
+    xchg    %rbx,                   %rcx            # Swap TCS and next instruction.
+    cmp     $0,                     %rax            # If CSSA > 0...
+    jne     .Levent                                 # ... restore stack from AEX[CSSA-1].
+
+# CSSA == 0
+    rdrand  %r11                                    # Get a random number
```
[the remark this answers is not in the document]
npmccallum: I will absolutely fix the `rdrand` problem. Unfortunately, Rust doesn't support VLAs yet. However, I think I might have a workaround.

enarx-keep-sgx-shim/src/start.S
npmccallum: I'll address this.

Currently, the Enarx SGX libraries force the frame size to one 4K page. There is also, unfortunately, no access to the SECS from inside the enclave. Once we get EDMM support, all of this code will become dynamic. So the problem gets "solved" implicitly in the long run. But for now I can at least enshrine this in a constant.

enarx-keep-sgx-shim/src/start.S
npmccallum: We always attempt to clear the maximum registers for the exit type. Because we have different exit types, we have different registers cleared. I'll see if there is some way to consolidate this.

enarx-keep-sgx-shim/src/start.S
npmccallum: > It seems this is not sufficient? You don't clear all general purpose registers like R15 etc? AFAIK no ABI will guarantee they don't contain secrets and you should generally clear _all_ CPU state before doing EEXIT..

The preserved registers are loaded below on line 182. The only GPR register we don't clear is `%rax`, which is set to a constant for `EEXIT`.

> Also, crucially, your entry/exit code does _not_ consider CPU extended state (including XMM/YMM/etc registers used by for instance AES-NI)? You should use `XSAVE` and friends for this -- see eg Intel SGX SDK entry/exit code.

This was intentionally left out to reduce the review size on this already difficult-to-review PR. It is also something that can be done from Rust. We have an issue for it: https://github.com/enarx/enarx/issues/334

enarx-keep-sgx-shim/src/start.S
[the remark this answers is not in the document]
npmccallum: The workaround is to create a maximum sized array. Then take a random subslice, before turning it over to crt0stack. Then the top of the stack is randomized.

npmccallum:
Thanks everyone for the reviews! We still have more polish to do. But this is a good checkpoint. Merging.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -56,26 +117,112 @@ shp:                            # The bottom of the secure heap
...
+    .text
+    .globl syscall
+    .type syscall, @function
+syscall:
+    savep                                           # Save preserved registers
+    mov     %rsp,                   SRSP(%rcx)      # Save restoration stack pointer
+
+    mov     0x38(%rsp),             %r10            # Export %r10
+    mov     0x40(%rsp),             %r11            # Export %rax in %r11
+    mov     0x48(%rsp),             %rsp            # Get the exit context
```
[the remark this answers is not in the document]
lkatalin: An important thing to note is that, ex. `0x48 == 72 == 9 register offset == ctx`
where the stack looks like
`... | ctx | rax | r10 | rip | rbx | rbp | r12 | r13 | r14 | r15 |`

PR #341, opened 2020-03-17
Remove check for newline at EOF

connorkuehl:
We will defer to the Rust coding style enforced by `cargo fmt`. `cargo fmt` does not want newlines at EOF.


connorkuehl:
Er... I wonder if I've broken Github Actions by removing the `misc-eof` test...

npmccallum:
1. When you pushed your PR, both tests succeeded.
2. This test also verifies `.toml` files that `cargo fmt` doesn't check.

What's the motivation for removing it?

connorkuehl:
>     1. When you pushed your PR, both tests succeeded.

If I understand correctly, this test _should have failed_ when I pushed my PR because I removed the newlines at EOF that `cargo fmt` was complaining about, meaning that this test should have detected I did not have newlines at EOF.
 
>     2. This test also verifies `.toml` files that `cargo fmt` doesn't check.
> 
> What's the motivation for removing it?

Right, I didn't realize we were checking `.toml` files, too. The test says "ensure all code-related files have a newline at end of file" and I only thought of the Rust code (which shouldn't have newlines at EOF according to `cargo fmt`), so I thought it wouldn't make sense to have the test in that case.

I can update this PR to skip the `*.rs` files which won't have newlines at EOF due to `cargo fmt`.

PR #342, opened 2020-03-18
Update the constants crates

npmccallum:


npmccallum:
@connorkuehl Here's two more examples of `cargo fmt` behaving correctly...

linux-errno/src/lib.rs
```diff
@@ -4,8 +4,16 @@
 #![no_std]
 #![deny(clippy::all)]
 
+// ErrNo values generated with:
+//
+// ```
+// bindgen /usr/include/errno.h \
+//   | sed -rn 's|pub const (E[A-Z0-9_]*): u32 = ([0-9]+);|\1 = \2,|p' \
+//   | sort
```
connorkuehl: There is no objective basis for this, but would you consider sorting on the integer value instead of the symbol name:

```
| sort -g -t "=" -k 2
```

linux-syscall/src/lib.rs
```diff
@@ -4,8 +4,17 @@
 #![no_std]
 #![deny(clippy::all)]
 
+// SysCall values generated with:
+//
+// ```
+// bindgen /usr/include/sys/syscall.h \
+//   | sed -rn 's|pub const SYS_([a-z][a-z0-9_]*): u32 = ([0-9]+);|\1 = \2,|p' \
+//   | tr [:lower:] [:upper:] \
+//   | sort
```
connorkuehl: There is no objective basis for this, but would you consider sorting on the integer value instead of the symbol name:

```
| sort -g -t "=" -k 2
```

haraldh:
meh.. you didn't update the crate version with this breaking change!

npmccallum:
@haraldh We don't currently have *any* stable APIs. So we never bump any versions. But we do fix up all in-tree code.

haraldh:
> @haraldh We don't currently have _any_ stable APIs. So we never bump any versions. But we do fix up all in-tree code.

`0.1`, `0.2`, `0.3`, ... would be no stable version anyway.

But yeah.. if someone is referring to git repos, he should also refer to a specific branch or git-rev

connorkuehl:
> > @haraldh We don't currently have _any_ stable APIs. So we never bump any versions. But we do fix up all in-tree code.
> 
> `0.1`, `0.2`, `0.3`, ... would be no stable version anyway.
> 
> But yeah.. if someone is referring to git repos, he should also refer to a specific branch or git-rev

Yeah, that's true, but I think yours is a special case as you've been primarily working out of a separate repo. Once the microkernel and hypervisor are in-tree there will be no way to merge breaking changes across the codebases you work on.

PR #353, opened 2020-03-19
Use upstream libc mmap flags

### LINKED ISSUE #239: Rust's libc crate doesn't export MAP_FIXED_NOREPLACE on musl
We're building on musl libc now. But the Rust `libc` crate doesn't export `MAP_FIXED_NOREPLACE`. It is probably missing other flags too. We should patch upstream. I'm pretty sure we'll need a bunch of these flags at some point. (I need `MAP_FIXED_NOREPLACE` now...)

### LINKED ISSUE #339: Use upstream libc mmap flags for enarx-keep-sgx
These have been merged upstream, and when a new release comes out to crates.io they should be available for our consumption.

Once there's a new release, remove our definitions of these flags and use the ones exported by the libc crate.

Our definitions are here: https://github.com/enarx/enarx/blob/master/enarx-keep-sgx/src/map.rs#L10

connorkuehl:
Before this, libc did not export these flags for musl. Starting with libc 0.2.68 these flags are now exported.

Also, all I had to do to get these changes was run `cargo update` but I was expecting to see something change in a Cargo.lock... anyone have any thoughts? Otherwise people with this repo already checked out may need to run `cargo update` manually and if I had no background knowledge of this commit getting merged I would probably be confused.

Closes #239 
Closes #339

### DIFF
--- a/enarx-keep-sgx/Cargo.toml
+++ b/enarx-keep-sgx/Cargo.toml
@@ -16,7 +16,7 @@ vdso = { path = "../vdso" }
 bitflags = "1.2"
 openssl = "0.10"
 goblin = "0.2"
-libc = "0.2"
+libc = "0.2.68"
 
 [build-dependencies]
 cc = "1.0"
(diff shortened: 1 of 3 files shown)


connorkuehl:
> Do we need a hard dependency on `0.2.68`?

Yes, I will update the Cargo.toml to specify we need at least 0.2.68

PR #359, opened 2020-03-20
Conditionally remove SGX shim entry point if not on SGX-enabled hardware

### LINKED ISSUE #358: Skip tests in enarx-keep-sgx-shim if not running on SGX-enabled hardware
To reproduce, on non-Intel hardware:

```bash
bash-5.0# cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125
error: test failed, to rerun pass '--bin enarx-keep-sgx-shim'

Caused by:
  process didn't exit successfully: `/root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125` (signal: 11, SIGSEGV: invalid memory reference)
bash-5.0# gdb /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125

# [ .. ] snipped GDB startup messages

Reading symbols from /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125...
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) run
Starting program: /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125 
warning: Error disabling address space randomization: Operation not permitted

Program received signal SIGSEGV, Segmentation fault.
syscall () at src/start.S:209
209	    mov     %rsp,                   SRSP(%rcx)      # Save restoration stack pointer
(gdb) backtrace
#0  syscall () at src/start.S:209
#1  0x000000000046bec2 in getenv ()
#2  0x0000000000401779 in enclave () at src/start.S:201
#3  0x00007fff2d7db370 in ?? ()
#4  0x0000000000487800 in ?? ()
#5  0x0000000000000010 in ?? ()
#6  0x0000000000000000 in ?? ()
```

We need to somehow quickly exit the test binary without failing which may be difficult since the shim expects to be the lowest level entry point.

connorkuehl:
Otherwise the test suite will segfault and there's no real way to skip this since the shim presents itself as *the* entry point, even for `cargo test`

On non-Intel hardware:

Before the patch:

```bash
bash-5.0# cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125
error: test failed, to rerun pass '--bin enarx-keep-sgx-shim'

Caused by:
  process didn't exit successfully: `/root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125` (signal: 11, SIGSEGV: invalid memory reference)
```

After the patch:

```bash
bash-5.0# cargo test
   Compiling enarx-keep-sgx-shim v0.1.0 (/root/enarx/enarx-keep-sgx-shim)
    Finished test [unoptimized + debuginfo] target(s) in 0.57s
     Running /root/enarx/target/x86_64-unknown-linux-musl/debug/deps/enarx_keep_sgx_shim-328c79c1e5d7f125

running 3 tests
test libc::tests::bcmp ... ok
test libc::tests::memcpy ... ok
test libc::tests::memset ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Closes #358

### DIFF
--- a/enarx-keep-sgx-shim/link.json
+++ b/enarx-keep-sgx-shim/link.json
@@ -14,7 +14,7 @@
 
     "test": {
         "replace": {
-            "-lstart": []
+            "-lasm": []
         },
 
         "debug": false


connorkuehl:
I'm not in love with the fact that this solution requires programmers to know to put things _inside_ the compiler ifdef, but I think the alternative would be maybe we do something like if the flag is NOT defined we just attempt a `SYS_exit`? This would skip our tests though...

connorkuehl:
> I think it is saner to just avoid building and linking the asm altogether. That keeps the entire change inside the `build.rs`.

Agreed, however some of the routines in `src/main.rs` use the low level syscall routines defined by `start.S`. I suppose it would be cleaner to tease those out into their own translation unit which is always compiled and linked than the way I've just pushed which just turns off our `enter` function. Then, we can just not compile the sgx-specific stuff.

PR #365, opened 2020-03-24
cc: add a specific rlib from the target directory

haraldh:
adds "insert-target-rlib"

link.json:
…
        "insert-target-rlib": ["libc"],
…

will add

$HOME/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-musl/lib/liblibc-6c4492b949101b15.rlib

as a static library

### DIFF
--- a/cc
+++ b/cc
@@ -58,6 +58,7 @@ import json
 import sys
 import os
 import re
+import glob
 
 def replace(argv, regex, values):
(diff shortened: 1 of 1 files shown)


haraldh:
ah.. not needed now :)

PR #366, opened 2020-03-24
Add common code that is shared between SEV keep and shim

connorkuehl:
This will allow us to more easily parallelize our efforts to porting in Harald's work. There will be two projects consuming this common core:

- 'vmrun' AKA enarx-keep-sev
- 'kernel' AKA enarx-keep-sev-shim

I'd prefer we err on opening debt issues than making any large or invasive changes. This is the beginning of a rather delicate porting of a lot of SEV code.

### DIFF
--- a/vmsyscall/src/bootinfo.rs
+++ b/vmsyscall/src/bootinfo.rs
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `bootinfo` module includes the hard-coded memory layout and
+//! the `BootInfo` structure which is handed to the kernel.
+//!
+//! Initially copied from:
+//! https://github.com/rust-osdev/bootloader/blob/90f5b8910d146d6d489b70a6341d778253663cfa/src/bootinfo/mod.rs
+
+use crate::memory_map::Map;
+use core::fmt;
+
+/// This structure represents the information that the bootloader passes to the kernel.
+///
+/// The information is passed as an argument to the entry point:
+///
+/// ```ignore
+/// pub extern "C" fn _start(boot_info: &'static BootInfo) -> ! {
+///    // […]
+/// }
+/// ```
+///
+/// Note that no type checking occurs for the entry point function, so be careful to
+/// use the correct argument types. To ensure that the entry point function has the correct
+/// signature, use the [`entry_point`] macro.
+#[derive(Clone)]
+#[repr(C)]
+pub struct BootInfo {
+    /// A map of the physical memory regions of the underlying machine.
+    ///
+    /// The bootloader queries this information from the BIOS/UEFI firmware and translates this
+    /// information to Rust types. It also marks any memory regions that the bootloader uses in
+    /// the memory map before passing it to the kernel. Regions marked as usable can be freely
+    /// used by the kernel.
+    pub memory_map: Map,
+
+    /// Elf entry point of the ring3 executable
+    pub entry_point: *const u8,
+
+    /// Elf program headers of the ring3 executable
+    pub load_addr: *const u8,
+
+    /// Elf number of program headers of the ring3 executable
+    pub elf_phnum: usize,
+
+    /// Syscall trigger port
+    pub syscall_trigger_port: u16,
+}
+
+impl fmt::Debug for BootInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("BootInfo")
+            .field("memory_map", &self.memory_map)
+            .field(
+                "entry_point",
+                &format_args!("{:#x}", self.entry_point as u64),
+            )
+            .field("load_addr", &format_args!("{:#x}", self.load_addr as u64))
+            .field("elf_phnum", &self.elf_phnum)
+            .field("syscall_trigger_port", &self.syscall_trigger_port)
+            .finish()
+    }
+}
+
+#[test]
+fn check_bootinfo_size() {
+    use crate::memory_map::PAGE_SIZE;
+    assert!(core::mem::size_of::<BootInfo>() <= (PAGE_SIZE as _));
+}
--- a/vmsyscall/src/lib.rs
+++ b/vmsyscall/src/lib.rs
@@ -1,94 +1,92 @@
 // SPDX-License-Identifier: Apache-2.0
 
-//! syscall serialize/deserialize
+//! This crate is the interface between the SEV hypervisor (keep) and
+//! microkernel (shim). It enables system call proxying to the host
+//! and serialization of system calls.
 //!
 //! Currently it uses a hard coded page and an I/O trigger.
 //! We might want to switch to MMIO.
 
 #![deny(missing_docs)]
 #![deny(clippy::all)]
+#![deny(improper_ctypes)]
 #![no_std]
 
-use serde::{Deserialize, Serialize};
+pub mod bootinfo;
+pub mod memory_map;
 
-/// The syscalls to be serialized/deserialized via serde
-/// for the Hypervisor <-> VM syscall proxy
-#[derive(Serialize, Deserialize, Debug)]
+use core::fmt::{Debug, Formatter};
+
+impl Debug for VmSyscall {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
+            VmSyscall::Read { .. } => f.write_str("read(2)"),
+            VmSyscall::Write { .. } => f.write_str("write(2)"),
+            VmSyscall::Madvise { .. } => f.write_str("madvise(2)"),
+            VmSyscall::Mmap { .. } => f.write_str("mmap(2)"),
+            VmSyscall::Mremap { .. } => f.write_str("mremap(2)"),
+            VmSyscall::Munmap { .. } => f.write_str("munmap(2)"),
+            VmSyscall::Mprotect { .. } => f.write_str("mprotect(2)"),
+        }
+    }
+}
+
+/// maximum length of write(2) buffer
+pub const WRITE_BUF_LEN: usize = 4000;
+
+/// The syscalls for the Hypervisor <-> VM syscall proxy
+#[allow(clippy::large_enum_variant, missing_docs)]
 pub enum VmSyscall {
-    /// int madvise(void *addr, size_t length, int advice);
+    Read {
+        fd: u32,
+        count: usize,
+    },
+    Write {
+        fd: u32,
+        count: usize,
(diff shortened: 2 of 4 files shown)


vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,335 @@
...
+    Bootloader,
+    /// Frame at address zero.
+    ///
+    /// (shouldn't be used because it's easy to make mistakes related to null pointers)
+    FrameZero,
+    /// An empty region with size 0
+    Empty,
+    /// Additional variant to ensure that we can add more variants in the future without
+    /// breaking backwards compatibility.
+    NonExhaustive,
```
haraldh: why did you remove `#[doc(hidden)]` ??

connorkuehl: No good reason. Just saw the comment above it and thought it'd be useful information to appear in the generated documentation.

vmsyscall/src/lib.rs
```diff
@@ -1,94 +1,98 @@
...
+#[allow(missing_docs)]
+#[derive(Debug, PartialEq)]
 pub enum Error {
-    /// standard error
     Errno(i64),
-    /// serialize error
     SerializeError,
-    /// deserialize error
     DeSerializeError,
 }
```
npmccallum: I don't think Ser/DeSer errors are used anymore. Which also raises the question about whether this enum is needed.

vmsyscall/src/memory_map.rs
npmccallum: Is there a reason we can't use `#![non_exhaustive]` here?

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Returns the physical start address of the memory region.
+    pub fn start_addr(&self) -> u64 {
+        self.start_frame_number * PAGE_SIZE
+    }
+
+    /// Returns the physical end address of the memory region.
+    pub fn end_addr(&self) -> u64 {
+        self.end_frame_number * PAGE_SIZE
+    }
+}
```
npmccallum: I think we should use `core::ops::Range` here instead. This is also isomorphic with `span::Span`. Make a newtype `Frame(u64)` with conversions to/from addresses. Then just do `Range<Frame>` or `Span<Frame>`.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /// Mark a region as usable.
+    pub fn set_region_type_usable(&mut self, region_type: MemoryRegionType) {
+        self.iter_mut().for_each(|r| {
+            if r.region_type == region_type {
+                r.region_type = MemoryRegionType::Usable
+            }
+        });
+    }
```
npmccallum: This seems like this could be made more generic as `fn set_region_type(rtype: MemoryRegionType, whentype: Option<MemoryRegionType>)`.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+                && last_region.range.end_frame_number >= region.range.start_frame_number
+                && last_region.range.end_frame_number <= region.range.end_frame_number
+            {
+                last_region.range.end_frame_number = region.range.end_frame_number;
+                return Err(());
+            }
+            Ok(())
+        }) {
+            return;
+        };
```
npmccallum: I think this is more clearly expressed as a `for` loop where you test for exclusions and then `continue`. If you bypass all the exclusion tests then you modify and return.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /// Mark a region as allocated.
+    pub fn mark_allocated_region(&mut self, region: MemoryRegion) {
+        let mut region = region;
+        for r in self.iter_mut() {
+            // New region inside region of same type
+            if r.region_type == region.region_type
+                && r.range.start_frame_number <= region.range.start_frame_number
+                && r.range.end_frame_number >= region.range.end_frame_number
```
npmccallum: This pattern or similar occurs frequently. Could we make this an operation on the `MemoryRegion` struct?

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+                panic!(
+                    "region {:x?} overlaps with non-usable region {:x?}",
+                    region, r
+                );
+            }
+
+            match region
+                .range
+                .start_frame_number
+                .cmp(&r.range.start_frame_number)
```
npmccallum: I strongly dislike wraps in a match statement. Please use a temporary variable.

npmccallum: Also, this is screaming for an operation on the `Frame` newtype (see below).

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+            {
+                Ordering::Equal => {
+                    if region.range.end_frame_number < r.range.end_frame_number {
+                        // Case: (r = `r`, R = `region`)
+                        // ----rrrrrrrrrrr----
+                        // ----RRRR-----------
```
npmccallum: These comments are as good as ciphertext to me. I don't understand them.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub fn set_region_type_usable(&mut self, region_type: MemoryRegionType) {
+        self.iter_mut().for_each(|r| {
+            if r.region_type == region_type {
+                r.region_type = MemoryRegionType::Usable
+            }
+        });
+    }
+
+    /// Add a region to the `MemoryMap`.
+    pub fn add_region(&mut self, region: MemoryRegion) {
```
npmccallum: I think `add_region()` can be private.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,317 @@
...
+                    .start_frame_number
+                    .cmp(&r2.range.start_frame_number);
+
+                if ordering == Ordering::Equal {
+                    r1.range.end_frame_number.cmp(&r2.range.end_frame_number)
+                } else {
+                    ordering
+                }
+            }
+        });
```
npmccallum: This would be far more readable if you had a `Frame` newtype that implemented `PartialOrd` and then implemented `PartialOrd` on `Region`.

vmsyscall/src/memory_map.rs
haraldh: > Is there a reason we can't use `#![non_exhaustive]` here?

probably not

vmsyscall/src/lib.rs
haraldh: correct

vmsyscall/src/bootinfo.rs
```diff
@@ -0,0 +1,71 @@
...
+//! the `BootInfo` structure which is handed to the kernel.
+//!
+//! Initially copied from:
+//! https://github.com/rust-osdev/bootloader/blob/90f5b8910d146d6d489b70a6341d778253663cfa/src/bootinfo/mod.rs
+
+use crate::memory_map::MemoryMap;
+use core::fmt;
+
+/// Hard-coded trigger port
+pub const SYSCALL_TRIGGER_PORT: u16 = 0xFF;
```
haraldh: oh, we can move that into `vmrun` ... doesn't have to be here anymore, because it's a BootInfo argument

vmsyscall/src/bootinfo.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Elf program headers of the ring3 executable
+    pub load_addr: *const u8,
+
+    /// Elf number of program headers of the ring3 executable
+    pub elf_phnum: usize,
+
+    /// Syscall trigger port
+    pub syscall_trigger_port: u16,
+}
```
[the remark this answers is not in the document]
haraldh: Since it's part of the memory now (vmsyscall page) , wouldn't it be measured like the other parts?

vmsyscall/src/memory_map.rs
haraldh: Why? I would like to use it.

vmsyscall/src/memory_map.rs
haraldh: These are both regions, which start at the same point.. one is bigger than the other.

vmsyscall/src/memory_map.rs
npmccallum: Sure, no problem. I just didn't see any immediate reference to it outside the context.

vmsyscall/src/memory_map.rs
npmccallum: The `Option` might not make sense depending on how this is used.

vmsyscall/src/bootinfo.rs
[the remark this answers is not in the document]
npmccallum: Probably. I'm just noting that it is critical that the host cannot choose the starting point for the guest without it being measured. A comment here could be a helpful warning to future coders.

vmsyscall/src/memory_map.rs
haraldh: Right.. but the generic use might not make sense. At least I can't think of any currently.

connorkuehl:
Most recent push:

- [x] Remove `SYSCALL_TRIGGER_PORT` const as it will be in `enarx-keep-sev` instead.
- [x] Remove wrapping match statement
- [x] Simplify seek operation in `add_region`
- [x] Use `#[non_exhaustive]` on enum
- [x] Remove now-unused Error enum
- [x] Remove Memory prefix from types in `memory_map`
- [x] Implement Ord and PartialOrd on the memory `Region` type to simplify sorting... stepping stone toward Nathaniel's feedback based on the Frame newtype.
- [ ] Use Span/Range instead of FrameRange (started to spiral a bit in complexity)
- [ ] Frame newtype
- [ ] Make `set_region_type` more generic

I'm still working on the changes that aren't checked off above.

connorkuehl:
Closing in favor of #370

npmccallum:
@connorkuehl Unless you have a specific objection, I'd like to keep this open and merge this onto master and rebase `demo2020` on top of this. I think we also probably have time now to address the other issues in the review. Thoughts?

connorkuehl:
Discussed this briefly in gitter. I'm punting the refactor with a `Range` type until we have a tree with `vmsyscalls`, `enarx-keep-sev`, and `enarx-keep-sev-shim` all in one tree so that testing is easier and the replacement can be easily done all in one fell-swoop without affecting others' PRs.

PR #369, opened 2020-03-26
Add readv and writev syscalls for SGX

### LINKED ISSUE #375: Add more security checks in `ualloc()`
1. We should test that the end of the memory allocated by `ualloc()` does not refer to an address inside the enclave.

2. We should test that return value from `mmap()` inside `ualloc()` is either:
- an error number OR
- page aligned AND outside the enclave

The only variation from that is `MAP_FIXED` in which addr is not `NULL`, but still must be page aligned.

### LINKED ISSUE #376: Add readv, writev syscalls for SGX

lkatalin:
This is a draft. In particular, it needs some help with casting pointers in a safe and correct way.

Resolves #375 and resolves #376

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -2,10 +2,10 @@
 
 use nolibc::x86_64::error::Number as ErrNo;
 use nolibc::x86_64::syscall::Number as SysCall;
-use nolibc::ArchPrctlTask;
+use nolibc::{ArchPrctlTask, Iovec};
 use sgx_types::{ssa::StateSaveArea, tcs::Tcs};
 
-use core::ops::Range;
+use core::{mem::size_of, ops::Range, slice::from_raw_parts_mut};
 
 extern "C" {
     #[no_mangle]
@@ -225,5 +225,76 @@ impl<'a> Handler<'a> {
 
             _ => ErrNo::EINVAL.into_syscall(),
         }
+    }
+
+    /// Do a readv() syscall
+    pub fn readv(&mut self) -> u64 {
+        let fd = self.aex.gpr.rdi;
+        let trusted = unsafe {
+            from_raw_parts_mut(self.aex.gpr.rsi as *mut Iovec, self.aex.gpr.rdx as usize)
+        };
+
+        // Add up total size of buffers and size of iovec array.
+        let bufsize = trusted.iter().fold(0, |a, e| a + e.size);
+        let iovecsize = size_of::<Iovec>() * trusted.len();
+        let size = bufsize + iovecsize;
+
+        // Allocate some unencrypted memory.
+        let map = match self.ualloc(size as u64) {
+            Err(errno) => return errno.into_syscall(),
+            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
+        };
+
+        // Split allocated memory into that used by the iovec struct array and that used by its buffers.
+        let (uiovec, ubuffer) = map.split_at_mut(iovecsize);
+
+        // Convert the prefix from a byte slice into an iovec slice.
+        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<Iovec>() };
+        if untrusted.len() != trusted.len() {
+            self.attacked();
+        }
+
+        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory.
+        // The offset is into the buffer area allocated immediately after the iovec struct
+        // array, measured in bytes.
+        let mut offset = 0;
+        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+            u.base = ubuffer[offset..].as_mut_ptr();
+            u.size = t.size;
+            offset += t.size;
+        }
+
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let ret = unsafe {
+            self.syscall(
+                SysCall::READV,
+                fd,
+                untrusted.as_ptr() as _,
+                untrusted.len() as u64,
+                0,
+                0,
+                0,
+            )
+        };
+
+        // Copy the unencrypted input into encrypted memory.
+        if ErrNo::from_syscall(ret).is_none() {
+            if ret > size as u64 {
+                self.attacked();
+            }
+
+            let mut offset = 0;
+            for (t, u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+                if u.base != ubuffer[offset..].as_mut_ptr() || u.size != t.size {
+                    self.attacked();
+                }
+
+                t.as_mut().copy_from_slice(u.as_ref());
+                offset += t.size;
+            }
+        }
+
+        unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
+        ret
     }
 }
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -297,4 +297,67 @@ impl<'a> Handler<'a> {
         unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
         ret
     }
+
+    /// Do a writev() syscall
+    pub fn writev(&mut self) -> u64 {
+        let fd = self.aex.gpr.rdi;
+        let trusted = unsafe {
+            from_raw_parts_mut(self.aex.gpr.rsi as *mut Iovec, self.aex.gpr.rdx as usize)
+        };
+
+        // Add up total size of buffers and size of iovec array.
+        let bufsize = trusted.iter().fold(0, |a, e| a + e.size);
+        let iovecsize = size_of::<Iovec>() * trusted.len();
+        let size = bufsize + iovecsize;
+
+        // Allocate some unencrypted memory.
+        let map = match self.ualloc(size as u64) {
+            Err(errno) => return errno.into_syscall(),
+            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
+        };
+
+        // Split allocated memory into that used by the iovec struct array
+        // and that used by its buffers.
+        let (uiovec, ubuffer) = map.split_at_mut(iovecsize);
+
+        // Convert the prefix from a byte slice into an iovec slice.
+        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<Iovec>() };
+        if untrusted.len() != trusted.len() {
+            self.attacked();
+        }
+
+        // Set pointers in unencrypted iovec slice to use the rest
+        // of the allocated memory, then copy the encrypted input
+        // into unencrypted memory. The offset is into the buffer
+        // area allocated immediately after the iovec struct array,
+        // measured in bytes.
+        let mut offset = 0;
+        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+            u.base = ubuffer[offset..].as_mut_ptr();
+            u.size = t.size;
+            u.as_mut().copy_from_slice(t.as_ref());
+            offset += t.size;
+        }
+
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let ret = unsafe {
+            self.syscall(
+                SysCall::WRITEV,
+                fd,
+                untrusted.as_ptr() as _,
+                untrusted.len() as u64,
+                0,
+                0,
+                0,
+            )
+        };
+
+        unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
+
+        if ErrNo::from_syscall(ret).is_none() && ret > size as u64 {
+            self.attacked()
+        }
+
+        ret
+    }
 }
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -158,7 +158,7 @@ impl<'a> Handler<'a> {
                     self.attacked();
                 }
 
-                core::ptr::copy_nonoverlapping(buf, map, ret as _);
+                core::ptr::copy_nonoverlapping(map, buf, ret as _);
             }
 
             ret
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -94,7 +94,9 @@ impl<'a> Handler<'a> {
             )
         };
 
-        if self.enclave.contains(&ret) {
+        // Make sure the allocated memory is page-aligned and outside of the enclave.
+        if self.enclave.contains(&ret) || self.enclave.contains(&(ret + bytes)) || ret & 0xfff != 0
+        {
             self.attacked();
         }
 
--- a/nolibc/src/lib.rs
+++ b/nolibc/src/lib.rs
@@ -25,3 +25,27 @@ enumerate::enumerate! {
         ArchGetGs = 0x1004,
     }
 }
+
+/// Buffer used by readv() and writev() syscalls
+#[repr(C)]
+pub struct Iovec<'a> {
+    /// Buffer start address
+    pub base: *mut u8,
+
+    /// Number of bytes to transfer
+    pub size: usize,
+
+    phantom: core::marker::PhantomData<&'a ()>,
+}
+
+impl<'a> AsRef<[u8]> for Iovec<'a> {
+    fn as_ref(&self) -> &[u8] {
+        unsafe { core::slice::from_raw_parts(self.base, self.size) }
+    }
+}
+
+impl<'a> AsMut<[u8]> for Iovec<'a> {
+    fn as_mut(&mut self) -> &mut [u8] {
+        unsafe { core::slice::from_raw_parts_mut(self.base, self.size) }
+    }
+}
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -26,6 +26,7 @@ pub extern "C" fn event(
                 SysCall::READ => h.read(),
                 SysCall::READV => h.readv(),
                 SysCall::WRITE => h.write(),
+                SysCall::WRITEV => h.writev(),
                 SysCall::EXIT => h.exit(None),
                 SysCall::GETUID => h.getuid(),
                 SysCall::ARCH_PRCTL => h.arch_prctl(),
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -24,6 +24,7 @@ pub extern "C" fn event(
 
             aex.gpr.rax = match syscall {
                 SysCall::READ => h.read(),
+                SysCall::READV => h.readv(),
                 SysCall::WRITE => h.write(),
                 SysCall::EXIT => h.exit(None),
                 SysCall::GETUID => h.getuid(),


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+        let fd = self.aex.gpr.rdi;
+        let iovec = self.aex.gpr.rsi as *mut u8;
+        let iovcnt = self.aex.gpr.rdx as usize;
+
+        // Add up total size of buffers.
+        let iovec = unsafe { from_raw_parts_mut(iovec as *mut Iovec, iovcnt) }; // ???
+        let mut bufsize: usize = 0;
+        for i in 0..iovcnt {
+            bufsize += iovec[i].size;
+        }
```
npmccallum:
```suggestion
		let bufsize = trusted.iter().fold(0, |a, e| a + e.size);
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+
+        // Allocate some unencrypted memory.
+        let size = bufsize + size_of::<Iovec>() * iovcnt;
```
npmccallum:
```suggestion
        let size = bufsize + size_of::<Iovec>() * trusted.len();
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+        let map = match self.ualloc(size as u64) {
+            Err(errno) => return errno.into_syscall(),
+            Ok(map) => map,
+        };
+
+        // Cast beginning of allocated memory as iovec slice of length iovcnt.
+        let map = unsafe { from_raw_parts_mut(map as *mut Iovec, iovcnt) };
```
npmccallum:
```suggestion
        let untrusted = unsafe { from_raw_parts_mut(map as *mut Iovec, iovcnt) };
```

npmccallum: The general strategy with my naming here is to always communicate to future readers whether the types are trusted or not. We want to avoid accidents...

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+
+        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory.
+        let mut chunk = map.as_ptr() as usize + size_of::<Iovec>() * iovcnt; // ??? May be unsafe or illegal as per from_raw_parts_mut rules
```
npmccallum: I'm not sure I understand the question here.

npmccallum:
```suggestion
        let mut chunk = map.add(size_of::<Iovec>() * untrusted.len());
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+        for i in 0..iovcnt {
+            if chunk > map.as_ptr() as usize + size {
+                // ???
+                // TODO: What error to return?
+                self.attacked();
+            }
```
npmccallum: I realized that my check inside `ualloc()` is incorrect. We currently check that the start of the returned memory isn't inside the enclave, but we don't check the end. So fix that in a separate commit. Then you don't need this check at all.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+            Err(errno) => return errno.into_syscall(),
+            Ok(map) => map,
+        };
+
+        // Cast beginning of allocated memory as iovec slice of length iovcnt.
+        let map = unsafe { from_raw_parts_mut(map as *mut Iovec, iovcnt) };
+
+        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory.
+        let mut chunk = map.as_ptr() as usize + size_of::<Iovec>() * iovcnt; // ??? May be unsafe or illegal as per from_raw_parts_mut rules
+        for i in 0..iovcnt {
```
npmccallum:
```suggestion
        for (t, u) in trusted.iter().zip(untrusted.iter()) {
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+            if chunk > map.as_ptr() as usize + size {
+                // ???
+                // TODO: What error to return?
+                self.attacked();
+            }
+            map[i].base = chunk as *mut u8;
+            chunk = chunk + iovec[i].size;
```
npmccallum:
```suggestion
			u.base = chunk;
			u.size = t.size;
            chunk = chunk.add(t.size);
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+            let ret = self.syscall(
+                SysCall::READV,
+                fd,
+                map.as_ptr() as _,
+                iovcnt as u64,
+                0,
+                0,
+                0,
+            ); // ???
+            self.ufree(map.as_ptr() as *mut u8, size as u64); // ???
```
npmccallum: Don't free this until after you are done copying it...

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
...
+                    self.attacked();
+                }
+
+                for i in 0..iovcnt {
+                    core::ptr::copy_nonoverlapping(
+                        map[i].base as *mut u8,
+                        iovec[i].base as *mut u8,
+                        ret as _,
+                    );
+                }
```
npmccallum: Iterate over the slice rather than using an index. However, keep in mind that we handed the whole array of iovecs to the untrusted host. That means that we need to validate that the values haven't been manipulated. For example, the host might replace `iovec[0].base` with a pointer to enclave memory in order to trick us into copying encrypted memory somewhere else. So, like we did when initializing the array, we need to iterate over both slices. Validate that the untrusted iovec hasn't been modified (and trigger `attacked()` if it has; that is suspicious behavior). Then do the copy.

Alternatively, we could take advantage of the contiguousness of the `map` buffer to bypass the `untrusted` array altogether. This would probably be fewer instructions. However, we'd bypass an opportunity to detect malicious behavior. So I'd prefer to be slightly slower and more security conscious.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +188,71 @@ impl<'a> Handler<'a> {
             ret
         }
     }
+
+    /// Do a readv() syscall
+    pub fn readv(&mut self) -> u64 {
+        let fd = self.aex.gpr.rdi;
+        let iovec = self.aex.gpr.rsi as *mut u8;
+        let iovcnt = self.aex.gpr.rdx as usize;
```
npmccallum:
```suggestion
		let trusted = unsafe {
			from_raw_parts_mut(
				self.aex.gpr.rsi as *mut Iovec,
				self.aex.gpr.rdx as usize
			);
		};
```

npmccallum:
@lkatalin Also note that `clippy` is giving you really good feedback here.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: I was thinking of whether I'm allowed to do `map.as_ptr()` due to: "The memory referenced by the returned slice must not be accessed through any other pointer (not derived from the return value) for the duration of lifetime 'a. Both read and write accesses are forbidden." from https://doc.rust-lang.org/core/slice/fn.from_raw_parts_mut.html. But since the pointer is derived from the original 'map', maybe it's okay.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: You mean after `copy_nonoverlapping`, right? I was confused about this because in `read()` the `ufree` is called immediately after the syscall, which I didn't understand.

npmccallum: @lkatalin You probably don't understand it because it probably doesn't work. I was a bad boy and didn't test `read()` obviously... :)

enarx-keep-sgx-shim/src/handler.rs
npmccallum: @lkatalin What it means in this case is don't access the *memory* in this region. You can still use the base pointer to calculate a new pointer to memory *after* the region without harm.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Fair enough!

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Ah okay, thanks as usual for the insights.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: > For example, the host might replace iovec[0].base with a pointer to enclave memory ... Validate that the untrusted iovec hasn't been modified

@npmccallum For now, I'm assuming this validation means to check that none of the `base` pointers refer to something inside enclave memory. If there are other checks that should be done here, please let me know. I'll also continue to think on it.

I'm fine with doing the slower and more security-conscious version.

npmccallum: Testing that the base pointers (and base+size pointers) aren't inside the enclave is a weaker version of the testing I had in mind. I wanted to recreate the exact expected `base` and `size` that it should be and validate that the results haven't been tampered with in any way.

lkatalin:
@npmccallum Good point; I've tried to do it better here without casting through an integer.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,70 @@ impl<'a> Handler<'a> {
...
+        };
+
+        // Split allocated memory into that used by the iovec struct array and that used by its buffers.
+        let (uiovec, ubuffer) = map.split_at_mut(size_of::<Iovec>() * trusted.len());
+
+        // Convert the prefix from a byte slice into an iovec slice.
+        let (pre, untrusted, suf) = unsafe { uiovec.align_to_mut::<Iovec>() };
+        if pre != [] || suf != [] {
+            self.attacked();
+        }
```
npmccallum:
```suggestion
        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<Iovec>() };
        if untrusted.len() != trusted.len() {
            self.attacked();
        }
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,70 @@ impl<'a> Handler<'a> {
...
+
+        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory.
+        let mut chunk = ubuffer.as_ptr();
+        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+            u.base = chunk as *mut u8;
+            u.size = t.size;
+            chunk = unsafe { chunk.add(t.size) };
+        }
```
npmccallum: Instead of converting the `ubuffer` to a pointer and then doing pointer math, instead track the offset and convert a subslice to a pointer. That will remove the untrusted code block.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,70 @@ impl<'a> Handler<'a> {
...
+                }
+
+                let mut chunk = ubuffer.as_ptr();
+                for (t, u) in trusted.iter().zip(untrusted.iter()) {
+                    if u.base != chunk as *mut u8 || u.size != t.size {
+                        self.attacked();
+                    }
+                    core::ptr::copy_nonoverlapping(u.base as *mut u8, t.base as *mut u8, ret as _);
+                    chunk = chunk.add(t.size);
+                }
```
npmccallum: Same as above.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,70 @@ impl<'a> Handler<'a> {
...
+                self.ufree(map.as_ptr() as *mut u8, size as u64);
+            }
+            ret
```
npmccallum:
```suggestion
            }

            self.ufree(map.as_ptr() as *mut u8, size as u64);
            ret
```

enarx-keep-sgx-shim/src/handler.rs
lkatalin: This does make way more sense as a method of checking the same thing. Thank you.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: This also makes lots more sense.

lkatalin: However, I can't move this because `ret` is known only within the `unsafe` block.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,72 @@ impl<'a> Handler<'a> {
...
+            // Do the syscall; replace encrypted memory with unencrypted memory.
+            let ret = self.syscall(
+                SysCall::READV,
+                fd,
+                untrusted.as_ptr() as _,
+                trusted.len() as u64,
+                0,
+                0,
+                0,
+            );
```
npmccallum:
```suggestion
            let ret = unsafe {
                self.syscall(
                    SysCall::READV,
                    fd,
                    untrusted.as_ptr() as _,
                    trusted.len() as u64,
                    0,
                    0,
                    0,
                )
            };
```

npmccallum: That enables you to reduce the size of the unsafe block significantly.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,72 @@ impl<'a> Handler<'a> {
...
+
+            // Copy the unencrypted input into encrypted memory.
+            if ErrNo::from_syscall(ret).is_none() {
+                if ret > size as u64 {
+                    self.attacked();
+                }
+
+                let mut offset = 0;
+                for (t, u) in trusted.iter().zip(untrusted.iter()) {
+                    if u.base != ubuffer[offset] as *mut u8 || u.size != t.size {
```
npmccallum: This cast doesn't do what you think it does. It converts the byte at offset to a `*const u8`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +191,72 @@ impl<'a> Handler<'a> {
...
+                        self.attacked();
+                    }
+                    core::ptr::copy_nonoverlapping(u.base as *mut u8, t.base as *mut u8, ret as _);
```
npmccallum: Use the slice method `copy_from_slice()` instead.

lkatalin: Since `u.base` is a `*mut u8`, would I not have to turn it into a slice with the unsafe `core::slice::from_raw_parts_mut` in order to use that method? That would result in the same number `unsafe` lines here. Is there a better way?

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Right! Yes, I should have done this.

enarx-keep-sgx-shim/src/handler.rs
npmccallum: You know the offset into the `ubuffer`, which is already a slice.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +195,74 @@ impl<'a> Handler<'a> {
...
+        let fd = self.aex.gpr.rdi;
+        let trusted = unsafe {
+            from_raw_parts_mut(self.aex.gpr.rsi as *mut Iovec, self.aex.gpr.rdx as usize)
+        };
+
+        // Add up total size of buffers.
+        let bufsize = trusted.iter().fold(0, |a, e| a + e.size);
+
+        // Allocate some unencrypted memory.
+        let size = bufsize + size_of::<Iovec>() * trusted.len();
```
npmccallum: `prefix = sizeof * len`

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +195,74 @@ impl<'a> Handler<'a> {
...
+            offset += t.size;
+        }
+
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let ret = unsafe {
+            self.syscall(
+                SysCall::READV,
+                fd,
+                untrusted.as_ptr() as _,
+                trusted.len() as u64,
```
npmccallum:
```suggestion
                untrusted.len() as u64,
```

These values are the same. But I think this is slightly more readable.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +195,74 @@ impl<'a> Handler<'a> {
...
+                if u.base != ubuffer[offset..].as_mut_ptr() || u.size != t.size {
+                    self.attacked();
+                }
+                let tslice = unsafe { from_raw_parts_mut(t.base, t.size) };
+                let uslice = unsafe { from_raw_parts(u.base, u.size) };
+                tslice.copy_from_slice(uslice);
+                offset += t.size;
+            }
+        }
+        unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
```
npmccallum:
```suggestion

        unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +195,74 @@ impl<'a> Handler<'a> {
...
+        if ErrNo::from_syscall(ret).is_none() {
+            if ret > size as u64 {
+                self.attacked();
+            }
+
+            let mut offset = 0;
+            for (t, u) in trusted.iter().zip(untrusted.iter()) {
+                if u.base != ubuffer[offset..].as_mut_ptr() || u.size != t.size {
+                    self.attacked();
+                }
```
npmccallum:
```suggestion
                }

```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -23,6 +27,15 @@ extern "C" {
...
 pub enum Context {}
 
+/// Buffer used by readv() and writev() syscalls
+#[repr(C)]
+pub struct Iovec {
+    /// Buffer start address
+    base: *mut u8,
+    /// Number of bytes to transfer
+    size: usize,
+}
```
npmccallum: This struct probably belongs in `nolibc`.

lkatalin:
@npmccallum I put the `Iovec` into `nolibc`. I hope I used the lifetimes correctly. There is probably some room for improvement there. 

I also went ahead and included the `writev` call as well.

I am pretty sure since a Rust slice is a wrapper for a pointer that the way I copy them after converting to a slice from `Iovec` should work to update the memory pointed to without any further assignments, but it is worth double checking (ex. ln [308](https://github.com/enarx/enarx/pull/369/files#diff-a4a049afda935f763d1aa8821cc9c89eR308)).

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +183,150 @@ impl<'a> Handler<'a> {
...
+        // Copy the encrypted input into unencrypted memory.
+        let mut offset = 0;
+        for (t, u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+            if u.base != ubuffer[offset..].as_mut_ptr() || u.size != t.size {
+                self.attacked();
+            }
+
+            u.as_mut_slice().copy_from_slice(t.as_mut_slice());
+            offset += t.size;
+        }
```
npmccallum, line 310: We don't need two loops in this case. Also, we can drop the attacked check.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -180,4 +183,150 @@ impl<'a> Handler<'a> {
...
+                0,
+                0,
+            )
+        };
+
+        if ErrNo::from_syscall(ret).is_none() && ret > size as u64 {
+            self.attacked()
+        }
+
+        unsafe { self.ufree(map.as_ptr() as *mut u8, size as u64) };
```
npmccallum, line 329: Move this to immediately after the syscall.

nolibc/src/lib.rs
```diff
@@ -18,6 +18,42 @@ pub use x86_64::*;
...
+#[repr(C)]
+pub struct Iovec<'a, T: 'a> {
+    /// Buffer start address
+    pub base: *mut T,
+
+    /// Number of bytes to transfer
+    pub size: usize,
+
+    phantom: core::marker::PhantomData<&'a T>,
+}
```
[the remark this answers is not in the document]
lkatalin: I need to learn more about this, obviously. I thought it would tie the lifetime to the `base`. For now, Rust won't let me compile it without the `T` in the struct parameters: `pub struct Iovec<'a, T>` since it's required in `PhantomData`. I changed the `T` in the base to a `u8`.

nolibc/src/lib.rs
```diff
@@ -18,6 +18,30 @@ pub use x86_64::*;
...
+#[repr(C)]
+pub struct Iovec<'a, T> {
+    /// Buffer start address
+    pub base: *mut u8,
+
+    /// Number of bytes to transfer
+    pub size: usize,
+
+    phantom: core::marker::PhantomData<&'a T>,
+}
```
npmccallum, line 31:
```suggestion
#[repr(C)]
pub struct Iovec<'a> {
    /// Buffer start address
    pub base: *mut u8,

    /// Number of bytes to transfer
    pub size: usize,

    phantom: core::marker::PhantomData<&'a ()>,
}
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -182,4 +183,78 @@ impl<'a> Handler<'a> {
...
+                self.attacked();
+            }
+
+            let mut offset = 0;
+            for (t, u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+                if u.base != ubuffer[offset..].as_mut_ptr() || u.size != t.size {
+                    self.attacked();
+                }
+
+                t.as_mut().copy_from_slice(u.as_mut());
```
npmccallum, line 252:
```suggestion
                t.as_mut().copy_from_slice(u.as_ref());
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -257,4 +257,68 @@ impl<'a> Handler<'a> {
...
+        // Convert the prefix from a byte slice into an iovec slice.
+        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<Iovec<u8>>() };
+        if untrusted.len() != trusted.len() {
+            self.attacked();
+        }
+
+        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory,
+        // then copy the encrypted input into unencrypted memory.
+        // The offset is into the buffer area allocated immediately after the iovec struct
+        // array, measured in bytes.
```
npmccallum, line 294: Properly wrap the paragraph.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -257,4 +257,68 @@ impl<'a> Handler<'a> {
...
+        let mut offset = 0;
+        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
+            u.base = ubuffer[offset..].as_mut_ptr();
+            u.size = t.size;
+            u.as_mut().copy_from_slice(t.as_mut());
```
npmccallum, line 299:
```suggestion
            u.as_mut().copy_from_slice(t.as_ref());
```

nolibc/src/lib.rs
lkatalin: I didn't realize one could do this. Seems much better!

Issue #393, opened 2020-04-01
Implement support for multiple threads in exit_group


lkatalin:
The syscall handler for exit_group in the SGX shim should handle multiple threads and exit all of the threads in the calling process's thread group.

PR #395, opened 2020-04-01
Add preliminary exit_group() syscall handler

### LINKED ISSUE #394: Add SGX exit_group syscall handler
This is currently the same as the exit syscall handler, since we are only using one thread at the moment.

lkatalin:
Filed a debt issue to support multiple threads: #393 

Resolves #394


npmccallum:
@lkatalin :1st_place_medal:

PR #397, opened 2020-04-02
Stub for set_tid_address

### LINKED ISSUE #396: Add a stub for get_tid_address
Incomplete implementation that just returns a 1

whitebrandy:
Really not sure how to represent int *, is this the right way ?
(Potentially) resolves #396

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -191,4 +191,10 @@ impl<'a> Handler<'a> {
             ret
         }
     }
+
+    /// Do a set_tid_address() syscall
+    // This is currently unimplemented and returns a dummy thread id.
+    pub fn set_tid_address(&mut self) -> u64 {
+        1
+    }
 }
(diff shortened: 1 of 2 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +191,16 @@ impl<'a> Handler<'a> {
...
+    pub fn set_tid_address(&mut self) -> u64 {
+        // Unimplemented, this is used by the kernel to wake
+        // up a thread that is doing a futex wait on a memory
+        // location pointed to by _tidptr when this thread exits
+        // and returns the caller's thread id (1 for now)
+        let _tidptr = self.aex.gpr.rdi;
+        let tid: u64 = 1;
+
+        tid
+    }
```
npmccallum, line 205:
```suggestion
    /// Do a set_tid_address() syscall
    // This is currently unimplemented and returns a dummy thread id.
    pub fn set_tid_address(&mut self) -> u64 {
		1
    }
```

whitebrandy: Thank you, I updated the comment.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +191,13 @@ impl<'a> Handler<'a> {
...
         }
     }
+
+    /// Do a set_tid_address() syscall
+    // This is currently unimplemented and returns a dummy thread id.
+    pub fn set_tid_address(&mut self) -> u64 {
+        let _tidptr = self.aex.gpr.rdi;
+        let tid: u64 = 1;
+
+        tid
```
npmccallum, line 201:
```suggestion
        1
```

PR #400, opened 2020-04-02
Archprctl

### LINKED ISSUE #398: Implement arch_prctl SGX syscall handler
Based on http://man7.org/linux/man-pages/man2/arch_prctl.2.html. 
We only need "set" for now, not "get."

lkatalin:
Resolves #398


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -23,6 +23,28 @@ extern "C" {
...
+    pub enum ArchPrctlTask: u64 {
+        ArchSetFs = 0x1002,
+        ArchGetFs = 0x1003,
+        ArchSetGs = 0x1001,
+        ArchGetGs = 0x1004,
+    }
+}
+
+impl ArchPrctlTask {
+    pub fn from_u64(val: u64) -> Option<Self> {
```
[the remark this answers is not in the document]
lkatalin: You're right! That's great.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -23,6 +23,28 @@ extern "C" {
...
+        match val {
+            1 => Some(ArchPrctlTask::ArchSetGs),
```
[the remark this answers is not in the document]
lkatalin: I somehow totally misread the number.... but you are right, it doesn't matter now.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +213,36 @@ impl<'a> Handler<'a> {
             ret
         }
     }
+
+    /// Do an arch_prctl() syscall
+    pub fn arch_prctl(&mut self) -> u64 {
+        // TODO: Check that addr in %rdx does not point to an unmapped address
+        // and is not outside of the process address space.
+
+        // TODO: Should we return -1 on error? We can't do this with the current
```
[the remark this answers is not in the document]
lkatalin: So you're saying it's okay as it is? Our other handlers don't return -1 either, but I just wanted to check.

lkatalin:
Thanks for reviewing, @haraldh. I've updated the code with your suggestion.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -23,6 +23,16 @@ extern "C" {
...
 
+enumerate::enumerate! {
+    #[derive(Copy, Clone)]
+    pub enum ArchPrctlTask: u64 {
+        ArchSetGs = 0x1001,
+        ArchSetFs = 0x1002,
+        ArchGetFs = 0x1003,
+        ArchGetGs = 0x1004,
+    }
+}
```
npmccallum, line 34: This belongs in `nolibc`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +201,30 @@ impl<'a> Handler<'a> {
...
+        // and is not outside of the process address space.
+
+        match ArchPrctlTask::from(self.aex.gpr.rsi) {
+            ArchPrctlTask::ArchSetFs => {
+                self.aex.gpr.fsbase = self.aex.gpr.rdx;
+                0
+            }
+
+            // TODO: Fix me
+            ArchPrctlTask::ArchGetFs => ErrNo::EINVAL.into_syscall(),
```
npmccallum, line 217:
```suggestion
            ArchPrctlTask::ArchGetFs => ErrNo::ENOSYS.into_syscall(),
```

The value is valid, just not implemented.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +201,30 @@ impl<'a> Handler<'a> {
...
+
+            ArchPrctlTask::ArchSetGs => {
+                self.aex.gpr.gsbase = self.aex.gpr.rdx;
+                0
+            }
+
+            // TODO: Fix me
+            ArchPrctlTask::ArchGetGs => ErrNo::EINVAL.into_syscall(),
```
npmccallum, line 225:
```suggestion
            ArchPrctlTask::ArchGetGs => ErrNo::ENOSYS.into_syscall(),
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -6,6 +6,57 @@ use sgx_types::{ssa::StateSaveArea, tcs::Tcs};
...
+
+enumerate::enumerate! {
+    #[derive(Copy, Clone)]
+    pub enum ArchPrctlTask: u64 {
+        ArchSetGs = 0x1001,
+        ArchSetFs = 0x1002,
+        ArchGetFs = 0x1003,
+        ArchGetGs = 0x1004,
+    }
+}
```
npmccallum, line 19: This doesn't belong here.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -6,6 +6,57 @@ use sgx_types::{ssa::StateSaveArea, tcs::Tcs};
...
+#[cfg_attr(not(test), no_mangle)]
+pub unsafe extern "C" fn _bcmp(l: *const u8, r: *const u8, n: usize) -> usize {
+    let mut neq = 0;
+
+    for i in 0..n {
+        neq |= (*l.add(i) != *r.add(i)) as usize;
+    }
+
+    neq
+}
```
npmccallum, line 56: These should go into their own module. A top-level `libc` module is probably best. Also, remove the word temporary (don't forget the commit message). They might not be temporary.

lkatalin: Just FYI, what I tried to do initially is put them in their own module - a separate file in `enarx-keep-sgx-shim` and then `pub mod` for that in `main.rs`, but it gave me the same linking error until I put them all in this file.

npmccallum: @lkatalin Probably because you didn't remove them from `nolibc`.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: I just tried to put it back in `nolibc` and it resulted in the same linking error. I don't think relying on `nolibc` is a good idea at the moment.

lkatalin:
@npmccallum Unfortunately addressing either of your suggested changes here results in a linking error again.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Hmm, okay. I will try once more.

PR #402, opened 2020-04-02
enarx-keep-sev-shim update for demo2020

haraldh:


haraldh:
for demo2020

haraldh:
now has a nice `README.md` with instructions on how to debug with qemu and gdb :-)

PR #404, opened 2020-04-02
Add stubs for rt_sigaction and rt_sigprocmask in SGX syscall handler

### LINKED ISSUE #403: Add rt_sig* stubs for SGX syscall handler
Since the enclave never receives these signals, these can just return 0.

lkatalin:
Resolves #403 

Related debt issue is #405

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -191,4 +191,16 @@ impl<'a> Handler<'a> {
             ret
         }
     }
+
+    /// rt_sigaction() syscall stub
+    /// TODO: Implement correct behavior for this function.
+    pub fn rt_sigaction(&mut self) -> u64 {
+        0
+    }
+
+    /// rt_sigprocmask() syscall stub
+    /// TODO: Implement correct behavior for this function.
+    pub fn rt_sigprocmask(&mut self) -> u64 {
+        0
+    }
 }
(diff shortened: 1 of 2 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +191,18 @@ impl<'a> Handler<'a> {
             ret
         }
     }
+
+    /// rt_sigaction() syscall stub
+    /// This type of signal is not delivered to the enclave, so a dummy value
```
haraldh, line 196: Hmm, the comment is kind of wrong.

Even though, the outside world does not send a signal to the enclave process, the enclave could use `kill()` to send itsself a signal.

We also have all other kind of signal, see e.g. the `si_code` field documentation in the `rt_sigaction` man page.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -191,4 +191,18 @@ impl<'a> Handler<'a> {
...
+    /// is returned.
+    pub fn rt_sigaction(&mut self) -> u64 {
+        0
+    }
+
+    /// rt_sigprocmask() syscall stub
+    /// This type of signal is not delivered to the enclave, so a dummy value
+    /// is returned.
```
haraldh, line 204: same as for rt_sigaction.

npmccallum:
I agree with @haraldh's comments. Address those and I'll approve.

lkatalin:
@npmccallum @haraldh I updated the comments and filed a debt issue: #405 .

npmccallum:
Superseded by #408.

Issue #405, opened 2020-04-03
Implement correct behavior for rt_sig* functions in SGX syscall handler


lkatalin:
Currently there are stubs that return a dummy value of zero. 
However, for certain cases we will need to implement behavior where the enclave receives a signal, ex. if the enclave sends itself the `kill()` signal or for the many signals in the `si_code` field documentation in the `rt_sigaction` man page.

PR #408, opened 2020-04-06
Improve CPUID handling in SGX shim

npmccallum:
This previously contained a very large PR with many commits. We split up the PR into multiple PRs. This contains only the CPUID patches now.

### DIFF
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -46,6 +46,27 @@ pub extern "C" fn event(
             aex.gpr.rip += 2;
         }
 
+        // cpuid
+        [0x0f, 0xa2] => {
+            let (rax, rbx, rcx, rdx) = match (h.aex.gpr.rax, h.aex.gpr.rcx) {
+                (a, c) => {
+                    h.print("unsupported cpuid: (");
+                    h.print(&a);
+                    h.print(", ");
+                    h.print(&c);
+                    h.print(")\n");
+
+                    (0, 0, 0, 0)
+                }
+            };
+
+            aex.gpr.rax = rax;
+            aex.gpr.rbx = rbx;
+            aex.gpr.rcx = rcx;
+            aex.gpr.rdx = rdx;
+            aex.gpr.rip += 2;
+        }
+
         r => {
             let opcode = (r[0] as u16) << 8 | r[1] as u16;
             h.print("unsupported opcode: ");
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -49,6 +49,8 @@ pub extern "C" fn event(
         // cpuid
         [0x0f, 0xa2] => {
             let (rax, rbx, rcx, rdx) = match (h.aex.gpr.rax, h.aex.gpr.rcx) {
+                (0, _) => (0, 0x756e_6547, 0x6c65_746e, 0x4965_6e69), // "GenuineIntel"
+
                 (a, c) => {
                     h.print("unsupported cpuid: (");
                     h.print(&a);


npmccallum:
@mbestavros The labeller bot is broken. It keeps adding and removing (!) labels.

npmccallum:
@lkatalin I know you've already started a review. I'll wait for your review to push.

connorkuehl:
@npmccallum I'm also in the middle of reviewing

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -33,6 +37,74 @@ pub struct Handler<'a> {
     ctx: &'a Context,
 }
 
+impl<'a> Print<str> for Handler<'a> {
+    fn print(&mut self, data: &str) {
+        const STDERR: u64 = 2;
```
connorkuehl, line 44: Should this fd number be put in `nolibc`?

enarx-keep-sgx-shim/src/heap.rs
```diff
@@ -0,0 +1,198 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use core::mem::size_of;
+use core::slice::from_raw_parts_mut as slice;
+
+use nolibc::x86_64::error::Number as ErrNo;
+use page::Page;
+use span::{Line, Span};
+
+type Word = usize;
```
connorkuehl, line 10: I wonder if just plain `u64` is more clear here.

enarx-keep-sgx-shim/src/heap.rs
```diff
@@ -0,0 +1,198 @@
...
+        let prot = prot as u64 & !RWX;
+        if addr != 0 || fd != -1 || offset != 0 || prot != 0 || flags as u64 != PA {
+            return ErrNo::EINVAL.into_syscall() as _;
+        }
+
+        // The number of pages we need for the given length.
+        let pages = (length + Page::size() - 1) / Page::size();
+
+        // Find the brk page offset.
+        let brk = self.offset_page_up(self.metadata.brk.end).unwrap();
```
connorkuehl, line 153: If we trip the unwrap, is it a failure case we can handle and recover from in here?

enarx-keep-sgx-shim/src/heap.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        ErrNo::ENOMEM.into_syscall() as _
+    }
+
+    pub fn munmap(&mut self, addr: usize, length: usize) -> usize {
+        if addr % Page::size() != 0 {
+            return ErrNo::EINVAL.into_syscall() as _;
+        }
+
+        let brk = self.offset_page_up(self.metadata.brk.end).unwrap();
```
connorkuehl, line 176: If we trigger the unwrap, is it a failure can we can recover from in here?

enarx-keep-sgx/src/builder.rs
```diff
@@ -137,8 +137,8 @@ impl Builder {
                 }
             }
 
-            let line = Line::from(span);
-            eprintln!("{:016x}-{:016x} {:?}", line.start, line.end, si);
+            //let line = span::Line::from(span);
+            //eprintln!("{:016x}-{:016x} {:?}", line.start, line.end, si);
```
connorkuehl, line 141: Should we just delete these lines?

enarx-keep-sgx-shim/src/elf.rs
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub const MAGIC: [u8; 4] = [0x7f, 0x45, 0x4c, 0x46];
+//pub const CLASS_64: u8 = 2;
+//pub const DATA_LE: u8 = 1;
+//pub const VERSION: u8 = 1;
+//pub const OSABI_LINUX: u8 = 3;
+//pub const TYPE_DYN: u16 = 3;
+//pub const MACHINE_AMD64: u16 = 0x3e;
```
connorkuehl, line 9: Should we remove these?

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -22,37 +27,90 @@ extern "C" {
     ) -> u64;
 }
 
+pub trait Print<T: ?Sized> {
+    fn print(&mut self, data: &T);
+}
+
```
connorkuehl, line 33: Could Handler just implement `core::fmt::Write`? Then we get the print macros.

enarx-keep-sgx-shim/src/elf.rs
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub const MAGIC: [u8; 4] = [0x7f, 0x45, 0x4c, 0x46];
```
lkatalin, line 3: Curious about the `MAGIC`.

enarx-keep-sgx-shim/src/main.rs
```diff
@@ -23,3 +23,20 @@ mod entry;
 mod event;
 mod handler;
 mod libc;
+
+use span::Line;
+
+// NOTE: this must be kept in sync with enarx-keep-sgx
```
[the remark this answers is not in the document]
npmccallum: I'd much rather branch it into a separate library consumed by both. I did it this way for now because I suspect this needs `cargo make`. Also, `enarx-keep-sgx` uses `usize` and `enarx-keep-sgx-shim` uses `u64`. That's also a bigger strategy problem that boils down to the `StateSaveArea` struct definition. I think the right solution is probably adding some debt bugs.

enarx-keep-sgx-shim/src/handler.rs
npmccallum: Probably. I want a debt bug to remove `nolibc` altogether. But I think this will require `cargo make`.

enarx-keep-sgx-shim/src/heap.rs
npmccallum: I'd prefer to move all our `u64`s to `usize`s. But `StateSaveArea` is the biggest problem. We probably need to generate getters/setters for `usize`.

enarx-keep-sgx-shim/src/heap.rs
npmccallum: We need a better solution here, I agree. A failed `unwrap()` here causes a deadlock (infinite loop in the panic handler). I know, because I hit it.

enarx-keep-sgx-shim/src/heap.rs
npmccallum: Same as above.

enarx-keep-sgx/src/builder.rs
npmccallum: No. These are incredibly useful for debugging. But I agree we probably need a better strategy. This is something that we need to sync sgx/sev on. Do we want to enable debug output with debug builds and disable for release builds? @connorkuehl @haraldh ?

enarx-keep-sgx-shim/src/elf.rs
npmccallum: My preference would be to remove all of it and depend on a real elf library.

Which brings up one question we need to answer: goblin or xmas-elf? @connorkuehl @haraldh

enarx-keep-sgx-shim/src/handler.rs
npmccallum: https://www.youtube.com/watch?v=YJDNbhrGFy0

span/src/lib.rs
```diff
@@ -1,52 +1,38 @@
 // SPDX-License-Identifier: Apache-2.0
 
+//! This crate contains types for measuring linear sets by either the end
+//! points (`Bounds`) or by a starting point and the number of elements (`Span`).
+//!
+//! In the interest of zero-cost abstractions, all method are always inlined
```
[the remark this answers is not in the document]
npmccallum: Agreed.

enarx-keep-sgx/src/component.rs
```diff
@@ -1,28 +1,32 @@
...
-use std::fs::File;
-use std::io::{Read, Result};
+use std::io::Result;
 use std::ops::Range;
 use std::path::Path;
 
 /// A loadable segment of code
 pub struct Segment {
-    pub src: Contents,
+    src: Vec<Page>,
```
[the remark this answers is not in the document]
npmccallum: This is probably a mistake on my part.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -265,17 +265,17 @@ impl<'a> Handler<'a> {
         // TODO: Check that addr in %rdx does not point to an unmapped address
         // and is not outside of the process address space.
 
-        match ArchPrctlTask::from(self.aex.gpr.rsi) {
+        match ArchPrctlTask::from(self.aex.gpr.rdi) {
```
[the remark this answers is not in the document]
npmccallum: Yup. I was confused by that too. But when I inspected the raw syscall values, it was clear that this was the case. There's probably some context I'm missing.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -171,10 +173,39 @@ impl<'a> Handler<'a> {
         self.syscall(SysCall::MUNMAP, map as _, bytes, 0, 0, 0, 0)
     }
 
+    fn trace(&mut self, name: &str, argc: usize) {
```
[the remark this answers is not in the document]
npmccallum: Better to be defensive. Let's fix it.

enarx-keep-sgx-shim/src/main.rs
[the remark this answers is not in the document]
npmccallum: Agreed. See my comment to @connorkuehl.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -1,16 +1,13 @@
...
-#define STCK    (8 * 1024 * 1024)
-#define HEAP    (128 * 1024 * 1024)
-
 #define GPR     (4096 - 184)
 #define RSP     (GPR + 32)
 
 #define MISC    (GPR - 16)
 #define SRSP    (MISC - 8)
 
+#define STACK   (9 * 8)
```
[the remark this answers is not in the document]
npmccallum: That is the offset into the top of the stack in the `Layout` struct.

enarx-keep-sgx/src/component.rs
```diff
@@ -44,22 +45,22 @@ impl<'a> Segment {
...
             start: ph.p_vaddr as usize,
             count: ph.p_memsz as usize,
         });
 
-        let frame = Range {
-            start: unaligned.start / Page::size(),
-            end: (unaligned.end + Page::size() - 1) / Page::size(),
+        let frame = Line {
+            start: unaligned.start / size_of::<Page>(),
+            end: (unaligned.end + size_of::<Page>() - 1) / size_of::<Page>(),
```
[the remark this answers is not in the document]
npmccallum: A mistake, most likely.

enarx-keep-sgx/src/layout.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+fn below(rel: impl Into<Line<usize>>, size: usize) -> Span<usize> {
+    Span {
+        start: lower(rel.into().start - size, ALIGN),
+        count: size,
+    }
+}
+
+// NOTE: this structure MUST be kept in sync with enarx-keep-sgx-shim
```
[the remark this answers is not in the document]
npmccallum: Agreed. See my comment in @connorkuehl's review.

enarx-keep-sgx-shim/src/elf.rs
npmccallum: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header

Also: https://en.wikipedia.org/wiki/Magic_number_(programming)

The `MAGIC` number lets a program like `file` scan the first four bytes of a file and detect that it is an ELF binary. In our case we're using them as a sanity test to make sure we're looking at the right memory.

enarx-keep-sgx/src/builder.rs
haraldh: My experience with customers using my software taught me, that it is incredibly useful to have some debugging switches, which show lots of debugging useful for bug reports.

enarx-keep-sgx-shim/src/elf.rs
haraldh: Open an issue for that we investigate which is better.

enarx-keep-sgx-shim/src/handler.rs
connorkuehl: Yeah, I understand the trepidation about adding to `nolibc` when we want to rip it out. On the bright side, if we add this to `nolibc`, the compiler will remind us to update the call sites when the time comes to remove `nolibc`.

enarx-keep-sgx/src/builder.rs
connorkuehl: Yes, I'd love debug build output.

haraldh: Normally I do:
```rust
#[cfg(debug_assertions)]
    eprintln!(...)
```
which prints for non `--release` builds... as long as `-C opt-level=3` is not specified in `.cargo/config` :cry:

connorkuehl: > which prints for non --release builds... as long as -C opt-level=3 is not specified in .cargo/config :cry:

That's probably ok, I can't imagine a debug build with opt-level 3 :nauseated_face: 

Though I'd prefer something like

```rust
#[cfg(debug)]
eprintln!("blorp");
```

But that syntactic sugar would require us to put some `cfg_attr` in a common area for each crate.

```rust
#[cfg_attr(debug_assertions, debug)]
```

enarx-keep-sgx-shim/src/main.rs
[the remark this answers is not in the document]
npmccallum: I added a debt bug: https://github.com/enarx/enarx/issues/424

enarx-keep-sgx-shim/src/handler.rs
npmccallum: Fixed.

enarx-keep-sgx-shim/src/heap.rs
npmccallum: This should be addressed by debt bug #424.

enarx-keep-sgx/src/builder.rs
npmccallum: @connorkuehl @haraldh We have an added problem though. We can't allow the host to turn on debugging switches since this leaks an incredible amount of information to the host. This is why I was thinking of gating this on the build release profile. For example, if you're running on Enarx built with `cargo build --release` you get no debugging messages and we don't enable the debug flag for the underlying technology. However, on the non-release builds, the debug flag for the underlying technology is set to on and the keep will dump debug output.

enarx-keep-sgx-shim/src/elf.rs
npmccallum: https://github.com/enarx/enarx/issues/425

I will remove the commented stuff for now.

enarx-keep-sgx/src/builder.rs
connorkuehl: @npmccallum @haraldh Would we have the same problem if a host built their own debug packages of Enarx?

npmccallum: The keep measurement would differ from the official release. So simply by attesting you know it isn't in debug mode.

enarx-keep-sgx-shim/src/handler.rs
npmccallum: I attempted to implement this yesterday with some success, but encountered two problems.

1. While we get the use of macros like `writeln!()`, this offloads error handling to every call site (the compiler warns us that `Result` is unused). However, we don't really have a good way to handle this error should it arise. We could potentially work around this with a local macro (maybe named `eprintln!()`?) that calls `attacked()` if an error occurs.

2. Some forms of format arguments don't work and it is unclear why. Frustratingly, debug formatting doesn't work (`{:?}`) and causes `writeln!()` to fail after outputting all previous data.

Unless we can solve (2) quickly, I suggest we make this a debt issue.

npmccallum: @connorkuehl I made a `debug!()` wrapper for handling. So (1) is solvable. However, I also discovered that attempting to format strings causes a page fault, for a reason I don't understand.

Are you okay with making this a `debt` issue?

connorkuehl: @npmccallum Yeah, let's make this a `debt` issue.

The debug formatter is a head scratcher. Theoretically format_args! is a compile-time primitive that is non-allocating, and I always just thought that the output of a debug format specifier is a `str` anyway, so I can't imagine why it's fussing.

lkatalin:
@npmccallum Btw, the commit you were looking for that closes #405 is [`72ada85e7f0ec776367686e52811788224fe99f8`](https://github.com/enarx/enarx/pull/408/commits/72ada85e7f0ec776367686e52811788224fe99f8) (`Fix brk()...`).

npmccallum:
In case it isn't clear, I've been breaking up this PR into smaller PRs for individual patches. This should make it easier to review (and rebase!).

enarx-keep-sgx-shim/src/handler.rs
npmccallum: https://github.com/enarx/enarx/issues/469

haraldh: Here is my println!() ...https://github.com/enarx/enarx/blob/demo2020/enarx-keep-sev-shim/src/arch/x86_64/serial/syscall_serial.rs

haraldh: needs cleanup.. historically grown...

npmccallum:
I'm closing this in favor of the smaller PRs.

PR #410, opened 2020-04-06
Added text.

MikeCamel:
My first Rust!

### DIFF
--- a/payload/src/main.rs
+++ b/payload/src/main.rs
@@ -28,4 +28,7 @@ fn main() {
     for aux in &mut reader {
         println!("aux: {:?}", aux);
     }
+
+    println!();
+    println!("Good morning, that's a nice Tnetennba!");
 }


payload/src/main.rs
```diff
@@ -28,4 +28,7 @@ fn main() {
     for aux in &mut reader {
         println!("aux: {:?}", aux);
     }
+
+    println!();
+    println!("that's a nice TNETENNBA!");
```
npmccallum, line 33:
```suggestion
    println!("Good morning, that's a nice tnetennba!");
```

MikeCamel: The capitalisation is disputed: I'll amend slightly.  The salutation should, indeed, be included.

payload/src/main.rs
```diff
@@ -28,4 +28,7 @@ fn main() {
     for aux in &mut reader {
         println!("aux: {:?}", aux);
     }
+
+    println!();
+    println!("Good morning, that's a nice Tnetennba!");
```
lkatalin, line 33:
```suggestion
    println!("\nGood morning, that's a nice Tnetennba!");
```

MikeCamel: Is that considered good practice these days?  Fewer lines, but less legible.  Rejected!

npmccallum: `println!()` is preferred.

Technically, `\n` as newline character isn't portable. Though in practice it works pretty much everywhere except Windows.

lkatalin: I had no idea! Learned something new.

MikeCamel:
1. I've squashed the commits.  Do I need to repull?
2. Good idea.

MikeCamel:
>     1. Please squash the commits. We only want a single commit for this.
> 
>     2. Also, I'm wondering if this might be a better fit for a C-language example. We can demo the Rust payload. But then also show that C or C++ works just as well.

2. good idea.

npmccallum:
@MikeCamel Let's pause on this PR until @haraldh gets `musl-gcc -static-pie` working. Then we'll see if we can't come up with a C-lang demo.

npmccallum:
The demo is done and this PR isn't needed anymore. But we did use your text. :)

Issue #371, opened 2020-03-26
Refactor changes to 'vmsyscall' and merge into master


connorkuehl:
Chiefly, instead of `FrameRange` we could have a `Frame` newtype and `FrameRange` could be replaced with `core::ops::Range` or `enarx/span::Span`. Implementing `PartialOrd` and `Ord` on these new types will simplify some of the ordering done to the memory `Region`s.

The fact that `core::ops::Range` is not `Copy` may complicate things. This change will also require changes to `enarx-keep-sev` and possibly `enarx-keep-sev-shim`.

See #370, this could should be on the staging branch once that is merged.

connorkuehl:
Actually, with the new shape everything is taking, I think it makes more sense to track the progress of the keep and shim in their respective meta issues.

Issue #313, opened 2020-03-11
Implement enarx-keep-sev


connorkuehl:
**Description**

Port [Harald's hypervisor implementation](https://github.com/haraldh/enarx_sev_kvm_demo/tree/master/vmrun) into the Enarx repo. This is called `vmrun` in his repo. It will likely take a slightly different shape than it is in Harald's repo for the sake of integrating it in with the Enarx repo and/or due to some refactoring that Harald mentioned it could benefit from.

**To do**

- [x] `enarx-keep-sev` is a binary crate that we can pass command line arguments to supply file paths to the kernel binary and the "code" binary (the application we will run)
- [x] Create a KVM VM context
- [x] Load the microkernel binary into the virtual machine address space
- [x] Expose a port to the microkernel and add an event loop to service requests from the microkernel (for proxying syscalls/shutting down)
- [x] Inject the program binary that we are running (`code`)
  - [x] Load and copy this binary into the VM's address space
  - [x] Provide a boot/command line argument to the microkernel so that it knows where to find the user binary to run

npmccallum:
@connorkuehl If this is done, please close it.

PR #416, opened 2020-04-08
Move vmsyscall changes from demo2020 into master

connorkuehl:
note: leaving this as a draft because I cherry-picked Nathaniel's commit from his other PR that contains the `Line` type to base this on for some changes to `memory_map.rs`; so once that type is concretely in the repo I can rebase.

edit: woops guess it's not a draft

This is the first of probably several PRs that will gradually move code off of the `demo2020` branch.

This contains all of the feedback in #366 sans a more generic `set_region_type` method.

Related: #371 
Related: #313

### DIFF
--- a/memory/src/page.rs
+++ b/memory/src/page.rs
@@ -3,7 +3,7 @@
 /// A single page of memory
 ///
 /// This type is page-aligned and page-sized.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
 #[repr(C, align(4096))]
 pub struct Page([[u64; 32]; 16]);
 
--- a/vmsyscall/src/bootinfo.rs
+++ b/vmsyscall/src/bootinfo.rs
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `bootinfo` module includes the hard-coded memory layout and
+//! the `BootInfo` structure which is handed to the kernel.
+//!
+//! Initially copied from:
+//! https://github.com/rust-osdev/bootloader/blob/90f5b8910d146d6d489b70a6341d778253663cfa/src/bootinfo/mod.rs
+
+use crate::memory_map::Map;
+use core::fmt;
+
+/// This structure represents the information that the bootloader passes to the kernel.
+///
+/// The information is passed as an argument to the entry point:
+///
+/// ```ignore
+/// pub extern "C" fn _start(boot_info: &'static BootInfo) -> ! {
+///    // […]
+/// }
+/// ```
+///
+/// Note that no type checking occurs for the entry point function, so be careful to
+/// use the correct argument types. To ensure that the entry point function has the correct
+/// signature, use the [`entry_point`] macro.
+#[derive(Clone)]
+#[repr(C)]
+pub struct BootInfo {
+    /// A map of the physical memory regions of the underlying machine.
+    ///
+    /// The bootloader queries this information from the BIOS/UEFI firmware and translates this
+    /// information to Rust types. It also marks any memory regions that the bootloader uses in
+    /// the memory map before passing it to the kernel. Regions marked as usable can be freely
+    /// used by the kernel.
+    pub memory_map: Map,
+
+    /// ELF entry point of the ring 3 executable
+    pub entry_point: *const u8,
+
+    /// ELF program headers of the ring 3 executable
+    pub load_addr: *const u8,
+
+    /// ELF number of program headers of the ring 3 executable
+    pub elf_phnum: usize,
+
+    /// Syscall trigger port
+    pub syscall_trigger_port: u16,
+}
+
+impl fmt::Debug for BootInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("BootInfo")
+            .field("memory_map", &self.memory_map)
+            .field(
+                "entry_point",
+                &format_args!("{:#x}", self.entry_point as u64),
+            )
+            .field("load_addr", &format_args!("{:#x}", self.load_addr as u64))
+            .field("elf_phnum", &self.elf_phnum)
+            .field("syscall_trigger_port", &self.syscall_trigger_port)
+            .finish()
+    }
+}
+
+#[test]
+fn check_bootinfo_size() {
+    assert!(core::mem::size_of::<BootInfo>() <= (memory::Page::size() as _));
+}
--- a/vmsyscall/src/lib.rs
+++ b/vmsyscall/src/lib.rs
@@ -10,6 +10,8 @@
 #![deny(clippy::all)]
 #![no_std]
 
+pub mod memory_map;
+
 /// maximum length of write buffer
 pub const WRITE_BUF_LEN: usize = 4000;
 
--- a/vmsyscall/src/lib.rs
+++ b/vmsyscall/src/lib.rs
@@ -10,9 +10,21 @@
 #![deny(clippy::all)]
 #![no_std]
 
+/// maximum length of write buffer
+pub const WRITE_BUF_LEN: usize = 4000;
+
 /// System call requests originating from the microkernel.
 #[allow(clippy::large_enum_variant, missing_docs)]
 pub enum VmSyscall {
+    Read {
+        fd: u32,
+        count: usize,
+    },
+    Write {
+        fd: u32,
+        count: usize,
+        data: [u8; WRITE_BUF_LEN],
+    },
     Madvise {
         addr: usize,
         length: usize,
@@ -44,6 +56,10 @@ pub enum VmSyscall {
 impl core::fmt::Debug for VmSyscall {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
         match self {
+            VmSyscall::Read { fd, count } => write!(f, "read(fd={}, count={})", fd, count),
+            VmSyscall::Write { fd, count, data } => {
+                write!(f, "write(fd={}, count={}, data={:?})", fd, count, &data[..])
+            }
             VmSyscall::Madvise {
                 addr,
                 length,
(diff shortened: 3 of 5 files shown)


connorkuehl:
Hmm, our builder appears to be running into some weirdness

haraldh:
> Hmm, our builder appears to be running into some weirdness

I would say, it is picking up an upstream openssl version again

npmccallum:
@connorkuehl I rebased the openssl branch and re-ran the tests.

whitebrandy:
> he `Line` type to base this on for some changes to `memory_map.rs`; so once that type is concretely in the repo I can rebase.

Can we just not leave the Line type for later and rebase the rest of the stuff for now ?

connorkuehl:
> Can we just not leave the Line type for later and rebase the rest of the stuff for now ?

Looking back at my initial comment on the pull request I don't think it really matters which branch gets the Line commit in sooner. It's trivial to rebase and drop it off whichever branch is redundant.

connorkuehl:
@npmccallum I think I addressed the initial set of comments from our conversation the other day. Commit 1bc2fc0c8e3ae1ac1029e6524b5d6cd7f4c71d39 is my attempt at ripping out the enum to avoid matching on hypervisor-provided data inside the microkernel. It didn't end up coming across as cleanly as just an opaque `usize` struct because there are some syscalls where "output" parameters must be recorded onto the shared syscall page so the microkernel can unpack these and update the relevant datum. An example of this would be commit b5dd3fcbecc4fc6c5a16856df11eae43805f231a.

I also tried to represent appropriate return types using types from `memory` (for example, `mmap` will return an address) and these are accessible by the implementations of core::convert::Into.

connorkuehl:
Hmm, weird, I am not sure why GitHub is ordering the commits from my latest force push like that, but this patchset should be ordered like this (just for those that like to review commit-by-commit):

```
[ckuehl@redhat enarx]$ git log --reverse --oneline upstream/master..origin/vmsyscall 
5533efe Add Line type to the span crate
aee52e8 Remove syscall serialization
ed4da1e Update doc comments
0be2e5e Implement Debug for VmSyscall
d2e71f2 Replace enum representation of syscall results
85d894c Add the Read and Write syscalls
2c36a85 Derive other useful traits for Page
4a4106a Add types for working with the memory map
d57257e Add BootInfo type
```

connorkuehl:
Ping for review

cc: @npmccallum @haraldh @whitebrandy

vmsyscall/src/bootinfo.rs
```diff
@@ -0,0 +1,67 @@
...
+///    // […]
+/// }
+/// ```
+///
+/// Note that no type checking occurs for the entry point function, so be careful to
+/// use the correct argument types. To ensure that the entry point function has the correct
+/// signature, use the [`entry_point`] macro.
+#[derive(Clone)]
+#[repr(C)]
+pub struct BootInfo {
```
npmccallum, line 27: This is just a note on general strategy. Where possible, we want the guest to determine layout.  So rather than the host doing the whole layout and passing it to the guest, the host should do minimal layout and allow the guest to determine things like the code location. This gives the host less advantage on attacking the guest.

The general principle is that the host should be in charge of allocation, but the guest should be in charge of its own layout.

vmsyscall/src/bootinfo.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub memory_map: Map,
+
+    /// ELF entry point of the ring 3 executable
+    pub entry_point: *const u8,
+
+    /// ELF program headers of the ring 3 executable
+    pub load_addr: *const u8,
+
+    /// ELF number of program headers of the ring 3 executable
+    pub elf_phnum: usize,
```
npmccallum, line 43: We only support `static-pie` binaries. If the kernel passes the (randomized) code address to the host, the remaining items can be parsed from the ELF header (trivially).

vmsyscall/src/bootinfo.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// Syscall trigger port
+    pub syscall_trigger_port: u16,
```
npmccallum, line 46: I think we can use `HLT` instead of an io port for inducing VM exits.

vmsyscall/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
+//! This crate is the interface between the SEV hypervisor (keep) and
+//! microkernel (shim). It enables system call proxying to the host.
+//!
 //! Currently it uses a hard coded page and an I/O trigger.
 //! We might want to switch to MMIO.
```
npmccallum, line 7: We probably want a page and `HLT` for now.

memory/src/page.rs
```diff
@@ -3,7 +3,7 @@
 /// A single page of memory
 ///
 /// This type is page-aligned and page-sized.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
```
npmccallum, line 6: I don't understand how we can support `PartialOrd` and `Ord` in this case. Also, the default debug output isn't pretty.

npmccallum:
@connorkuehl Also, we should consider renaming this crate. Perhaps we need a common scheme:

```
enarx-keep-sev
enarx-keep-sev-lib (formerly vmsyscall)
enarx-keep-sev-shim
```

vmsyscall/src/bootinfo.rs
haraldh: Do you mean, that `load_addr` which basically is the address of [ElfHeader64](https://docs.rs/goblin/0.2.1/goblin/elf/header/struct.Header.html) already contains `entry_point` and `elf_phnum`? If so... yes.
Yesterday I implemented ASLR for the ring3 `code` with this information.
The loader can load the code anywhere in this case. The shim can then construct the page tables with ASLR.

vmsyscall/src/bootinfo.rs
haraldh: Yeah.. maybe.

vmsyscall/src/memory_map.rs
```diff
@@ -0,0 +1,283 @@
...
+            .finish()
+    }
+}
+
+/// Represents possible types for memory regions.
+#[allow(missing_docs)]
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(C)]
+pub enum RegionType {
```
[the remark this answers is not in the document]
haraldh: That is part of the measured code. I don't see how this could be exploited then?

vmsyscall/src/bootinfo.rs
haraldh: That is already the case. I reduced it to the bare minimum already.

haraldh:
> Besides the comments here, @connorkuehl and I had a call to discuss this interface.
> 
> Our primary goals for this interface are security and performance. We can't use a rich enum interface because of undefined behavior which the host could tamper with. So I proposed an alternate interface which is more similar to what we are using in SGX. This is the conceptual idea:
> 
> ```rust
> pub struct Message {
>     pub rax: Register<u64>,
>     pub rdi: Register<u64>,
>     pub rsi: Register<u64>,
>     pub rdx: Register<u64>,
>     pub r10: Register<u64>,
>     pub r8: Register<u64>,
>     pub r9: Register<u64>,
> }
> 
> pub struct Block {
>     pub msg: Message,
>     pub buf: [u8; BLOCK_SIZE - size_of::<Message>],
> }
> ```

This is why I had `serde` there in the first place... for struct integrity.

Anyway.. we have to specify `#[repr(C)]` then and only use primitive types, so we don't run into UD.

But yeah. We can always have a `impl TryFrom<Message> for SyscallMmap` with safe guards.

vmsyscall/src/memory_map.rs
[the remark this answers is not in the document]
npmccallum: My understanding is that the Map, and therefore Region and RegionType, are all part of the shared interface between the host and guest. Enums can only work in this case on read-only pages. But even then, they are likely to creep into RW pages over time. As soon as that happens, the host can set the memory to an undefined discriminator in order to exploit undefined behavior in the guest.

vmsyscall/src/bootinfo.rs
haraldh: But yeah, I think `BootInfo` can be reduced to just `&ElfHeader` of the code.

npmccallum:
> > Besides the comments here, @connorkuehl and I had a call to discuss this interface.
> > 
> > Our primary goals for this interface are security and performance. We can't use a rich enum interface because of undefined behavior which the host could tamper with. So I proposed an alternate interface which is more similar to what we are using in SGX. This is the conceptual idea:
> > 
> > ```rust
> > pub struct Message {
> >     pub rax: Register<u64>,
> >     pub rdi: Register<u64>,
> >     pub rsi: Register<u64>,
> >     pub rdx: Register<u64>,
> >     pub r10: Register<u64>,
> >     pub r8: Register<u64>,
> >     pub r9: Register<u64>,
> > }
> > 
> > pub struct Block {
> >     pub msg: Message,
> >     pub buf: [u8; BLOCK_SIZE - size_of::<Message>],
> > }
> > ```
> 
> This is why I had `serde` there in the first place... for struct integrity.
> 
> Anyway.. we have to specify `#[repr(C)]` then and only use primitive types, so we don't run into UD.
> 
> But yeah. We can always have a `impl TryFrom<Message> for SyscallMmap` with safe guards.

I didn't see that as needed. Can you explain that to me? I might be missing something.

haraldh:
> > > Besides the comments here, @connorkuehl and I had a call to discuss this interface.
> > > Our primary goals for this interface are security and performance. We can't use a rich enum interface because of undefined behavior which the host could tamper with. So I proposed an alternate interface which is more similar to what we are using in SGX. This is the conceptual idea:
> > > ```rust
> > > pub struct Message {
> > >     pub rax: Register<u64>,
> > >     pub rdi: Register<u64>,
> > >     pub rsi: Register<u64>,
> > >     pub rdx: Register<u64>,
> > >     pub r10: Register<u64>,
> > >     pub r8: Register<u64>,
> > >     pub r9: Register<u64>,
> > > }
> > > 
> > > pub struct Block {
> > >     pub msg: Message,
> > >     pub buf: [u8; BLOCK_SIZE - size_of::<Message>],
> > > }
> > > ```
> > 
> > 
> > This is why I had `serde` there in the first place... for struct integrity.
> > Anyway.. we have to specify `#[repr(C)]` then and only use primitive types, so we don't run into UD.
> > But yeah. We can always have a `impl TryFrom<Message> for SyscallMmap` with safe guards.
> 
> I didn't see that as needed. Can you explain that to me? I might be missing something.

Isn't this the same argument as "we can't use rust enums"?

vmsyscall/src/bootinfo.rs
haraldh: Ah, no... I need at least the reserved virtual memory range.

haraldh: The rest is nice to have, so I don't have to calculate it myself... like the range of the `shim` and the range of the `code`.

connorkuehl:
This PR was superceded by smaller PRs. The issue this is linked to is being promoted to track feature parity with `demo2020` instead. Further parity will emerge as more work gets merged to the keep and shim.

PR #420, opened 2020-04-15
Integrate 'cargo make'

### LINKED ISSUE #417: Experiment with 'cargo make' integration
We've outgrown the workspace and our unique loading & testing needs require something with more flexibility.

See if 'cargo make' is a viable tool. Start with getting our entire tree building.

https://sagiegurari.github.io/cargo-make/

connorkuehl:
Closes #417 

Initial draft description follows below:

Note: I am _not_ proposing this for merge yet and so none of the commits are cleaned up.

Also, I anticipate when I open this pull request, the CI probably won't work, as I don't want to propose changes to our containers to install 'cargo make' out-of-band of this draft pull request.

I've only been running builds and tests, so I haven't ran a keep or done anything useful with the builds that ensue from here.

**I wanted to open this up to the floor for some discussion on things that we want to know 'cargo make' can do before we seriously consider it.**

The primary goal here is to move _away_ from the workspace. There are certain things like feature coalescing that prevent us from using already established crates like `libc` in `no_std` mode for some crates and in `std` mode for others. This results in us carrying our own ad-hoc `nolibc`. By moving away from the workspace, we are not constrained by something like that anymore. The transition from `nolibc` to `libc` `no_std` is not yet included on this branch.

As far as the primary goal, we can accomplish that with 'cargo make' or just a regular Makefile. I would argue that 'cargo make' comes with some quality of life improvements.

For example, we depend on a number of binary cargo packages for checking license compatibility or auditing for downstream vulnerabilities. 'Cargo make' targets support instructing it to install cargo binary packages if it is not already installed on the user's system. 

Cargo make also has a number of recipes already defined implicitly. We've got the usual ones (build, test, clean, etc) but it also defines a CI target which prints more information about the environment.

All of these targets can be used out of the box, or extended, or overridden completely. New targets may be added.

On this branch I've overridden the 'test' target and have supplied a new target for checking license compatibility. I'd imagine this will be useful for integrating this into our CI by just delegating certain cargo runs to Cargo Make targets.

I did experiment a little and 'cargo make' _does_ check for a `Makefile.toml` in the current working directory and if detected, it will use that.

My current solution just has 1 Makefile with the custom targets so I don't need to symlink or copy it into the crate directories. I use a bash script as a wrapper to find the crates and then it will descend into each crate directory and run 'cargo make' with whatever argument you passed into it. I don't know if this recursive descent is already feature of 'cargo make' at the time of this writing, which is why I worked around it in a bash script.

For example:

```bash
$ cargo install cargo-make
$ ./cargo-make-recursive.sh ci-flow
$ ./cargo-make-recursive.sh clean
$ ./cargo-make-recursive.sh build
```


whitebrandy:
> The primary goal here is to move _away_ from the workspace. There are certain things like feature coalescing that prevent us from using already established crates like `libc` in `no_std` mode for some crates and in `std` mode for others. This results in us carrying our own ad-hoc `nolibc`. By moving away from the workspace, we are not constrained by something like that anymore. The transition from `nolibc` to `libc` `no_std` is not yet included on this branch.

A layman question - I am curious to understand why can't we mix modes with libc currently and how does cargo make help ?

connorkuehl:
> A layman question - I am curious to understand why can't we mix modes with libc currently and how does cargo make help ?

Dependency features are coalesced throughout the cargo workspace. So if Crate A uses libc without any features (i.e., no std) and Crate B uses libc _with_ the "std" feature, then the "std" feature would be used for both Crate A and Crate B.

This would be a big issue for Crate A because Crate A must be "no std" and in this case, the workspace would try to force Crate A to link against std.

connorkuehl:
Ping for review.

I think 'cargo make' is a comfortable way out of the workspace. [I'm actually getting blocked by the workspace right now](https://github.com/enarx/enarx/issues/431#issuecomment-618020630) so I'm hoping to tie up any unanswered questions here so we can decide if we want to move forward with this. If we do want to move forward with it, I'll clean up the PR and propose patches against the other relevant areas that will need to be updated (such as our container repos).

cargo make's developer [has kindly implemented workspace emulation](https://github.com/sagiegurari/cargo-make/issues/415) so that we can just run something like `cargo make` at the root of our project instead of using my bash wrapper.

connorkuehl:
> I tried out the commands you listed above and everything seems to work well. Is there any other functionality you wanted us to test? Otherwise LGTM.

Thanks! I guess a good test would be building the workspace and seeing that our SGX keep still runs to make sure no linker-y bits have been ruined.

connorkuehl:
Blocked on https://github.com/enarx/docker-fedora/pull/4

connorkuehl:
Once the latest test container materializes on quay, I'll squash the commits and re-push then this should be just about polished up enough for the PR to come out of draft form.

payload/.cargo/config
```diff
@@ -0,0 +1,7 @@
+[build]
+target = "x86_64-unknown-linux-musl"
+rustflags = [
+    "-C", "relocation-model=pic",
+    "-C", "opt-level=3",
```
haraldh, line 5: I don't think we need `opt-level=3` here by default.

connorkuehl: Addressing this in my next push

.github/workflows/cargo.yml
```diff
@@ -3,68 +3,59 @@ name: cargo
...
-      run: cargo test
+      - uses: actions/checkout@v2
+      - name: No unresolved reported vulnerabilities.
+        run: cargo make audit
 
-    # Run Rust tests with OpenSSL-specific features enabled for SEV.
-    - name: Run tests (SEV-specific)
-      run: cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
-
-  # Lint check with rust-clippy.
```
mbestavros, line 42: @connorkuehl Is this test no longer needed?

connorkuehl: @mbestavros It's still there, the diff is misleading because when I re-wrote the tests with 'cargo-make' I put them in a different ordering from top-to-bottom.

connorkuehl: Er woops, I didn't read closely enough. Let me add this back in

mbestavros: I don't see an SEV-specific test with OpenSSL features enabled in the new diff, though. I'm wondering if that tests is obsolete. If not, we should have a `make` task for that.

.github/workflows/cargo.yml
```diff
@@ -3,68 +3,59 @@ name: cargo
...
-    # Run any specified Rust tests.
-    - name: Run tests
-      run: cargo test
+      - uses: actions/checkout@v2
+      - name: No unresolved reported vulnerabilities.
+        run: cargo make audit
 
-    # Run Rust tests with OpenSSL-specific features enabled for SEV.
-    - name: Run tests (SEV-specific)
-      run: cargo test --manifest-path sev/Cargo.toml --verbose --features=openssl
```
mbestavros, line 40: This test needs to be added back in.

Makefile.toml
```diff
@@ -0,0 +1,107 @@
...
+		"misc-diagrams",
+		"cargo-toml-package-license",
+		"misc-licenses-crate",
+		"check-format",
+		"deny",
+		"audit",
+		"clippy",
+		"build",
+		"test"
+] }]
```
mbestavros, line 107: This is really handy. :+1:

.github/workflows/cargo.yml
connorkuehl: You're quite right, my order of operations here was bad and my force push came after my lunch break. I've amended the PR to retain the 'openssl' feature enablement for sev.

.github/workflows/cargo.yml
connorkuehl: It's not added back in as a separate run step under GitHub actions, but the `Makefile.toml` inside the `sev` crate extends the `cargo test` target to enable the feature here: https://github.com/enarx/enarx/pull/420/files#diff-8ed61aa622ed1d7f91c707e0fd48d01d

Makefile.toml
```diff
@@ -0,0 +1,107 @@
...
+command = "cargo"
+args = ["deny", "check", "licenses"]
+
+[tasks.test]
+command = "cargo"
+args = ["test"]
+
+[tasks.check-format]
+command = "cargo"
+args = ["fmt", "--", "--check"]
```
npmccallum, line 44: `install_crate = "cargo-fmt"`?

Makefile.toml
```diff
@@ -0,0 +1,107 @@
...
+args = ["audit"]
+
+[tasks.deny]
+install_crate = "cargo-deny"
+command = "cargo"
+args = ["deny", "check", "licenses"]
+
+[tasks.test]
+command = "cargo"
+args = ["test"]
```
npmccallum, line 40: Is this overriding the default `tasks.test`? If so why? Do you need to strip out the additional functionality for some reason?

https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/rust.toml#L375

Makefile.toml
npmccallum: Is there a reason we can't use this implicitly? https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/rust.toml#L205

Makefile.toml
```diff
@@ -0,0 +1,107 @@
...
+	"testing",
+	"units",
+	"vdso",
+	"vmsyscall"
+]
+
+[tasks.audit]
+install_crate = "cargo-audit"
+command = "cargo"
+args = ["audit"]
```
npmccallum, line 31: Can't we use this?

https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/rust.toml#L544

Makefile.toml
```diff
@@ -0,0 +1,107 @@
...
+[tasks.test]
+command = "cargo"
+args = ["test"]
+
+[tasks.check-format]
+command = "cargo"
+args = ["fmt", "--", "--check"]
+
+[tasks.misc-lints-missing-docs]
+workspace = false
```
npmccallum, line 47: What does `workspace = false` do here?

Makefile.toml
npmccallum: This section seems to me to be patching `tasks.ci` just in order to add additional custom tests. But I think [there is already a hook for that](https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/rust.toml#L394). If so, using the hook would be better because we can then pick up upstream changes to `tasks.ci` automatically.

sev/Makefile.toml
```diff
@@ -0,0 +1,3 @@
+[tasks.test]
+command = "cargo"
+args = ["test", "--features=openssl"]
```
npmccallum, line 3: Does `cargo make` offer us the ability to run tests on all combinations of feature flags?

span/src/lib.rs
```diff
@@ -14,8 +14,8 @@
 mod line;
 mod span;
 
-pub use line::Line;
-pub use span::Span;
+pub use crate::line::Line;
+pub use crate::span::Span;
```
npmccallum, line 18: I don't understand this patch. Where is the ambiguity? We clearly defined the modules in the previous lines.

iocuddle/src/lib.rs
```diff
@@ -46,7 +46,7 @@
 //! Otherwise, classic `ioctl` interfaces can be defined and used via the
 //! `Ioctl::classic()` constructor as follows:
 //!
-//! ```
+//! ```no_run
```
npmccallum, line 49: Hrm. I carefully designed this test to use device nodes that should be available everywhere. What causes this error?

Makefile.toml
connorkuehl: I think plain `cargo fmt` changes the source files and then returns `0` as its exit code which would allow our CI test to pass even if the source code is not properly formatted. `cargo fmt --check` will exit with a non-zero code which will be noticed by CI.

Makefile.toml
connorkuehl: Ah true. The default `test` target turns on all features for the crate, which I wanted to avoid the `dangerous_tests` features for SEV. However, I can remove this from the root Makefile and specify not to use dangerous_tests by default in the `sev` crate Makefile.

Makefile.toml
connorkuehl: I didn't see that one! I will remove this.

Makefile.toml
connorkuehl: This prevents `cargo-make` from descending into member crates. I removed that because I think our bash CI scripts glob the crates from the root-level directory.

Makefile.toml
npmccallum: @connorkuehl Since upstream is so active, we should add a `--check` version of that to their built-in test flow. That way:
1. We don't have to maintain it.
2. Others can benefit from it.

However, I'm willing to take this as a `debt` issue to keep our momentum.

Makefile.toml
npmccallum: I think it means that it doesn't recurse into subprojects. Correct?

sev/Makefile.toml
npmccallum: I think that would be a very powerful upstream feature. But, again, I'm happy to take that as `debt`.

enarx-keep-sgx-shim/.cargo/config
```diff
@@ -3,5 +3,5 @@ target = "x86_64-unknown-linux-musl"
 rustflags = [
     "-C", "relocation-model=pic",
     "-C", "opt-level=3",
-    "-C", "linker=./cc",
+    "-C", "linker=../cc",
 ]
```
npmccallum, line 7: Just a note that we can probably build the shim on any `libc` since we enforce `-nostdlib`. `debt` `enhancement`

npmccallum: Actually, that probably conflicts with moving the shim to `static-pie` and using the musl relocation.

span/src/lib.rs
connorkuehl: I started seeing this fail in the `cargo make test` target:

```bash
[cargo-make][1] INFO - Project: span
[cargo-make][1] INFO - Build File: Makefile.toml
[cargo-make][1] INFO - Task: test
[cargo-make][1] INFO - Profile: development
[cargo-make][1] INFO - Running Task: test
[cargo-make][1] INFO - Execute Command: "cargo" "test"
   Compiling span v0.1.0 (/home/ckuehl/src/enarx/span)
    Finished test [unoptimized + debuginfo] target(s) in 0.35s
     Running target/debug/deps/span-e770ea83353494aa

running 8 tests
test line::test::convert ... ok
test line::test::contains ... ok
test line::test::is_empty ... ok
test line::test::split ... ok
test span::test::contains ... ok
test span::test::convert ... ok
test span::test::is_empty ... ok
test span::test::split ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests span
error[E0659]: `span` is ambiguous (name vs any other name during import resolution)
  --> /home/ckuehl/src/enarx/span/src/lib.rs:18:9
   |
18 | pub use span::Span;
   |         ^^^^ ambiguous name
   |
   = note: `span` could refer to a crate passed with `--extern`
   = help: use `::span` to refer to this crate unambiguously
note: `span` could also refer to the module defined here
  --> /home/ckuehl/src/enarx/span/src/lib.rs:15:1
   |
15 | mod span;
   | ^^^^^^^^^
   = help: use `crate::span` to refer to this module unambiguously

error: aborting due to previous error

For more information about this error, try `rustc --explain E0659`.
error: test failed, to rerun pass '--doc'
[cargo-make][1] ERROR - Error while executing command, exit code: 1
[cargo-make][1] WARN - Build Failed.
[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
[ckuehl@redhat enarx]$ rustc --version
rustc 1.43.0 (4fb7144ed 2020-04-20)
```

iocuddle/src/lib.rs
connorkuehl: I actually can't reproduce this in a container or on my laptop, which is unsurprising because the device node is present in both. I must have gotten my wires crossed somewhere when I was playing around with my Enarx tree. I'll drop this patch.

span/src/lib.rs
npmccallum: I wonder if that is a `1.43.0`ism. It appears to be ambiguous on the crate vs the module. I think I'd prefer to rename the crate.

Makefile.toml
connorkuehl: This is my mistake again, there is indeed a builtin for this. However, the `install_crate` step is gated behind nightly (which used to be the case before `rustfmt` was included in the stable toolchain). I've [filed an issue upstream](https://github.com/sagiegurari/cargo-make/issues/418) to remove this reqiurement.

npmccallum: I'm aware it is installed on recent rust toolchain installs. But we might have old toolchain installs or even people who have removed it.

.github/workflows/cargo.yml
```diff
@@ -86,4 +77,4 @@ jobs:
     
     # Ensure Cargo.toml's license field is Apache 2.0.
     - name: All Cargo.toml files have a license field
-      run: .tests/cargo-toml-package-license
+      run: cargo make cargo-toml-package-license
```
npmccallum, line 80: Do we want to reduce this whole file to just `cargo make test` and have that task run all the tests?

1. The downside is that we lose per-test indication in GitHub.
2. One upside is that we don't have to manage which tests are required in GitHub.
3. Another upside is that we could run all tests locally with a single command, making dev easier.

.github/workflows/miscellaneous.yml
```diff
@@ -45,19 +47,21 @@ jobs:
...
   misc-lints-missing-docs:
     runs-on: ubuntu-latest
+    container: quay.io/enarx/fedora-test
     steps:
     - uses: actions/checkout@v2
 
     # Ensure that the `missing_docs` lint is enabled.
     - name: All Rust crates have `missing_docs` enabled
-      run: .tests/misc-lints-missing-docs
+      run: cargo make misc-lints-missing-docs
```
npmccallum, line 67: Same here. Do we want to subsume this under `cargo make test`?

Makefile.toml
```diff
@@ -0,0 +1,77 @@
...
+	"units",
+	"vdso",
+	"vmsyscall"
+]
+CARGO_WORKSPACE_ROOT = { script = ["git rev-parse --show-toplevel"] }
+
+[tasks.deny]
+install_crate = "cargo-deny"
+command = "cargo"
+args = ["deny", "check", "licenses"]
```
npmccallum, line 32: As `debt`, we might consider pushing this upstream to `cargo-make`.

Makefile.toml
```diff
@@ -0,0 +1,77 @@
...
+
+[tasks.misc-lints-missing-docs]
+command = "${CARGO_WORKSPACE_ROOT}/.tests/misc-lints-missing-docs"
+
+[tasks.cargo-toml-package-edition]
+install_script = [
+'''
+which toml || cargo install toml-cli
+'''
+]
```
npmccallum, line 42: This is a lot of wasted vertical space. Can we make this a single line?

iocuddle/src/lib.rs
```diff
@@ -50,7 +50,11 @@
 //! use std::os::raw::{c_void, c_int, c_uint};
 //! use iocuddle::*;
 //!
-//! let mut file = std::fs::File::open("/dev/tty").unwrap();
+//! let file = std::fs::File::open("/dev/tty");
+//! let mut file = match file {
+//!     Ok(f) => f,
+//!     Err(_) => std::process::exit(1),
+//! };
```
npmccallum, line 57:
```suggestion
//! let mut file = match std::fs::File::open("/dev/tty") {
//!     Err(_) => std::process::exit(1),
//!     Ok(f) => f,
//! };
```

I find that more readable.

npmccallum:
5. We also need documentation for how to build and test.

iocuddle/src/lib.rs
haraldh: why not `unwrap_or_else`???

npmccallum: That would indeed have been better. PR welcome. :)

haraldh: https://github.com/enarx/enarx/pull/491

PR #421, opened 2020-04-15
Add memory crate

npmccallum:
This provides various different utility types for managing memory.


npmccallum:
@connorkuehl This implements (I think) everything we discussed today. Please review and let me know if I've forgotten anything.

For example, the `Frame` type can now be expressed as `Offset<usize, Page>`. These types should be usable in `span::Span` and `span::Line`.

Issue #334, opened 2020-03-16
save/restore/clear extended CPU state


npmccallum:
Since #326, we can now context switch between an enclave and the outside world. We don't currently do anything with the extended CPU state. The workflow should look like this:

1. Clear exCPU state when entering and exiting enclave.
2. Save exCPU state before calling syscall().
3. Restore exCPU state after syscall() returns.

lkatalin:
@npmccallum I'm looking through the [doc](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3d-part-4-manual.pdf), 42.7 INTERACTIONS WITH THE PROCESSOR EXTENDED STATE AND MISCELLANEOUS STATE. It describes the way `xfrm` specifies which extended states to save in SSA on an AEX, among other things. Does this sound right?

Specifically, it's describing:

- At enclave entry, `xcr0` is saved internally to the enclave and replaced by `xfrm`. On exit, `xcr0` is restored. Is this the same as (1)?
- On an AEX, `xfrm`-enabled states are saved into SSA. Is this the same as (2)?
- On resuming from AEX, the states from SSA are restored. Is this the same as (3)?

You also mentioned our asm function can take a pointer to a Rust buffer to set/clear the state. Does this correspond to all 3 workflow steps?

npmccallum:
> @npmccallum I'm looking through the [doc](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3d-part-4-manual.pdf), 42.7 INTERACTIONS WITH THE PROCESSOR EXTENDED STATE AND MISCELLANEOUS STATE. It describes the way `xfrm` specifies which extended states to save in SSA on an AEX, among other things. Does this sound right?
> 
> Specifically, it's describing:
> 
> * At enclave entry, `xcr0` is saved internally to the enclave and replaced by `xfrm`. On exit, `xcr0` is restored. Is this the same as (1)?
> * On an AEX, `xfrm`-enabled states are saved into SSA. Is this the same as (2)?
> * On resuming from AEX, the states from SSA are restored. Is this the same as (3)?

No. None of this will be in the manual. It is presumed that we know what we're doing.

Specifically, we need to use the following instructions:
* https://www.felixcloutier.com/x86/xsaveopt (or one of its variants)
* https://www.felixcloutier.com/x86/xrstor

[...]

PR #427, opened 2020-04-16
Clear, save, restore extended CPU state

lkatalin:
This has a linking error:
`/enarx/enarx-keep-sgx-shim/src/start.S:144:(.text+0x32): relocation truncated to fit: R_X86_64_32S against `.rodata`

Creating a read only data symbol with `.rodata` seems like it would be a good way to do this, but it may not be.

Edit: RIP-relative addressing fixed the linking error. :tada:

### DIFF
--- a/enarx-keep-sgx-shim/src/start.S
+++ b/enarx-keep-sgx-shim/src/start.S
@@ -60,6 +60,28 @@
     pop     %rbx
 .endm
 
+# Clear the extended CPU state (clobbers %rax and %rdx)
+.macro zerox
+    movq    $~0,                    %rdx            # set mask for xrstor in rdx
+    movq    $~0,                    %rax            # set mask for xrstor in rax
+    xrstor  xsave(%rip)                             # clear xtd cpu state with synthetic state
+.endm
+
+    .section .rodata
+    .align 64
+xsave:                          # An initialized synthetic xsave area
+# Legacy
+    .fill   1, 4, 0x037F        # FCW
+    .fill   5, 4, 0
+    .fill   1, 4, 0x1F80        # MXCSR
+    .fill   1, 4, 0xFFFF        # MXCSR_MASK
+    .fill   60, 8, 0
+
+# Header
+    .fill   1, 8, 0             # XSTATE_BV
+    .fill   1, 8, 1 << 63       # XCOMP_BV (compaction mode)
+    .fill   6, 8, 0
+
 # This function is called during EENTER. Its inputs are as follows:
 #  %rax = The current SSA index. (i.e. %rbx->cssa)
 #  %rbx = The address of the TCS.
@@ -76,10 +98,15 @@ enclave:
     cmp     $0,                     %rax            # If CSSA > 0...
     jne     .Levent                                 # ... restore stack from AEX[CSSA-1].
 
-    # Clear unused registers, set the stack pointer and jump to Rust
+    # Clear unused registers and clear extended CPU state
     zerop
     zerot
     xor     %rax,                   %rax
+    push    %rdx
+    zerox
+    pop     %rdx
+
+    # Set the stack pointer and jump to Rust
     mov     STACK(%rcx),            %rsp
     jmp     entry
 
@@ -130,6 +157,7 @@ enclave:
 .Leexit:
     loadp                                           # Load preserved registers
     pop     %rsp                                    # Restore the untrusted stack
+    zerox                                           # Clear the extended CPU state
     mov     $4,                     %rax
     enclu
 
@@ -142,8 +170,14 @@ enclave:
 .Lsyscall:
     movq    $0,                     SRSP(%r11)      # Clear syscall return stack pointer field
     mov     %rax,                   %rsp            # Restore the syscall return stack pointer
-    mov     %rdi,                   %rax            # Correct syscall return value register
     loadp                                           # Restore trusted preserved registers
+
+    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
+    movq    $~0,                    %rax            # Set mask for xrstor in rax
+    xrstor  (%rsp)                                  # Restore extended CPU state
+    mov     %rbx,                   %rsp            # Restore stack after aligned xsave area
+
+    mov     %rdi,                   %rax            # Correct syscall return value register
     zeroa                                           # Clear the argument registers
     zerot                                           # Clear the temporary registers
     zerof   %r11                                    # Clear CPU flags
@@ -154,6 +188,27 @@ enclave:
     .globl syscall
     .type syscall, @function
 syscall:
+    # Save stack pointer, allocate space, and set
+    # 64-byte alignment on stack for xsavec
+    mov     %rsp,                   %rbx
+    sub     $4096,                  %rsp
+    and     $(-0x40),               %rsp
+
+    # Zero the xsave header area on stack
+    movq    $0,        512(%rsp)
+    movq    $0,        520(%rsp)
+    movq    $0,        528(%rsp)
+    movq    $0,        536(%rsp)
+    movq    $0,        544(%rsp)
+    movq    $0,        552(%rsp)
+    movq    $0,        560(%rsp)
+    movq    $0,        568(%rsp)
+
+    # Set mask and perform xsavec
+    movq    $~0,                   %rdx
+    movq    $~0,                   %rax
+    xsavec (%rsp)
+
     savep                                           # Save preserved registers
     mov     %rsp,                   SRSP(%rcx)      # Save restoration stack pointer
 


lkatalin:
> Maybe name it `zerox`?

Happily.

haraldh:
One thing, which came to my mind, is that in future CPUs, there might be more registers, which should be cleared.

I don't know, if it is technically feasible.

```rust
 let xsave_size = core::arch::x86_64::__cpuid(0xD).ebx;
```
You would need `xsave_size` bytes and would dump a full `xsave` there on startup, which you would restore on.

Also `xsavebuff` might be too small. At least an assert!() is needed if `xsave_size` exceeds the size of `xsavebuff`.

haraldh:
I think you don't have `cpuid` in the enclave, but @npmccallum knows.

haraldh:
For the reference:
https://static.sched.com/hosted_files/lsseu2019/65/SEV-SNP%20Slides%20Nov%201%202019.pdf
Slide 18:

> SEV-SNP adds new capabilities for CPUID filtering so guests can ensure that security sensitive information is correct
> For example: XSAVE state size is considered security sensitive to avoid potential buffer overflows

... might not be a good idea to ask the host for the xsave size ...

lkatalin:
@haraldh Thanks for the review. I've looked more closely now at your suggestion. You were right in your follow-up: we do not have access to `CPUID` from the enclave. Currently we're using a synthetic `CPUID` result: https://github.com/enarx/enarx/blob/master/enarx-keep-sgx-shim/src/event.rs#L63 I assume there is no way to check how many registers should be cleared, and the `xsave` size, without this? In any case, it does seem like we definitely do not want to trust the host to provide this info.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -142,6 +174,9 @@ enclave:
 .Lsyscall:
     movq    $0,                     SRSP(%r11)      # Clear syscall return stack pointer field
     mov     %rax,                   %rsp            # Restore the syscall return stack pointer
+    movq    $~0,                    %rax            # Set mask for xrstor in rax
+    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
```
haraldh, line 178: You should probably only enable the supported bits from https://github.com/enarx/enarx/blob/master/sgx-types/src/attr.rs#L37 so that the size is not exceeded.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +198,9 @@ syscall:
 
     xor     %rcx,                   %rcx            # Clear %rcx
     zerof   %rcx                                    # Clear CPU flags
+    movq    $~0,                    %rax            # Set mask for xsaveopt in rax
+    movq    $~0,                    %rdx            # Set mask for xsaveopt in rdx
```
haraldh, line 202: You should probably only enable the supported bits from https://github.com/enarx/enarx/blob/master/sgx-types/src/attr.rs#L37 so that the size is not exceeded.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -60,6 +60,33 @@
...
+ 
+# Header
+    .fill   1, 8, 0             # XSTATE_BV
+    .fill   1, 8, 1 << 63       # XCOMP_BV (compaction mode)
+    .fill   6, 8, 0
+    
+    .section .data
+xsavebuff:                      # An empty buffer to save xsave state
+    .align 64
+    .fill 512, 1, 0
```
npmccallum, line 88: Since this value is a global, it could only be used by one thread. That works today, but would break as soon as we added a second thread. Instead, `xsavebuff` should just be put on the stack  in `syscall` and loaded from the stack after it is restored during `.Lsyscall`.

haraldh:
```console
$ cpuid -1 -l 0xD
CPU:
   XSAVE features (0xd/0):
      XCR0 lower 32 bits valid bit field mask = 0x0000001f
      XCR0 upper 32 bits valid bit field mask = 0x00000000
         XCR0 supported: x87 state            = true
         XCR0 supported: SSE state            = true
         XCR0 supported: AVX state            = true
         XCR0 supported: MPX BNDREGS          = true
         XCR0 supported: MPX BNDCSR           = true
         XCR0 supported: AVX-512 opmask       = false
         XCR0 supported: AVX-512 ZMM_Hi256    = false
         XCR0 supported: AVX-512 Hi16_ZMM     = false
         IA32_XSS supported: PT state         = false
         XCR0 supported: PKRU state           = false
         XCR0 supported: CET_U state          = false
         XCR0 supported: CET_S state          = false
         IA32_XSS supported: HDC state        = false
      bytes required by fields in XCR0        = 0x00000440 (1088)
      bytes required by XSAVE/XRSTOR area     = 0x00000440 (1088)
```

haraldh:
>       XCR0 lower 32 bits valid bit field mask = 0x0000001f
>       XCR0 upper 32 bits valid bit field mask = 0x00000000
which means `XCR0 AND EDX:EAX` with `EDX=0` and `EAX=0x1f` would result in 1088 bytes.

A safety check might be reading XCR0 and see, if there are additional bits set by default and refuse to continue.

Future Rust could use future CPU extra registers.

We might want to have some safety checks somewhere...

enarx-keep-sgx-shim/src/start.S
lkatalin: @haraldh This is a good point. As far as I understand it, we can't read from `SECS`, so we can't access the current `xfrm` value. So I've done the next best thing - our default `xfrm` value sets bits `1:0` for `x87` and `SSE`, and no others. The new mask sets these same bits.

enarx-keep-sgx-shim/src/start.S
lkatalin: I've fixed it to do this, in probably an awkward way, because `xrstor` and `xsaveopt` only accept values in memory as their operands, for some reason.

lkatalin:
> > ```
> >   XCR0 lower 32 bits valid bit field mask = 0x0000001f
> >   XCR0 upper 32 bits valid bit field mask = 0x00000000
> > ```
> 
> which means `XCR0 AND EDX:EAX` with `EDX=0` and `EAX=0x1f` would result in 1088 bytes.
> 
> A safety check might be reading XCR0 and see, if there are additional bits set by default and refuse to continue.
> 
> Future Rust could use future CPU extra registers.
> 
> We might want to have some safety checks somewhere...

@haraldh Does using a mask of only bits `1:0` for `EDX:EAX` address this?

lkatalin:
@npmccallum @haraldh This is a WIP, but I would appreciate more comments, especially on the following:

1. @haraldh pointed out that we should base our mask on the `xfrm` values. However, I believe we can't access these directly from `SECS`. So I've used a mask of `0x3` because that's our default `xfrm` value. I *do not* think this is a great solution, because the `xfrm` value could change. Any thoughts?

2. I do not know why I get an `operand mismatch` linking error for both `xrstor %rsp` and `xsaveopt %rsp`. It will only accept `xrstor (%rsp)`, for example. Due to this, I'm awkwardly trying to store the previous `%rsp` value (that is, the beginning address of the `xsave buffer`, which is now on the stack) on the stack, use it, and then pop it. This seems terrible, but I don't have other ideas at the moment.

3. I'd like to know if @haraldh 's concerns about the `xsave` size, future CPUs having more registers, etc., still needs to be addressed - if we're only saving and restoring what is specified in the default `xfrm` value, is that enough to know the size won't be exceeded?

enarx-keep-sgx-shim/src/start.S
```diff
@@ -154,15 +187,42 @@ enclave:
...
+
+    # Clear %rcx and CPU flags
+    xor     %rcx,                   %rcx
+    zerof   %rcx
+
+    # Save stack pointer, allocate space, and set
+    # 64-byte alignment on stack for xsavec
+    mov     %rsp,                   %r8
+    sub     $4096,                  %rsp
+    and     $0x40,                  %rsp
```
haraldh, line 209: This should be `and $(-0x40), %rsp` or `and $(~(0x40-1)), %rsp`

enarx-keep-sgx-shim/src/start.S
```diff
@@ -142,6 +169,12 @@ enclave:
 .Lsyscall:
     movq    $0,                     SRSP(%r11)      # Clear syscall return stack pointer field
     mov     %rax,                   %rsp            # Restore the syscall return stack pointer
+
+    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
+    movq    $~0,                    %rax            # Set mask for xrstor in rax
+    xrstor  (%rsp)                                  # Restore extended CPU state
+    mov     %r8,                    %rsp            # Restore stack alignment
```
haraldh, line 176: is `%r8` saved somewhere?

lkatalin: This is a good question. I'm not sure how to tell which registers are guaranteed to not be used during the syscall. Any advice? Or maybe I can put the old `%rsp` on the newly-aligned stack?

lkatalin: @haraldh Sorry, maybe you are asking for context. The pre-alignment `%rsp` is saved in `%r8` before the syscall and I'm only assuming it will still be there after the syscall.

lkatalin: The value is saved here: https://github.com/enarx/enarx/pull/427/files#diff-c44d5e117cccb73d5356de2ed1a1b8a2R207

lkatalin: @haraldh I see you're using `%rbx` [here](https://github.com/haraldh/enarx/blob/sev_make/enarx-keep-sev-shim/src/arch/x86_64/asm/syscall.s#L27) - maybe that one is better in this case too.

haraldh: I don't know.. just asking, if `%r8` has to be pushed on `%rsp` later also, or if that is the case anyway.

haraldh: I mean, if it is saved and restored on enclave exit/enter anyway.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -154,15 +187,42 @@ enclave:
...
+
+    # Export %r10, %rax in %r11, and get exit
+    # context
+    mov     0x38(%rsp),             %r10
+    mov     0x40(%rsp),             %r11
+    mov     0x48(%rsp),             %rsp
+
+    # Clear %rcx and CPU flags
+    xor     %rcx,                   %rcx
+    zerof   %rcx
```
npmccallum, line 203: Please don't change comments unnecessarily. Assembly is hard enough to get right. We don't need to add patch churn to the problem.

haraldh: For the reference this is `xcr0`, no `cr0`

lkatalin: Sure, I can put it back. I was trying to make it more readable with the huge chunk zeroing the xsave header.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +196,28 @@ syscall:
 
     xor     %rcx,                   %rcx            # Clear %rcx
     zerof   %rcx                                    # Clear CPU flags
+
+    # Save stack pointer, allocate space, and set
+    # 64-byte alignment on stack for xsavec
+    mov     %rsp,                   %r8
```
npmccallum, line 202: What is the purpose of this line? The `%r8` register should be passed directly to the outside world. But you're modifying it here.

npmccallum: Oh, I see what you're trying to do. You can't store `%rsp` in `%r8` for reuse later in `.Lsyscall`.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +196,28 @@ syscall:
...
+    movq    $0,        536(%rsp)
+    movq    $0,        544(%rsp)
+    movq    $0,        552(%rsp)
+    movq    $0,        560(%rsp)
+    movq    $0,        568(%rsp)
+
+    # Set mask and perform xsavec
+    movq    $~0,                   %rdx
+    movq    $~0,                   %rax
+    xsavec (%rsp)
```
haraldh, line 219:
```suggestion
    xsavec (%rsp)
    pushq  %rbx
```

enarx-keep-sgx-shim/src/start.S
```diff
@@ -142,6 +169,12 @@ enclave:
 .Lsyscall:
     movq    $0,                     SRSP(%r11)      # Clear syscall return stack pointer field
     mov     %rax,                   %rsp            # Restore the syscall return stack pointer
+
+    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
+    movq    $~0,                    %rax            # Set mask for xrstor in rax
+    xrstor  (%rsp)                                  # Restore extended CPU state
```
haraldh, line 175:
```suggestion
    popq    %rbx
    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
    movq    $~0,                    %rax            # Set mask for xrstor in rax
    xrstor  (%rsp)                                  # Restore extended CPU state
```

enarx-keep-sgx-shim/src/start.S
haraldh: save and restore %rbx ... otherwise it will be overriden by the non-enclave-rbx

enarx-keep-sgx-shim/src/start.S
haraldh: save and restore %rbx ... otherwise it will be overriden by the non-enclave-rbx

enarx-keep-sgx-shim/src/start.S
lkatalin: @haraldh Thanks. I was trying to think through which (if any) registers are preserved when we do a syscall. `%rbx` is callee-saved, but I guess we don't trust any "saved" info from the host anyway.

npmccallum: @lkatalin There are no preserved registers across a trust boundary. Calling conventions work because you trust the thing you call not to do what it isn't supposed to do. We can't presume that in this case.

lkatalin: @npmccallum Thanks, that makes more sense the more I think about it.

npmccallum: @lkatalin Anything that you want to save needs to be pushed onto the trusted stack before exit and then popped off the trusted stack when you return.

lkatalin:
@npmccallum @haraldh Wondering what you think about this. I put in a check for the `xcr0` flags. I'm assuming the known, okay flags are bits 0:9 per the [wiki](https://en.wikipedia.org/wiki/Control_register): x87, SSE, AVX, BNDREG, BNDCSR, OPMASK, ZMM_HI256, HI16_ZMM, and PKRU.

Should it do something other than go to `.Leexit` if any unknown flags are detected?

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +197,40 @@ syscall:
 
     xor     %rcx,                   %rcx            # Clear %rcx
     zerof   %rcx                                    # Clear CPU flags
+
+    # Check for unexpected flags in xcr0 before
+    # executing xsavec
+    mov    $0,                      %rcx            # Specify xcr0 for xgetbv
+    xgetbv
+    and    $~0x3FF,                 %eax            # Mask xgetbv lowest 9 bits
```
haraldh, line 205: for 0x3FF that is: `lowest 10 bits (0-9 incl.)`

enarx-keep-sgx-shim/src/start.S
```diff
@@ -142,6 +169,13 @@ enclave:
 .Lsyscall:
     movq    $0,                     SRSP(%r11)      # Clear syscall return stack pointer field
     mov     %rax,                   %rsp            # Restore the syscall return stack pointer
+
+    popq    %rbx                                    # Retrieve saved stack pointer value
+    movq    $~0,                    %rdx            # Set mask for xrstor in rdx
+    movq    $~0,                    %rax            # Set mask for xrstor in rax
+    xrstor  (%rsp)                                  # Restore extended CPU state
+    mov     %rbx,                   %rsp            # Restore stack alignment
```
haraldh, line 177:
```suggestion
    mov     %rbx,                   %rsp            # Restore stack after aligned xsave area
```

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +197,40 @@ syscall:
...
+
+    # Check for unexpected flags in xcr0 before
+    # executing xsavec
+    mov    $0,                      %rcx            # Specify xcr0 for xgetbv
+    xgetbv
+    and    $~0x3FF,                 %eax            # Mask xgetbv lowest 9 bits
+    cmp    $0,                      %eax            # No other flags should be set
+    jne    .Leexit                                  # Exit if extra flags set low
+    cmp    $0,                      %edx           # No other flags should be set
+    jne    .Leexit                                  # Exit if extra flags set hi
```
npmccallum, line 209: Let's do this check in Rust in `entry()`.

npmccallum: My thinking here is that the flags will be established at enclave creation and can't be changed. Therefore, we only need to check it once when the enclave first starts. This also means we can use the `exit(1)` syscall directly. It also implies we don't have to waste these instructions on every context switch.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -60,6 +60,28 @@
     pop     %rbx
 .endm
 
+# Clear the extended CPU state (load synthetic state)
+.macro zerox
+    movq    $~0,                    %rdx	    # set mask for xrstor in rdx
+    movq    $~0,                    %rax	    # set mask for xrstor in rax
```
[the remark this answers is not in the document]
lkatalin: Why the switch? The mask is `rdx:rax` so it seemed more readable in that order.

enarx-keep-sgx-shim/src/start.S
lkatalin: Yeah, of course, you're right.

enarx-keep-sgx-shim/src/start.S
lkatalin: Makes sense.

enarx-keep-sgx-shim/src/start.S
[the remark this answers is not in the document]
npmccallum: You've convinced me.

lkatalin:
@npmccallum I've got this output on the branch:

```
[lkatalin@localhost enarx-keep-sgx]$ ./target/debug/enarx-keep-sgx --code ../payload/target/x86_64-unknown-linux-musl/debug/payload --shim ../enarx-keep-sgx-shim/target/x86_64-unknown-linux-musl/debug/enarx-keep-sgx-shim
arg: "/init"
env: "LANG=C"
aux: ExecFilename("/init")
aux: Platform("x86_64")
aux: Uid(1000)
aux: EUid(1000)
aux: Gid(1000)
aux: EGid(1000)
aux: PageSize(4096)
aux: Secure(false)
aux: ClockTick(100)
aux: Flags(0)
aux: HwCap(0)
aux: HwCap2(0)
aux: PHdr(35184376283200)
aux: PHent(56)
aux: PHnum(10)
aux: Random([73, 128, 91, 110, 0, 79, 149, 25, 67, 8, 197, 61, 239, 57, 191, 153])
```

enarx-keep-sgx-shim/src/start.S
```diff
@@ -60,6 +60,28 @@
     pop     %rbx
 .endm
 
+# Clear the extended CPU state (clobbers %rax and %rdx)
+.macro zerox
+    movq    $~0,                    %rdx	    # set mask for xrstor in rdx
+    movq    $~0,                    %rax	    # set mask for xrstor in rax
+    xrstor  xsave(%rip)                           # clear xtd cpu state with synthetic state
```
npmccallum, line 67: This section mixes tabs and spaces.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -162,7 +197,30 @@ syscall:
     mov     0x48(%rsp),             %rsp            # Get the exit context
 
     xor     %rcx,                   %rcx            # Clear %rcx
-    zerof   %rcx                                    # Clear CPU flags
+    zerof   %rcx                                    # Clear CPU flags
```
npmccallum, line 200: Why is this line showing up in the diff? I don't see any changes...

enarx-keep-sgx-shim/src/start.S
```diff
@@ -162,7 +197,30 @@ syscall:
     mov     0x48(%rsp),             %rsp            # Get the exit context
```
npmccallum, line 197: You `MUST NOT` modify the stack after this point. I'm honestly not sure how the below code even works. We need to save the extended CPU state immediately before or after the `savep` macro above.

enarx-keep-sgx-shim/src/start.S
lkatalin: I'm fixing it; but how can you tell?

enarx-keep-sgx-shim/src/start.S
lkatalin: It was an extra space issue; fixed it now using `git diff`.

enarx-keep-sgx-shim/src/start.S
lkatalin: I've relocated the code saving extended CPU state to immediately before `savep`.

lkatalin:
Perhaps some explanation would be useful about the `syscall` and `.Lsyscall` sections. 

`syscall`:
The stack pointer is saved because (if I understand correctly) we will lose it in the process of aligning the trusted stack to 64-byte as required for `xsavec`. It's saved in `%rbx` (this is the same one used in SEV, I believe) and that value is pushed onto the trusted stack during the `savep` that occurs later. (Previously, there was a `pushq %rbx`, but I realized this is redundant). 

The xsave header is zeroed based on a conversation on gitter with @haraldh and @npmccallum indicating there is a GPF otherwise.

`.Lsyscall`
Here there is already a `loadp` that gives us the previous un-64-byte-aligned trusted stack pointer back in `%rbx`. Also, it's important that `mov %rax, %rsp` occurs before `xrstor`, because the mask for latter will be using `%rax` and so the earlier value would be lost. Likewise, it is important that `mov %rdi, %rax ` occurs after `xrstor` and this is why that line seems to be arbitrarily moved several lines down. After `xrstor`, the initial trusted stack pointer value (from before the alignment for `xsavec`) is moved back to `%rsp`.

It is entirely possible that I have misunderstood something, but hopefully this explains the reasoning behind this current version of the code.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -154,6 +188,27 @@ enclave:
...
+    movq    $0,        552(%rsp)
+    movq    $0,        560(%rsp)
+    movq    $0,        568(%rsp)
+
+    # Set mask and perform xsavec
+    movq    $~0,                   %rdx
+    movq    $~0,                   %rax
+    xsavec (%rsp)
+
     savep                                           # Save preserved registers
```
haraldh, line 212: With `savep` you saved %rbx, but the original contents was destroyed earlier by `mov     %rsp,                   %rbx`.

So, I would suggest `savep` after `syscall:` immediately and `pushq %rbx` here instead.

You really should check, that all is restored in reverse order...

syscall:
1. savep
2. rsp -> rbx
3. xsave
4. push rbx aka rsp from 2.

syscall return:
4. pop rbx aka rsp for 2.
3. xrstor
2. rbx -> rsp
1. loadp

lkatalin:
Thanks @haraldh for the good advice.
I'm submitting #562 as a smaller PR for just clearing the extended state.

lkatalin:
Closed per the comments in #334 .

PR #429, opened 2020-04-20
cc: add a specific rlib from the target directory

haraldh:
adds "insert-target-rlib"

link.json:
…
        "insert-target-rlib": ["libc"],
…

will add

$HOME/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-musl/lib/liblibc-6c4492b949101b15.rlib

as a static library

### DIFF
--- a/cc
+++ b/cc
@@ -18,18 +18,21 @@
 #       "replace": { "-lba[rz]": [ "-Wl,--whole-archive", "-lbar", "-lbaz" ] },
 #       "prepend": [ "-lbar" ],
 #       "append": [ "-lbaz" ],
+#       "append-target-rlib": ["libc"],
 #       "debug": false
 #     },
 #     "build": {
 #       "replace": { "-lba[rz]": [ "-Wl,--whole-archive", "-lbar", "-lbaz" ] },
 #       "prepend": [ "-lbar" ],
 #       "append": [ "-lbaz" ]
+#       "append-target-rlib": ["libc"],
 #       "debug": false
 #     },
 #     "test": {
 #       "replace": { "-lba[rz]": [ "-Wl,--whole-archive", "-lbar", "-lbaz" ] },
 #       "prepend": [ "-lbar" ],
 #       "append": [ "-lbaz" ]
+#       "append-target-rlib": ["libc"],
 #       "debug": false
 #     }
 #  }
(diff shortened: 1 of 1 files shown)


haraldh:
This turns out to be very useful!

haraldh:
With link.json containing the libc, this saves us from opt level 3 in debug and implementing memcpy(), et.al.

haraldh:
Tested on coffeelake:
```console
$ ./enarx-keep-sgx/target/debug/enarx-keep-sgx --shim ./enarx-keep-sgx-shim/target/x86_64-unknown-linux-musl/debug/enarx-keep-sgx-shim --code payload/target/x86_64-unknown-linux-musl/debug/payload
arg: "/init"
env: "LANG=C"
aux: ExecFilename("/init")
aux: Platform("x86_64")
aux: Uid(1000)
aux: EUid(1000)
aux: Gid(1000)
aux: EGid(1000)
aux: PageSize(4096)
aux: Secure(false)
aux: ClockTick(100)
aux: Flags(0)
aux: HwCap(0)
aux: HwCap2(0)
aux: PHdr(35184376283200)
aux: PHent(56)
aux: PHnum(10)
aux: Random([96, 132, 88, 62, 151, 101, 22, 42, 31, 214, 123, 20, 208, 105, 180, 88])
```

cc
```diff
@@ -76,6 +77,14 @@ if cc is None:
 assert(cc is not None)
 argv = sys.argv[1:]
 
+
+target_rlib_dir = \
+    os.path.dirname(
+        list(
+            filter(lambda x: x.find("libcompiler_builtins") != -1 and x.endswith(".rlib"), argv)
+        )[0]
```
npmccallum, line 85: I suspect this could be cleaner as a list comprehension.

cc
```diff
@@ -98,6 +107,12 @@ try:
     # Prepend and append new items.
     argv = link.get("prepend", []) + argv + link.get("append", [])
 
+    for name in link.get("insert-target-rlib", []):
```
npmccallum, line 110:
```suggestion
    for name in link.get("insert-rlib", []):
```

Is this an improvement or not? I'm on the fence.

npmccallum:
```suggestion
    for name in link.get("append-rlib", []):
```

Using "append" matches the existing language.

haraldh:
> Do you also plan on submitting your patch for the sgx shim?

yes

cc
haraldh: hmm, I wanted to distinguish normal `rlib` from the `target-rlib`s ..

cc
```diff
@@ -98,6 +107,12 @@ try:
...
+        rlibs = glob.glob(target_rlib_dir + "/lib" + name + "-*.rlib")
+        for rlib in rlibs:
+            argv.append("-Wl,-Bstatic")
+            argv.append(rlib)
```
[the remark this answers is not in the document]
haraldh: right

PR #432, opened 2020-04-22
Introduce enarx-keep-sev

connorkuehl:
If this PR is too fine-grained let me know. I think this is as far as I can get without duplicating code from #416.

This just adds the crate as a basis for the rest of the code that is transitioning off of `demo2020`. The keep is pretty much a "stub" for now, but following PRs will add more functionality.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -3,6 +3,7 @@ CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = "true"
 CARGO_MAKE_WORKSPACE_EMULATION = true
 CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = [
 	"crt0stack",
+	"enarx-keep-sev",
 	"enarx-keep-sgx",
 	"enarx-keep-sgx-shim",
 	"enumerate",
--- a/enarx-keep-sev/Cargo.toml
+++ b/enarx-keep-sev/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "enarx-keep-sev"
+version = "0.1.0"
+authors = ["Connor Kuehl <ckuehl@redhat.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+structopt = "0.3"
--- a/enarx-keep-sev/LICENSE
+++ b/enarx-keep-sev/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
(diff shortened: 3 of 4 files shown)


enarx-keep-sev/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "enarx-keep-sev"
+version = "0.1.0"
+authors = ["Harald Hoyer <harald@redhat.com>"]
```
haraldh, line 4: you might as well add your name here, or replace it with some "The Enarx Development Team"

enarx-keep-sev/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+edition = "2018"
+license = "Apache-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+failure = "0.1.7"
```
haraldh, line 11: hmm..  I thought `failure` is dead? How about `anyhow`?

enarx-keep-sev/Cargo.toml
connorkuehl: I'd prefer to change it to something along the lines of team attribution, so I've opened an issue: https://github.com/enarx/enarx/issues/479.

enarx-keep-sev/Cargo.toml
connorkuehl: :scream: all my research led me to `failure`! Indeed, the repository doesn't look very active.

It's a shame I kind of like the crate overall though. Let me look and see what's out there.

haraldh: Yeah, `failure` was a try from withoutboats to tackle error handling, but I think he moved on to other things.

Meanwhile I created [chainerror](https://haraldh.github.io/chainerror/), which gives a nice error traceback (different from the panic backtrace). Also I added https://doc.rust-lang.org/std/error/trait.Error.html#method.chain but it's nightly only.

From my experience with error handling, if you are writing a `lib` crate, use something like https://github.com/tailhook/quick-error, and implement s.th. like `Chainerror` [manually](https://github.com/varlink/rust/blob/master/varlink/src/error.rs#L117-L140)

If you are creating a `bin` crate, you might use `anyhow`, but I prefer manual implementation.

npmccallum: See also this: https://www.youtube.com/watch?v=qWY2sUpt-Ek

I think we should have a discussion about error handling. My knowledge here is very poor. But I also know we need good error handling across a variety of contexts (std, no_std, no_std w/ nostdlib).

enarx-keep-sev/Cargo.toml
npmccallum: @connorkuehl Please add your name for now. FLF

enarx-keep-sev/Cargo.toml
npmccallum: In the meantime, failure is dead so lets remove it.

connorkuehl: > I think we should have a discussion about error handling. My knowledge here is very poor. But I also know we need good error handling across a variety of contexts (std, no_std, no_std w/ nostdlib).

I did some research and watched the video linked above. The latest push is my first stab at what I perceive to be "consensus" among the internet without depending on the error-handling crate du jour. However, it will not work in `no_std` due to Boxing the `Error` trait objects.

Makefile.toml
```diff
@@ -3,7 +3,7 @@ CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = "true"
 CARGO_MAKE_WORKSPACE_EMULATION = true
 CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = [
 	"crt0stack",
-	"enarx-keep-sev-lib",
```
haraldh, line 6: Is this intentional?

enarx-keep-sev/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+#[macro_export]
+macro_rules! context {
+    ($msg:expr, $err:expr) => {
+        KeepError::context($msg, std::boxed::Box::new($err))
+    };
+}
+
+/// The primary error type for the crate.
+#[derive(Debug)]
+pub struct KeepError {
```
haraldh, line 31: I would like to see `KeepError` impl `std::error::Error` with the `source()` method.

Makefile.toml
connorkuehl: No, thank you for catching this!

npmccallum:
@haraldh Your changes requested is blocking this from merge.

PR #433, opened 2020-04-23
Check for invalid opcode in event()

### LINKED ISSUE #336: test for exception type in sgx shim handler
Currently (#326), we presume the event that caused a subsequent (`CSSA == 1`) entry into the enclave was an invalid opcode. However, a malicious host could try to send other events or even fake events. We should test for the actual events that caused the exception from the `aex` structure, since these are trusted.

lkatalin:
Resolves #336

### DIFF
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use intel_types::Exception;
 use nolibc::x86_64::error::Number as ErrNo;
 use nolibc::x86_64::syscall::Number as SysCall;
 use sgx_types::ssa::StateSaveArea;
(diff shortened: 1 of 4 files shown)


sgx-types/src/ssa.rs
```diff
@@ -32,7 +32,8 @@ enumerate! {
 #[repr(C, align(4))]
 #[derive(Copy, Clone)]
 pub struct ExitInfo {
-    vector: Exception,
+    /// The exception type. See intel-types crate for more.
+    pub vector: Exception,
     exit_type: ExitType,
     reserved: u8,
     valid: u8,
```
connorkuehl, line 39: I think `#[repr(C)]` suggests this type will be exactly 4 bytes as described by the Intel SDM, right? I'm just thinking out loud (because of some of the SEV types there are) if it's worthwhile having a unit test assert that the size of these structures are what we expect them to be. Hmm. Or on the other hand, maybe such a test is redundant.

connorkuehl: Ah nevermind, I see that we're accomplishing this with `testaso` on the enclosing structures.

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -16,29 +17,38 @@ pub extern "C" fn event(
     aex: &mut StateSaveArea,
     ctx: &Context,
 ) {
-    match unsafe { core::slice::from_raw_parts(aex.gpr.rip as *const u8, 2) } {
-        // syscall
-        [0x0f, 0x05] => {
-            let syscall = aex.gpr.rax.into();
-            let mut h = Handler::new(tcs, aex, ctx);
+    // Test for exception type.
```
connorkuehl, line 20: I think something like

```
// Exception Vector Table
```

better describes what's happening here.

sgx-types/src/ssa.rs
npmccallum: Yup. This is not an issue.

lkatalin:
@npmccallum Ah, you're right about the `vector`. I'll fix it.

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -16,29 +17,38 @@ pub extern "C" fn event(
     aex: &mut StateSaveArea,
     ctx: &Context,
 ) {
-    match unsafe { core::slice::from_raw_parts(aex.gpr.rip as *const u8, 2) } {
-        // syscall
-        [0x0f, 0x05] => {
-            let syscall = aex.gpr.rax.into();
-            let mut h = Handler::new(tcs, aex, ctx);
+    // Exception Vector Table
+    match aex.gpr.exitinfo.exception().unwrap() {
```
npmccallum, line 21: We can't use `unwrap()` in the shim. It will cause an infinite loop if it doesn't succeed.

Also, please note that our `exception()` function before was theoretical. We aren't actually sure if it is correct. So we definitely need to test it.

npmccallum:
@lkatalin I see that you have added tests for the exceptions. This is good.

However, my concern was not that the `exception()` function works as we designed it, but that our design might mismatch the actual behavior of SGX. Perhaps we should look at the structs again given what we now know about SGX exceptions.

npmccallum:
@lkatalin Please rebase as there is now a conflict.

sgx-types/src/ssa.rs
```diff
@@ -341,6 +341,25 @@ impl StateSaveArea {
...
+    fn exc() {
+        let ssa = StateSaveArea::default();
+        assert_eq!(None, ssa.gpr.exitinfo.exception());
+        let mut ssa = StateSaveArea::default();
+        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Hardware, Exception::InvalidOpcode);
+        assert_eq!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));
+        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Software, Exception::InvalidOpcode);
+        assert_eq!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));
+        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Hardware, Exception::DivideByZero);
+        assert_ne!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));
```
npmccallum, line 359:
```suggestion
        let mut ssa = StateSaveArea::default();
        assert_eq!(None, ssa.gpr.exitinfo.exception());

        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Hardware, Exception::InvalidOpcode);
        assert_eq!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));

        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Software, Exception::InvalidOpcode);
        assert_eq!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));

        ssa.gpr.exitinfo = ExitInfo::new(ExitType::Hardware, Exception::DivideByZero);
        assert_ne!(ssa.gpr.exitinfo.exception(), Some(Exception::InvalidOpcode));
```

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -18,70 +19,81 @@ pub extern "C" fn event(
     aex: &mut StateSaveArea,
     ctx: &Context,
 ) {
-    let mut h = Handler::new(layout, aex, ctx);
```
npmccallum, line 21: You should leave this line here and not move the handler instantiation into the match structure. You create it in all cases anyway.

npmccallum:
@lkatalin I have tested it and it works.

enarx-keep-sgx-shim/src/event.rs
lkatalin: @npmccallum I was getting some anger from the Rust borrow checker, but I'll try again.

lkatalin: I got it.

Issue #348, opened 2020-03-18
Add docs to `span`


mbestavros:
The `span` crate currently fails the `#![deny(missing_docs)]` lint test. It's been grandfathered in to make the tests pass, but we should add docs to make the test pass properly.

PR #450, opened 2020-04-27
Add Line type to the span crate

npmccallum:
This type is similar to core::ops::Range, but is not an iterator and
therefore can implement Copy. Additionally, we make sure all exported
items are documented.

Fixes: #348

### DIFF
--- a/span/src/lib.rs
+++ b/span/src/lib.rs
@@ -1,52 +1,38 @@
 // SPDX-License-Identifier: Apache-2.0
 
+//! This crate contains types for measuring linear sets by either the end
+//! points (`Bounds`) or by a starting point and the number of elements (`Span`).
+//!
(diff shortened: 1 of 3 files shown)


span/src/lib.rs
```diff
@@ -1,52 +1,38 @@
 // SPDX-License-Identifier: Apache-2.0
 
+//! This crate contains types for measuring linear sets by either the end
+//! points (`Bounds`) or by a starting point and the number of elements (`Span`).
+//!
+//! In the interest of zero-cost abstractions, all method are always inlined
```
lkatalin, line 6:
```suggestion
//! In the interest of zero-cost abstractions, all methods are always inlined
```

npmccallum: Fixed.

PR #453, opened 2020-04-27
Rework how segments are loaded

npmccallum:
We can now handle segments that aren't page aligned.


enarx-keep-sgx/src/component.rs
```diff
@@ -1,28 +1,32 @@
...
-use std::fs::File;
-use std::io::{Read, Result};
+use std::io::Result;
 use std::ops::Range;
 use std::path::Path;
 
 /// A loadable segment of code
 pub struct Segment {
-    pub src: Contents,
+    src: Vec<Page>,
```
lkatalin, line 16: Pretty sure this was the part where `pub` is removed here, but was added back again in a later commit. Will we want this to be `pub` in the end?

npmccallum: Fixed.

PR #455, opened 2020-04-28
move Build struct to sev-types

### LINKED ISSUE #446: Move Build struct from 'sev' into 'sev-types'
Right now it lives in [`sev/src/lib.rs`](https://github.com/enarx/enarx/blob/master/sev/src/lib.rs#L19) but it belongs in the `sev-types` crate.

Move it to the `sev-types` crate.

michiboo:
Move Build struct from 'sev' into 'sev-types' 

Closes #446

### DIFF
--- a/sev-types/src/lib.rs
+++ b/sev-types/src/lib.rs
@@ -16,22 +16,9 @@
 //! the SEV API specification document version; the specification publication
 //! date; the link to the SEV API specification.
 
+#![no_std]
 #![deny(missing_docs)]
 #![deny(clippy::all)]
-#![allow(missing_docs)]
 
 pub mod command;
 pub mod platform;
-
-#[repr(C)]
-#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Build {
-    pub version: platform::Version,
-    pub build: u8,
-}
-
-impl std::fmt::Display for Build {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}.{}", self.version, self.build)
-    }
-}
--- a/sev-types/src/lib.rs
+++ b/sev-types/src/lib.rs
@@ -16,9 +16,22 @@
 //! the SEV API specification document version; the specification publication
 //! date; the link to the SEV API specification.
 
-#![no_std]
 #![deny(missing_docs)]
 #![deny(clippy::all)]
+#![allow(missing_docs)]
 
 pub mod command;
 pub mod platform;
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Build {
+    pub version: platform::Version,
+    pub build: u8,
+}
+
+impl std::fmt::Display for Build {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}.{}", self.version, self.build)
+    }
+}
--- a/sev-types/src/platform.rs
+++ b/sev-types/src/platform.rs
@@ -21,3 +21,20 @@ impl core::fmt::Display for Version {
         write!(f, "{}.{}", self.major, self.minor)
     }
 }
+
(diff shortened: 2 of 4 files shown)


sev-types/src/lib.rs
```diff
@@ -16,9 +16,22 @@
 //! the SEV API specification document version; the specification publication
 //! date; the link to the SEV API specification.
 
-#![no_std]
 #![deny(missing_docs)]
 #![deny(clippy::all)]
+#![allow(missing_docs)]
```
connorkuehl, line 21: Let's remove this `allow` attribute, because we want to make sure the entire crate is documented. Putting this `allow` statement here would turn off the automatic check for associated doc comments. If you're unsure of what a structure represents, the AMD SEV specification is linked in the comment at the top of `lib.rs`

sev-types/src/lib.rs
```diff
@@ -16,9 +16,22 @@
 //! the SEV API specification document version; the specification publication
 //! date; the link to the SEV API specification.
 
-#![no_std]
```
connorkuehl, line 19: We don't want to remove this. We want `sev-types` to be `no_std`.

sev-types/src/lib.rs
```diff
@@ -16,9 +16,22 @@
...
+pub struct Build {
+    pub version: platform::Version,
+    pub build: u8,
+}
+
+impl std::fmt::Display for Build {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}.{}", self.version, self.build)
+    }
+}
```
connorkuehl, line 37: Replace the `std` module with `core` so that the crate compiles in `no_std`.

sev-types/src/lib.rs
```diff
@@ -16,9 +16,22 @@
...
 #![deny(missing_docs)]
 #![deny(clippy::all)]
+#![allow(missing_docs)]
 
 pub mod command;
 pub mod platform;
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Build {
```
connorkuehl, line 28: Since the struct represents the platform firmware build, I'd argue that it belongs in the `platform` module.

Furthermore, this struct will require a doc comment describing it.

sev-types/src/platform.rs
```diff
@@ -21,3 +21,20 @@ impl core::fmt::Display for Version {
         write!(f, "{}.{}", self.major, self.minor)
     }
 }
+
+/// The fireware 'Build' is the Build ID for a particular API version.
```
haraldh, line 25: s/fireware/firmware/ ? typo?

sev-types/src/platform.rs
```diff
@@ -21,3 +21,20 @@ impl core::fmt::Display for Version {
...
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Build {
+    /// The fireware's version.
```
haraldh, line 29: s/fireware/firmware/ ? typo?

sev/tests/api.rs
```diff
@@ -2,7 +2,9 @@
 
 use sev_types::platform::Version;
 
-use sev::{certs::sev::Usage, firmware::Firmware, Build};
+use sev::{certs::sev::Usage, firmware::Firmware};
+
+use sev_types::platform::Build;
```
connorkuehl, line 7: Let's combine the `use sev_types` statements into one:

```rust
use sev_types::platform::{Build, Version};
```

sev/src/lib.rs
```diff
@@ -9,22 +9,10 @@
 #![allow(missing_docs)]
 
 use sev_types::platform;
+use sev_types::platform::Build;
```
connorkuehl, line 12: I'm fine with this either way but I think my only nitpick is that I'd prefer to remove line 12 here and just refer to the build as `platform::Build` as I think the namespacing provided by the module adds some more precision to the readability.

npmccallum:
Sorry, I hit the wrong button.

PR #456, opened 2020-04-28
Move enclave creation logic into a separate function

npmccallum:


npmccallum:
This PR should be mostly whitespace changes. We're just moving a big chunk of code into a separate function.

enarx-keep-sgx/src/main.rs
```diff
@@ -16,65 +16,67 @@ use span::Span;
...
+        let mut src = unsafe { seg.src.align_to().1 };
+        let mut off = seg.dst.start - span.start;
+
+        // The first page of the shim entry is the TCS page.
+        if seg.dst.start == shim.entry {
+            builder
+                .load(&src[..4096], off, SecInfo::tcs())
+                .expect("Unable to add TCS page");
+            src = &src[4096..];
+            off += 4096;
```
connorkuehl, line 52: Should we pull in `memory` for `Page::size()` here to replace the `4096` literal?

npmccallum: This happens already in future patches. This patch just moves existing code into a function.

npmccallum: My point is that this is a good idea and I've already implemented it, but it is out of scope for this commit.

connorkuehl: Ah, I see. No worries then :slightly_smiling_face:

Issue #388, opened 2020-03-30
Measure the SGX entry point parameter


npmccallum:
We currently take the entry point to the payload as an unmeasured parameter. We need a way to measure it. We can't let the host choose an arbitrary entry point.

PR #462, opened 2020-04-28
Rework the SGX loader to pass measured argument

npmccallum:
We pass a full Layout struct describing the layout of the memory space.

Fixes: #388

### DIFF
--- a/enarx-keep-sgx/src/layout.rs
+++ b/enarx-keep-sgx/src/layout.rs
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use super::component::Component;
+
+use span::{Contains as _, Line, Span};
+
+const PREFIX: usize = units::bytes![4; MiB];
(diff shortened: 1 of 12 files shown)


enarx-keep-sgx/src/layout.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+fn below(rel: impl Into<Line<usize>>, size: usize) -> Span<usize> {
+    Span {
+        start: lower(rel.into().start - size, ALIGN),
+        count: size,
+    }
+}
+
+// NOTE: this structure MUST be kept in sync with enarx-keep-sgx-shim
```
lkatalin, line 35: Should this be a debt issue? I can file it.

npmccallum: Yes, please do.

lkatalin: https://github.com/enarx/enarx/issues/463

PR #468, opened 2020-04-28
Add printing and debug facilities to the shim

npmccallum:


npmccallum:
A related debt issue: https://github.com/enarx/enarx/issues/469

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -10,6 +10,8 @@ use span::{Contains, Line, Span};
 
 use core::{mem::size_of, slice::from_raw_parts_mut};
 
+const TRACE: bool = false;
```
connorkuehl, line 13: 1. Can we replace this with a compiler attribute like `#[cfg(trace)]`
2. If so, could we remove the `if !TRACE` if-statement from the `trace` function and instead define a `trace!()` macro that checks if the attribute is set and if so, invokes the `trace` fn? That way we aren't enforcing a policy inside the `trace` fn but the macro still takes up the heavy lifting of checking whether we should trace or not.

npmccallum: As I mentioned previously, I think the right approach for this is to conditionalize on the release profile of the build.

Regardless, I'd like to come up with a common experience plan with SEV. But at the moment we're just trying to minimize the number of PRs.

So I'm inclined to treat this as a `debt` issue for both SGX and SEV. Thoughts?

connorkuehl: > So I'm inclined to treat this as a `debt` issue for both SGX and SEV. Thoughts?

Works for me!

npmccallum: https://github.com/enarx/enarx/issues/472

Issue #82, opened 2020-01-03
Run a musl static-pie binary in SGX


npmccallum:
This depends on #81.

PR #473, opened 2020-04-28
Fix brk(); implement remaining syscalls for payload

npmccallum:
Fixes: #406
Fixes: #405
Fixes: #82


npmccallum:
Tests have actually succeeded. If you notice, `Mergeable` had a hiccup.

Issue #483, opened 2020-05-01
Miscellaneous errors with automated labeling bot


mbestavros:
Seems like two independent instances of `KeyError`; hopefully a quick fix.

PR #484, opened 2020-05-01
Better error handling for project automation bot

mbestavros:
There were two small `KeyError` bugs with the project automation bot raised in #483:

- The bot didn't know what to do if an issue was labeled with a label it didn't recognize
- The bot was grabbing the wrong project due to a typo

Looking at those, I also noticed a third one:

- If a user manually added an issue to both the project and its corresponding label at the same time, the bot would fail due to Github's API returning that the card was already in the project.

 This PR resolves all the above bugs by improving the bot's error handling.

Resolves: #483

### DIFF
--- a/.automation/add-to-project-when-labeled
+++ b/.automation/add-to-project-when-labeled
@@ -3,7 +3,7 @@
 
 import json
 import sys
-from github import Github
+import github
 
 # A mapping between labels and projects to add when they're applied.
 LABEL_PROJECTS = {
@@ -17,7 +17,7 @@ LABEL_PROJECTS = {
 (token, repository, path) = sys.argv[1:4]
 
 # Initialize repo
-repo = Github(token).get_repo(repository)
+repo = github.Github(token).get_repo(repository)
 
 # Open Github event JSON
 with open(path) as f:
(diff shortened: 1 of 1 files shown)


.automation/add-to-project-when-labeled
```diff
@@ -38,15 +38,26 @@ except KeyError:
...
 
 # If the label was added, we want to add to the project. If removed, we want to
 # remove from the project.
 if action == "labeled":
     project_column = project.get_columns()[1]
-    project_column.create_card(content_id=content_id, content_type=content_type)
+    try:
+        project_column.create_card(content_id=content_id, content_type=content_type)
+    except github.GithubException as e:
+        if e.data["errors"][0]["message"] == "Project already has the associated issue":
```
mbestavros, line 54: I don't know if the message that Github returns changes for pull requests (and we won't encounter it in any case at the moment since we can't run on PRs). Do we maybe want to change this to check if the error message starts with `"Project already has the associated"`?

npmccallum: Is there an error number we can check instead?

npmccallum: Matching on strings can be problematic; particularly when translations get involved.

mbestavros: We can match on the the error code (`422`). I'll make that change. Again, no idea if that changes for PRs.

.automation/add-to-project-when-labeled
```diff
@@ -38,15 +38,26 @@ except KeyError:
...
 
 # If the label was added, we want to add to the project. If removed, we want to
 # remove from the project.
 if action == "labeled":
     project_column = project.get_columns()[1]
-    project_column.create_card(content_id=content_id, content_type=content_type)
+    try:
+        project_column.create_card(content_id=content_id, content_type=content_type)
+    except github.GithubException as e:
+        if e.status == 422:
```
npmccallum, line 54: Don't use a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)).

mbestavros: Noted. Latest push fixes this.

.automation/add-to-project-when-labeled
```diff
@@ -38,15 +38,27 @@ except KeyError:
...
     project_column = project.get_columns()[1]
-    project_column.create_card(content_id=content_id, content_type=content_type)
+    try:
+        project_column.create_card(content_id=content_id, content_type=content_type)
+    except github.GithubException as e:
+        error = e.data["errors"][0]
+        if error["resource"] == "ProjectCard" and error["code"] == "unprocessable":
+            print("Card already in project.")
+        else:
+            raise
```
npmccallum, line 58:
```suggestion
        if error["resource"] != "ProjectCard" or error["code"] != "unprocessable":
            raise
        print("Card already in project.")
```

mbestavros: I like that better. Changed in latest push.

Issue #328, opened 2020-03-16
Instruction pattern match constants


npmccallum:
This is a debt bug after #326.

In `enarx-keep-sgx-shim`, we currently do something like this:

```rust
match core::slice::from_raw_parts(aex.gpr.rip as *const u8, 2) {
    // syscall
    [0x0f, 0x05] => { /* code here */ },

    // unknown invalid opcode
    _ => { /* code here */ }
}
```

The above code looks at the raw bytes for the instruction that produced the invalid opcode exception and then attempts to handle it. I don't know if this is possible, but it would be great to have constants for those match patterns.

lkatalin:
Looks like this was addressed by #490 , should we close? @npmccallum

npmccallum:
Yes.

Issue #442, opened 2020-04-24
Merge intel-types into x86_64


npmccallum:
@haraldh brought this up. Most of these types probably belong in x86_64 so we don't have to maintain them.

npmccallum:
@lkatalin FYI

lkatalin:
@npmccallum We don't have the `intel-types` crate anymore. Is this done?

npmccallum:
Yes. Most of `intel-types` is here https://github.com/enarx/xsave

PR #490, opened 2020-05-03
Instruction Pattern Match Constants

blazebissar:
Addresses Issue #328 

In this commit I define a handful of opcodes as defined in the [Intel® 64 and IA-32 Architectures Software Developer’s Manual](https://software.intel.com/sites/default/files/managed/39/c5/325462-sdm-vol-1-2abcd-3abcd.pdf) in order to enable nicer pattern matching. I'm a new contributor so I am eager for feedback, especially about the following:

1. Opcodes are varying-length, ranging from one to three bytes with additional modifiers that can be added. I found this an interesting design consideration because it seems to me that an enum (using the enumerate macro) would be a good way to incorporate so many opcode constants - but the only way I could figure out to do this was to use an enum of `&[u8]` values, which would have required specifying a lifetime. This seemed inefficient for a list of constants that could grow very large over time, and the best idea I had was to use a number of `pub const` values instead. Is there a better way to approach this problem?

2. Additionally, there are multiple syscall opcodes I was able to identify (specifically `OP_SYSCALL` and `OP_SYSENTER` and their corresponding exit operations. I cannot tell if we need to account for both in the current `entry.rs`, or if we have reason to believe that Enarx will only have to handle `OP_SYSCALL`s. I can add in additional matches if needed. In this vein, do we need additional consideration for the four `OP_SYSEXIT` and `OP_SYSRET` opcodes at this time? My assumption is no, because the handler seems to be pushing calls from the keep into the more privileged rings, rather than handling their results - but I would love to properly understand this piece.

3. If we will need to support different length opcodes, I could try to make the matching logic in `event.rs` more sophisticated so that it does not assume two byes for each opcode, this is a more generic observation compared to point two above.

4. Finally, what other opcodes will we need constants for? I imagine not every one, but if we know certain classes of instructions now then I am happy to also add them in.

I look forward to your feedback.

### DIFF
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -8,6 +8,11 @@ use sgx_types::ssa::StateSaveArea;
 use crate::handler::{Context, Handler, Print};
 use crate::Layout;
 
+// Opcode constants, details in Volume 2 of the Intel 64 and IA-32 Architectures Software
+// Developer's Manual
+const OP_SYSCALL: &[u8] = &[0x0f, 0x05];
+const OP_CPUID: &[u8] = &[0x0f, 0xa2];
+
 #[no_mangle]
 pub extern "C" fn event(
     _rdi: u64,
@@ -25,8 +30,7 @@ pub extern "C" fn event(
     match h.aex.gpr.exitinfo.exception() {
         Some(Exception::InvalidOpcode) => {
             match unsafe { core::slice::from_raw_parts(h.aex.gpr.rip as *const u8, 2) } {
-                // syscall
-                [0x0f, 0x05] => {
+                OP_SYSCALL => {
                     aex.gpr.rax = match h.aex.gpr.rax.into() {
                         SysCall::READ => h.read(),
                         SysCall::READV => h.readv(),
@@ -57,8 +61,7 @@ pub extern "C" fn event(
                     aex.gpr.rip += 2;
                 }
 
-                // cpuid
-                [0x0f, 0xa2] => {
+                OP_CPUID => {
                     let (rax, rbx, rcx, rdx) = match (h.aex.gpr.rax, h.aex.gpr.rcx) {
                         (0, _) => (0, 0x756e_6547, 0x6c65_746e, 0x4965_6e69), // "GenuineIntel"
 


npmccallum:
@blazebissar Thank you so much for your PR!

1. We are moving away from the `nolibc` crate. We intend to use the `libc` crate which has a `no_std` mode. However, doing this was blocked on #420, which is now merged. The `nolibc` crate will soon be removed.

2. Matching based on array constants will be problematic if we have to handle opcodes of differing lengths. This is why we had chosen to use [slice patterns](https://doc.rust-lang.org/edition-guide/rust-2018/slice-patterns.html) for matching. That being said, today we only match 2 byte opcodes. It might be more clear to cast `rip` to a `*const u16`.

@connorkuehl @haraldh @whitebrandy What are your thoughts here?

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -24,9 +27,13 @@ pub extern "C" fn event(
     // Exception Vector Table
     match h.aex.gpr.exitinfo.exception() {
         Some(Exception::InvalidOpcode) => {
-            match unsafe { core::slice::from_raw_parts(h.aex.gpr.rip as *const u8, 2) } {
+            match unsafe {
+                core::slice::from_raw_parts(h.aex.gpr.rip as *const u8, 2)
+                    .try_into()
+                    .unwrap()
```
lkatalin, line 33: Small comment here that unwrapping in the `sgx-shim` can cause an [infinite loop](https://github.com/enarx/enarx/blob/master/enarx-keep-sgx-shim/src/main.rs#L13), so we prefer not to unwrap.

blazebissar: Thank you both, very good to know!

I'm thinking it would be better to use slices for the opcode constants, like:
`pub const OP_SYSCALL: &[u8] = &[0x0f, 0x05];` 
so we don't risk that loop and also maintain the original flexibility of slice pattern matching.

As for a good crate for the opcode constants, my first instinct is to use `intel-types`, but I don't know if that is necessarily appropriate because of issue #442 . Would it be better to define them local to the `enarx-keep-sgx-shim` crate, similar to the ELF `Header` struct and `MAGIC` const?

blazebissar:
Committed some new code that I think will address the above issues.

Constants are now defined as slices to maintain the same flexibility. They are also now defined within the `enarx-keep-sgx-shim` crate. I also removed the unused opcodes to avoid dead code warnings and to reduce unnecessary clutter.

Eager to hear additional feedback.

npmccallum:
@blazebissar Please squash the two commits. We require logical commits rather than "development timeline" commits.

blazebissar:
Squashed; sorry for forgetting that.

npmccallum:
@blazebissar Let's just put the constants in the file directly. No need for a separate module.

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -26,7 +31,7 @@ pub extern "C" fn event(
         Some(Exception::InvalidOpcode) => {
             match unsafe { core::slice::from_raw_parts(h.aex.gpr.rip as *const u8, 2) } {
                 // syscall
-                [0x0f, 0x05] => {
+                OP_SYSCALL => {
```
connorkuehl, line 34: I think the constants provided by your patch make the comment above redundant. :slightly_smiling_face: 

:+1: for readability

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -58,7 +63,7 @@ pub extern "C" fn event(
                 }
 
                 // cpuid
-                [0x0f, 0xa2] => {
+                OP_CPUID => {
```
connorkuehl, line 66: Same comment as above

connorkuehl:
> 2. Matching based on array constants will be problematic if we have to handle opcodes of differing lengths. This is why we had chosen to use [slice patterns](https://doc.rust-lang.org/edition-guide/rust-2018/slice-patterns.html) for matching. That being said, today we only match 2 byte opcodes. It might be more clear to cast `rip` to a `*const u16`.

If we do end up decoding opcodes > 2 bytes, will we need to update match arms to pad from the left?

I.e.:

```rust
[0x0f, 0x05] => {
```

would need to become:

```rust
[0x0, 0x0f, 0x05] => {
```

I don't think the compiler would complain if we forgot in that case. I think the margin for human error when we go to update the table scales with the size of the table.

If we do believe we're going to be unpacking opcodes larger than 2 bytes then I think we should switch to a `*const u16` as you suggested (if we have to left-pad our existing slices, that is).

npmccallum:
> > 1. Matching based on array constants will be problematic if we have to handle opcodes of differing lengths. This is why we had chosen to use [slice patterns](https://doc.rust-lang.org/edition-guide/rust-2018/slice-patterns.html) for matching. That being said, today we only match 2 byte opcodes. It might be more clear to cast `rip` to a `*const u16`.
> 
> If we do end up decoding opcodes > 2 bytes, will we need to update match arms to pad from the left?
> 
> I.e.:
> 
> ```rust
> [0x0f, 0x05] => {
> ```
> 
> would need to become:
> 
> ```rust
> [0x0, 0x0f, 0x05] => {
> ```

It would need to become:

```rust
[0x0f, 0x05, _] => {
```

Unfortunately, slice patterns can't be constants, which is why we currently don't have constants.

However, I don't know of any three (or one) byte instructions we need to trap. So I'm in favor of improving the readability now and dealing with that potential problem later. Maybe Rust will grow slice pattern constants in the future...

npmccallum:
@connorkuehl I think you need to request changes so that @blazebissar knows to remove the comments and resubmit.

enarx-keep-sgx-shim/src/event.rs
blazebissar: Comment removed

enarx-keep-sgx-shim/src/event.rs
blazebissar: Comment removed

blazebissar:
Redundant `// syscall` and `// cpuid` comments removed. Very useful style input I'll keep in mind.

npmccallum:
@blazebissar Congratulations on your first commit!

PR #498, opened 2020-05-05
An alternate syscall interface proposal

npmccallum:
This still needs work. But it is based substantially on #494.

I was inspired by #494 to look through all the Linux syscall ABIs. This interface should work for all of them. We no longer needs getters/setters with this approach since the arguments are always stored by their position. Conversion to actual registers is delegated to the architecture specific code, which seems correct.

The only thing that is sort of weird here is the `Reply` struct. This is because `ppc64` doesn't follow the convention of the other architectures. Instead of reserving a space in the return register for error numbers, the error flag is parsed in `cr0.SO`. The data structure of `Reply` is designed to work for all these cases.

### DIFF
--- a/enarx-keep-sev-lib/src/proxy.rs
+++ b/enarx-keep-sev-lib/src/proxy.rs
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `proxy` module contains structures used to facilitate communication between
+//! the microkernel and the hypervisor. This is referred to as "proxying" in the
+//! project literature. This is a very thin and low-level layer that is meant to
+//! be as transparent as possible.
+
+use memory::Register;
+
+/// The `Message` struct is the most minimal representation of the register context
+/// needed for service requests between the microkernel and the hypervisor. An example
+/// of such a service would be proxying a system call.
+///
+/// This struct contains a number of convenient setter/getter functions to avoid accidentally
+/// mixing up registers when implementing requests.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct Message {
+    /// The `rax` register is where the microkernel will indicate what kind of request
+    /// (or syscall number) the hypervisor should carry out on its behalf. The hypervisor
+    /// will write the return code to this register once it has completed its work.
+    ///
+    /// Note: this register context corresponds to the x86_64 `syscall` instruction style
+    /// calling convention.
+    pub rax: Register<u64>,
+    /// 1st parameter register.
+    pub rdi: Register<u64>,
+    /// 2nd parameter register.
+    pub rsi: Register<u64>,
+    /// 3rd parameter register.
+    pub rdx: Register<u64>,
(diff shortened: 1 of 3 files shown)


haraldh:
Much better :-)

enarx-keep-sev-lib/src/proxy.rs
```diff
@@ -5,120 +5,81 @@
...
+/// with error numbers above `usize::max_value() - 4096`, `ppc64` uses
+/// the `cr0.SO` flag to indicate error instead. Unfortunately, we also
+/// can't use the built-in `Result` type for this, since its memory layout
+/// is undefined. Therefore, we use this layout with conversions for `Result`.
+#[repr(C)]
+#[derive(Copy, Clone, Default)]
+pub struct Reply {
+    val: Register<usize>,
+    err: bool,
+}
```
connorkuehl, line 46: If we're establishing a proxy interface for more than just SEV then we should probably revert e527737a0eef01fb8d6d5b8cb37edce7700575e0.

enarx-keep-sev-lib/src/proxy.rs
```diff
@@ -5,120 +5,81 @@
...
 
-    /// Assign a value to the sixth parameter register (r9).
-    pub fn sixth(mut self, val: Register<u64>) -> Self {
-        self.r9 = val;
-        self
-    }
+/// A message, which is either a request or a reply
+#[derive(Copy, Clone)]
+pub union Message {
```
connorkuehl, line 74: Should this be `#[repr(C)]`? I'm guessing not because both of its variants are `#[repr(C)]`, right?

connorkuehl:
Can I do this? Closes #494

enarx-keep-sev-lib/src/proxy.rs
npmccallum: This probably needs to be fixed. Otherwise the layout is undefined.

enarx-keep-sev-lib/src/proxy.rs
npmccallum: We might even use this in SGX. So perhaps we should name it something completely different than `vmsyscall`.

npmccallum:
@connorkuehl I need to move on to something else. Can you fold this into your commit (which has a great commit message) and rename it properly. Keep yourself as the author and just [add me as a co-author](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors).

enarx-keep-sev-lib/src/proxy.rs
connorkuehl: How about `sally-port`? [Sally Port](https://en.wikipedia.org/wiki/Sally_port) Or is that too much :sweat_smile:

npmccallum: @connorkuehl `sallyport` it is! At some point we're going to jump the shark. Today is not that day. :)

Issue #508, opened 2020-05-06
Wire up Rome certificates in 'sevctl'


connorkuehl:
I'm moving a pull request from enarx/sevctl to enarx/enarx that installs the Rome generation certificates into the `sevctl` project directory.

Unless the user specifies which certs to use, `sevctl` will fall back on using the Naples certs for commands like `verify`.

Make `sevctl` Rome-cert aware.

connorkuehl:
Blocked on #520

PR #509, opened 2020-05-06
Add Rome ark.cert and ask.cert

connorkuehl:
@lilienbm kindly posted this patch on the enarx/sevctl repo. Since we've moved sevctl to this repo, I'll effectively "move" the PR by posting it here.

I'm going to post follow-up patches to make sevctl aware of these certs.

~~Related: #508~~
Closes: #508


sevctl/src/platform.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Debug)]
+pub struct ModelNumberParsingFail(pub String);
+
+impl TryFrom<&str> for Generation {
+    type Error = ModelNumberParsingFail;
+
+    fn try_from(maybe_model: &str) -> Result<Self, Self::Error> {
+        let naples = regex::Regex::new(r"AMD EPYC 7..1[P]?").expect("Failed to compile regex");
+        let rome = regex::Regex::new(r"AMD EPYC 7..2[P]?").expect("Failed to compile regex");
+
```
connorkuehl, line 37: @npmccallum @haraldh - are you aware of any specific cpuid or combination of cpuid values that can be used to deduce the processor generation? I've been poking through the AMD developer manuals and haven't found anything conclusive, but then again, they aren't terribly loquacious in describing some of these fields.

This is my "solution" I hacked together based on how AMD names their product lines.

lilienbm:
You can use the family and model by using the following kernel level code

void GetFamilyModel(uint32_t *family, uint32_t *model)
{
    uint32_t family_model = gSEVDevice.GetFamilyModel();
    *family = ((family_model>>20)&0xFF) + ((family_model>>8)&0x0F);
     // ExtFamily + BaseFamily
    *model = (((family_model>>16)&0x0F)<<4) + ((family_model>>4)&0x0F);
 // {ExtModel,BaseModel}
}

    case PSP_GET_FAMILY_MODEL:
        data = cpuid_eax(0x00000001);
        #ifdef DEBUG
            printk(KERN_DEBUG
"pspsev_ioctl: BaseModel %02x, BaseFamily %02x, ExtModel %02x, ExtFamily %02x\n"
, \
                                     (data>>4)&0x0F,  (data>>8)&0x0F, \
                                     (data>>16)&0x0F, (data>>20)&0xFF);
        #endif
        ret = copy_to_user((uint64_t *)arg, &data, sizeof(uint64_t));
        break;



constexpr uint32_t NAPLES_FAMILY     = 0x17UL;      // 23
constexpr uint32_t NAPLES_MODEL_LOW  = 0x00UL;
constexpr uint32_t NAPLES_MODEL_HIGH = 0x0FUL;
constexpr uint32_t ROME_FAMILY       = 0x17UL;      // 23
constexpr uint32_t ROME_MODEL_LOW    = 0x30UL;
constexpr uint32_t ROME_MODEL_HIGH   = 0x3FUL;

connorkuehl:
Thanks @lilienbm! I initally turned away from using the family/model values because my friend's Ryzen box reported the same values for 'cpu_family' (23) and 'model' (1) as my Naples box did and I was worried it wasn't enough unique information to be absolutely sure we were on an EPYC Naples/Rome machine. I guess my main worry was that it would be somehow misleading in case someone with a Ryzen machine started running some of this code.

lilienbm:
You can another check to make sure the platform supports SEV and SEV-ES

    case PSP_CHECK_SEV_ENABLED:
        enable_sev = cpuid_eax(SECURE_ENC_EAX);
        if
 (((enable_sev & SECURE_ENC_EAX_SME_BIT) == SECURE_ENC_EAX_SME_BIT) &&
           ((enable_sev & SECURE_ENC_EAX_SEV_BIT) == SECURE_ENC_EAX_SEV_BIT)) {
            data = 1;
        }
        ret = copy_to_user((uint64_t *)arg, &data, sizeof(uint64_t));
        break;

    case PSP_CHECK_SEVES_ENABLED:
        enable_sev = cpuid_eax(SECURE_ENC_EAX);
        if
 (((enable_sev & SECURE_ENC_EAX_VMPGFLUSH_BIT) ==
SECURE_ENC_EAX_VMPGFLUSH_BIT) &&

((enable_sev & SECURE_ENC_EAX_SEVES_BIT) == SECURE_ENC_EAX_SEVES_BIT)) {
            data = 1;
        }
        ret = copy_to_user((uint64_t *)arg, &data, sizeof(uint64_t));
        break;

where
// CPUID_Fn8000001F_EAX (Core::X86::Cpuid::SecureEncryptionEax)
#define SECURE_ENC_EAX                  0x8000001FULL
#define SECURE_ENC_EAX_SME_BIT          (1ULL << 0)
#define SECURE_ENC_EAX_SEV_BIT          (1ULL << 1)
#define SECURE_ENC_EAX_VMPGFLUSH_BIT    (1ULL << 2)
#define SECURE_ENC_EAX_SEVES_BIT        (1ULL << 3)

lilienbm:
*You can add another check to make sure the platform supports SEV and
SEV-ES since only EPYC parts support SEV

sevctl/src/platform.rs
haraldh: @connorkuehl nope, haven't tried yet

connorkuehl:
> *You can add another check to make sure the platform supports SEV and SEV-ES since only EPYC parts support SEV

[I've been down this road before](https://github.com/enarx/enarx/pull/216#issuecomment-583202661) :-) -- Ryzen processors still advertise support in `cpuid`, so I think even with all of these checks we still don't have sufficient evidence that we're running on an EPYC naples/rome processor.

However, I'm wondering if my trepidation is uncalled for, since technically checking the family and model range should be enough information to identify the processor generation for EPYC processors which is really what I want here. Furthermore, the flow of control would never reach this code if it was a processor that didn't have the correct microcode support for SEV because it would fail one of the `ioctl` calls before reaching here.

lilienbm:
If Ryzen parts advertise support for SEV, then that sounds like a bug to
me. Can you take a screenshot and I can send it to the right people?

connorkuehl:
> If Ryzen parts advertise support for SEV, then that sounds like a bug to me. Can you take a screenshot and I can send it to the right people?

@lilienbm That'd be great, thank you!

First, please see this comment: https://github.com/enarx/enarx/pull/216#issuecomment-583202661

Which shows output from our `sev-show` utility which is basically a series of calls to `cpuid`. The author of that comment is running a Ryzen 1700 CPU.

The source code of that tool is here: https://github.com/enarx/enarx/blob/master/sev-show/src/main.rs

I think you will find the more interesting bits to check will be to see the cpuid call to check the supported encrypted memory ops here: https://github.com/enarx/enarx/blob/master/sev-show/src/main.rs#L46

As well as some of the parts where we check for SEV support: https://github.com/enarx/enarx/blob/master/sev-show/src/main.rs#L123

and SME support: https://github.com/enarx/enarx/blob/master/sev-show/src/main.rs#L95

both of which are being reported as supported on the comment showing the output on a Ryzen 1700.

lilienbm:
Can you send over the family/model you're expecting? I'm looking at the
table and every part does have unique numbers.

lilienbm:
Also, can you try this
https://elixir.bootlin.com/linux/latest/source/drivers/crypto/ccp/psp-dev.c#L79


But it does look like the cpuid would return 1 for SEV support; we're
looking to clean that up in the future.

connorkuehl:
> Can you send over the family/model you're expecting? I'm looking at the table and every part does have unique numbers.

I don't know off the top of my head. I was just expecting them to be different values than are reported on a Ryzen chip.

> Also, can you try this https://elixir.bootlin.com/linux/latest/source/drivers/crypto/ccp/psp-dev.c#L79

I skimmed through the code and I'm not sure that this is reachable from userspace. I'll come back to this on Monday. I'd be interested in sending a patch to expose this function to userspace if it does indeed conclusively state whether SEV is supported in microcode or not.

PR #510, opened 2020-05-07
sev/src/firmware/linux.rs: open("/dev/sev") RW

haraldh:
otherwise pek_csr fails
```
[pid 101835] openat(AT_FDCWD, "/dev/sev", O_RDONLY|O_CLOEXEC) = 3
[pid 101835] ioctl(3, SEV_ISSUE_CMD, 0x7f7d0c8f5fc0) = -1 EPERM (Operation not permitted)
```

### DIFF
--- a/sev/src/firmware/linux.rs
+++ b/sev/src/firmware/linux.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use std::fs::File;
+use std::fs::{File, OpenOptions};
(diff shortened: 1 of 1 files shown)


npmccallum:
@haraldh Can you file a debt issue to split up this struct into two structs, one that does RO operations and the other which does RW operations?

haraldh:
> @haraldh Can you file a debt issue to split up this struct into two structs, one that does RO operations and the other which does RW operations?

done in #511

PR #513, opened 2020-05-07
Calculate xsave buffer size at enclave load time

### LINKED ISSUE #501: Pre-populate SGX Layout struct with sanitized CPUID results
The loader should pre-compute the size of the `xsave` area given the attributes in the signature and stick the resulting size in the Layout struct. This will be part of the enclave measurement and will inform the size of the `xsave` buffer.

lkatalin:
Resolves #501


lkatalin:
For future reference, the register sizes are based on https://github.com/intel/linux-sgx/blob/master/sdk/sign_tool/SignTool/manage_metadata.cpp#L1043-L1046.

lkatalin:
Closed in favor of #538 and #539 (reasoning described in more detail in associated issue: #501 ).

PR #521, opened 2020-05-12
Add enarx-keep-sev Error

haraldh:


connorkuehl:
We have decided to close this since all of our error paths are described as `io::Error` (at least so far).

PR #528, opened 2020-05-13
Run SGX payload during cargo make ci-flow

lkatalin:
This runs `enarx-keep-sgx/target/debug/enarx-keep-sgx --code payload/target/x86_64-unknown-linux-musl/debug/payload --shim enarx-keep-sgx-shim/target/x86_64-unknown-linux-musl/debug/enarx-keep-sgx-shim` in `cargo make`'s `post-ci-flow`.

Since the ci-flow normally does a build and this is in post-ci-flow, the files should already exist, but in case they don't, the script checks for that.

It only runs for SGX, not SEV, currently.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -62,6 +62,36 @@ command = "${CARGO_WORKSPACE_ROOT}/.tests/cargo-toml-package-license"
 [tasks.misc-licenses-crate]
 command = "${CARGO_WORKSPACE_ROOT}/.tests/misc-licenses-crate"
 
+[tasks.build-sgx-payload-reqs]
+workspace = true
+env = { "CARGO_MAKE_WORKSPACE_INCLUDE_MEMBERS" = ["enarx-keep-sgx", "enarx-keep-sgx-shim", "payload"] }
+run_task = "build"
+
+#Find paths required by the SGX payload test.
+[tasks.find-sgx-paths]
+dependencies = ["build-sgx-payload-reqs"]
+env = { "SGX_KEEP" = { script = [
+"find ${CARGO_WORKSPACE_ROOT}/enarx-keep-sgx/target/ -name enarx-keep-sgx"
+] }, "SGX_SHIM" = { script = [
+"find ${CARGO_WORKSPACE_ROOT}/enarx-keep-sgx-shim/target/ -name enarx-keep-sgx-shim"
+] } }
+
+# Find path for the payload for SGX and (future) SEV tests.
+[tasks.find-payload]
+dependencies = ["build-sgx-payload-reqs"]
+env = { "PAYLOAD" = { script = [ "find ${CARGO_WORKSPACE_ROOT}/payload/target/ -name payload" ] } }
+
+# Detects whether SGX is present on the system.
+[tasks.detect-sgx]
+env = { "SGX_SYSTEM" = { script = [ "[ -e '/dev/sgx/enclave' ] && echo true || echo false" ] } }
+
+# Test that the payload binary runs correctly on SGX.
+[tasks.run-sgx-payload]
+dependencies = ["find-sgx-paths", "find-payload", "detect-sgx"]
+condition = {env_set = ["SGX_KEEP", "PAYLOAD", "SGX_SHIM"], env = { "SGX_SYSTEM" = "true"} }
+command = "${SGX_KEEP}"
+args = ["--code", "${PAYLOAD}", "--shim", "${SGX_SHIM}"]
+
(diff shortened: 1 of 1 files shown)


connorkuehl:
Just for posterity, as this was discussed in standup:

There's another approach here which might be simpler and probably easier to maintain than what was described above.

But the idea is to allow the bash script defined here to take in arguments to the keep, shim, and desired payload. These arguments can be made environment variables in a `Makefile.toml` task which invokes the bash script and passes in these values.

Then we might just have a series of `[tasks.keep-sev]` or `[tasks.keep-sgx]`-like targets that define the path to the proper binaries and then pass them into the script that Lily wrote and they can invoke the bash script using the new arguments functionality.

These `[tasks.keep-*]` tasks can then be grouped under a parent level task which can then join the other CI tasks under the `ci-flow` target.

lkatalin:
A note to self that these were two requirements discussed:
1. Should run `cargo make test` in root directory to determine whether it's on an SEV or SGX host; then run payload(s) to check it still works.
2. Don't break the ability to `cd` into crates and run `cargo build`, `cargo test`, etc.

lkatalin:
@connorkuehl So this current version does everything we want:
- Builds only the 3 dependent crates for running the SGX payload.
- Runs top level tests separately from crate-level tests.
- Makes sure SGX is on the system before running the payload.
- Passes the file paths for the SGX payload into the payload script.

Much of this was achieved through overriding `cargo make`'s default test and dev-test-flow flows. (Those are all the tasks at the bottom.) And forking a lot to enable the "workspace" flag for crate-level tests in the flows.

I have two thoughts:
1. It actually might be worth it to ditch `.tests/run-payload` and just do it in-line in the `cargo make` file. Any objections?
2. This file would look cleaner if some of the other tests that you wrote earlier moved into the crate-level or top-level flows. Then I could remove some of the extra overrides, I think.

lkatalin:
@connorkuehl This new version of the PR makes no-workspace the default, which removes the need for no-workspace-overrides to cargo make's default tasks. This makes it possible to use crate-level extension files, ex. for the `sev` crate, and makes the file less jumbled and complex overall.

It also removes the bash script running the SGX payload and does it in the cargo make file, which seems cleaner. 

One bug is that running `cargo make build-sgx-payload-reqs` by itself invokes the build for all crate members instead of just the 3 relevant ones. However, we should never need to call this as a standalone and it works well when called from any parent task, ex. `find-sgx-paths`.

A note: `cargo make test` will call just crate-level tests; `cargo make test-flow` calls both crate-level and top-level tests. 

Another thought is, do we want to keep all of cargo make's default tests in `ci-flow`? If we don't actually need these, we could remove them. `print-env-flow` doesn't seem that useful to me, for example. 

I filed a separate issue for figuring out why the SGX payload test hangs: #552 .

npmccallum:
I'm honestly not sure what to think of this PR. It feels to me like we are replicating `build.rs` inside `Makefile.toml`. The existing approach will work for a single payload. But I'm not sure it scales as we start writing syscall-specific test binaries.

I'm wondering if it would be cleaner to implement a top-level `tests` crate. This crate could make decisions about what hardware to run on in `build.rs`. Then the only thing we need to do in `Makefile.toml` is make sure that tests are run in this crate after all other crates are built.

@connorkuehl @lkatalin Can you make a counter argument?

connorkuehl:
> I'm honestly not sure what to think of this PR. It feels to me like we are replicating `build.rs` inside `Makefile.toml`. The existing approach will work for a single payload. But I'm not sure it scales as we start writing syscall-specific test binaries.
> 
> I'm wondering if it would be cleaner to implement a top-level `tests` crate. This crate could make decisions about what hardware to run on in `build.rs`. Then the only thing we need to do in `Makefile.toml` is make sure that tests are run in this crate after all other crates are built.
> 
> @connorkuehl @lkatalin Can you make a counter argument?

I've been reading this PR and thinking about it for quite a bit and I agree that a `build.rs` script inside a dummy crate is probably our best bet here.

@lkatalin, I apologize for the whiplash. My zeal for cargo make produced a tremendous amount of work for you.

lkatalin:
@npmccallum I completely agree that this replicates some of the `build.rs` functionality. So the concern is that as we are testing more binaries, the `Makefile.toml` will need to run each of them separately and it would be very cluttered? @connorkuehl No worries, I learned a lot about `cargo make`. I'll re-work this to have a top-level `tests` crate for binaries that runs tests based on `build.rs`.  I'm going to close this PR and submit a new one for the same issue, so that I can look back on this version for future reference.

PR #533, opened 2020-05-14
Enable #![deny(missing_docs)] on the SGX shim

npmccallum:


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -207,7 +207,7 @@ impl<'a> Handler<'a> {
         let code = code
             .into()
             .map(|x| x.into())
-            .unwrap_or(self.aex.gpr.rdi.raw());
+            .unwrap_or_else(|| self.aex.gpr.rdi.raw());
```
lkatalin, line 210: Is this related to missing docs?

npmccallum: I'm not quite sure. I flipped the docs flag and the compiler started warning about this. Either way, I fixed it.

PR #534, opened 2020-05-14
Consolidate sev crates

### LINKED ISSUE #518: Consolidate 'sev' crates
Supersedes: #280 
Obsoletes: #281 
Obsoletes: #113 
Obsoletes: #480 
Affects: #180 
Affects: #112 
Affects: #447 

We are no longer fighting against limitations imposed by the cargo workspace which was one of the motivating factors for decomposing these larger crates into smaller ones. Since this is no longer the case, we can simplify the project structure by consolidating the smaller pieces into 'sev'.

connorkuehl:
The 'sev' crate only "branched" once toward the goal of being broken down into several smaller crates; however, that decision has been reversed. We can effectively consolidate the `sev` components by "unwinding" the changes that produced the `sev-types` crate.

Note that we are dropping the SEV API codes commit. I initially advocated for these to go into the `sev-types` crate in the hopes that they'd be useful to other projects that perhaps might directly interact with the SEV platform. I'm wondering if there's a better upstream home for them because they are not useful to us seeing as we do not directly use these command codes; lower-level layers like the Linux kernel does.

Closes #518


npmccallum:
@connorkuehl We are the upstream crate for SEV. Further, providing access to codes without the data types used has negative value. IMHO, we need to hide the ioctl interface altogether and provide simply a high-level interface.

PR #536, opened 2020-05-14
Create virtual machine w/ a KVM context and simple address space

connorkuehl:
This is a baby step towards loading and running the shim.

Related: #313.

v3 -> v4:
    
* Separate x86_64 constants with whitespace
* Avoid linewrap for guest_pg_addr
* Avoid wrapping for PDPTE start variable
* Drop memmap crate, factor out mmap wrapper and use that instead due to mmap flag requirements
* Use MAP_HUGE_2MB flag for mmap
* Filed a [debt issue](https://github.com/enarx/enarx/issues/555) for the address types

v2 -> v3:

* Rename kvm_fd -> fd
* Alias kvm_userspace_memory_region to MemoryRegion
* Use a default memory value for VM address space size
* Remove unnecessary casts when setting guest_pg_addr
* Don't panic if PhysAddr::try_new fails
* Add comments for the page table constants

v1 -> v2:

* Use units crate for expressing 1GiB
* Remove pml2t_ident from page table
* Remove PDE_START constant
* Use x86_64 crate PageTableFlags
* Use x86_64 crate for representing page tables


enarx-keep-sev/src/main.rs
```diff
@@ -38,7 +41,7 @@ fn run(args: Args) -> Result<(), Box<dyn Error>> {
     let _kernel = File::open(args.shim)?;
     let _code = File::open(args.code)?;
 
-    // TODO: KVM context creation
+    let _vm = vm::VirtualMachine::new(/* 1 GiB */ 2 << 30)?;
```
haraldh, line 44: you can use the units crate from our repo :)

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        unsafe {
+            kvm_fd.set_user_memory_region(region)?;
+        }
+
+        // Set up the page tables
+        let mut page_tables = PageTables::default();
+        page_tables.pml4t[0] = PDPTE_START as u64 | 0x7;
+        page_tables.pml3t_ident[0] = PDE_START as u64 | 0x7;
+        page_tables.pml2t_ident[0] = 0x183u64;
```
haraldh, line 42:
```suggestion
        // Set up the page tables
        page_tables.pml4t[0] = PDPTE_START as u64 | 0x3; // page flags WRITABLE | PRESENT
        page_tables.pml3t_ident[0] = 0x83u64; // page flags HUGE_PAGE | WRITABLE | PRESENT
```

The shim doesn't need it that fine grained anymore.

You may also want to substitute the flags with some bitfield values.

enarx-keep-sev/src/x86_64.rs
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub const PML4_START: usize = 0x9000;
+pub const PDPTE_START: usize = 0xA000;
+pub const PDE_START: usize = 0xB000;
```
haraldh, line 5: not needed anymore

enarx-keep-sev/src/x86_64.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[repr(C)]
+pub struct PageTables {
+    pub pml4t: [u64; 512],
+    pub pml3t_ident: [u64; 512],
+    pub pml2t_ident: [u64; 512],
```
haraldh, line 11: not needed anymore

enarx-keep-sev/src/x86_64.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+impl Default for PageTables {
+    fn default() -> Self {
+        PageTables {
+            pml4t: [0u64; 512],
+            pml3t_ident: [0u64; 512],
+            pml2t_ident: [0u64; 512],
```
haraldh, line 19: not needed anymore

enarx-keep-sev/src/vm.rs
haraldh: https://github.com/rust-osdev/x86_64/blob/master/src/structures/paging/page_table.rs#L107-L169
Now, that I patched the `x86_64` crate to be usable with non-nightly rust.
```toml
x86_64 = { version = "0.11.0", default-features = false, features = ["stable"] }
```

haraldh: https://docs.rs/x86_64/0.11.0/x86_64/structures/paging/page_table/struct.PageTableFlags.html

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::x86_64::*;
+
+use std::error::Error;
+
+use kvm_bindings::kvm_userspace_memory_region;
```
npmccallum, line 7:
```suggestion
use kvm_bindings::kvm_userspace_memory_region as MemoryRegion;
```

I understand that `kvm-bindings` is auto-generated. But seeing a strangely named struct below is jarring. We can just rename it here.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub struct VirtualMachine {
+    kvm: Kvm,
+    kvm_fd: VmFd,
+    address_space: Vec<kvm_userspace_memory_region>,
+    mmap_handles: Vec<MmapMut>,
+}
+
+impl VirtualMachine {
+    pub fn new(mem_size: usize) -> Result<Self, Box<dyn Error>> {
```
npmccallum, line 21: We will, very soon, need to make it possible to grow and shrink the memory allocation to the VM. Can we just define an internal constant for the minimum amount of memory allocated to the VM rather than pawning off this problem to the caller who has no knowledge of the internals?

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,66 @@
...
+        let kvm_fd = kvm.create_vm()?;
+
+        // Create the guest address space
+        let guest_addr_space = memmap::MmapOptions::new().len(mem_size).map_anon()?;
+        let region = kvm_userspace_memory_region {
+            slot: 0,
+            flags: 0,
+            guest_phys_addr: 0,
+            memory_size: mem_size as _,
+            userspace_addr: &*guest_addr_space as *const _ as *const u8 as u64,
```
npmccallum, line 33:
```suggestion
            userspace_addr: &*guest_addr_space as *const _ as u64,
```

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,66 @@
...
+        };
+
+        unsafe {
+            kvm_fd.set_user_memory_region(region)?;
+        }
+
+        // Set up the page tables
+        let mut page_tables = PageTables::default();
+        let pdpte_start =
+            PhysAddr::try_new(PDPTE_START).expect("Couldn't express PDPTE_START as PhysAddr");
```
npmccallum, line 43: Can't we just use the `?` operator here?

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,66 @@
...
+            pdpte_start,
+            PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
+
+        page_tables.pml3t_ident[0].set_flags(
+            PageTableFlags::HUGE_PAGE | PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
+
+        let guest_pg_addr: *mut PageTables =
+            (region.userspace_addr as u64 + PML4_START as u64) as *mut _;
```
npmccallum, line 54:
```suggestion
        let guest_pg_addr = (region.userspace_addr + PML4_START) as *mut PageTables;
```

enarx-keep-sev/src/x86_64.rs
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use x86_64::structures::paging::page_table::PageTable;
+
+pub const PML4_START: u64 = 0x9000;
+pub const PDPTE_START: u64 = 0xA000;
```
npmccallum, line 6: What are these magic numbers? Where do they come from?

connorkuehl: @npmccallum `PML4_START` is the guest physical address of the root of the page table structure. `PDPTE_START` is the first top-level page table entry. However, as I was typing these definitions into my next revision, your question raised a question of my own:

@haraldh should we move the page table setup into the shim? In `demo2020` the keep sets up the page tables and then loads the address into the guest's `cr3` register when we're provisioning a vCPU. I think anything that can be done in the trusted microkernel should be done in the trusted microkernel.

haraldh: If we want 64bit mode, we need a page table for the CPU to be allowed to read instructions from `rip`. The shim can't di it, if it can't run. The addresses of the pages could be more or less dynamic. You just have to fill in the entry with the correct address.

enarx-keep-sev/src/vm.rs
connorkuehl: That was actually my initial stab at it, but I had offended the compiler:

```rust
error[E0606]: casting `*const [u8]` as `u64` is invalid
  --> src/vm.rs:37:29
   |
37 |             userspace_addr: &*guest_addr_space as *const _ as u64,
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: cast through a thin pointer first
```

connorkuehl:
Ping @npmccallum

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+    address_space: Vec<MemoryRegion>,
+    mmap_handles: Vec<MmapMut>,
+}
+
+impl VirtualMachine {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        // Create a KVM context
+        let kvm = Kvm::new()?;
+        let fd = kvm.create_vm()?;
+
```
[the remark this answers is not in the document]
connorkuehl: Yes, and yes. In the event of an `Err`, the `?` operator will return it immediately. Due to Rust's ownership and `Drop` semantics, any opened resources will be managed automatically.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+        let pdpte_start =
+            PhysAddr::try_new(PDPTE_START).map_err(|_| PhysAddrNotValid(PDPTE_START))?;
+        page_tables.pml4t[0].set_addr(
+            pdpte_start,
+            PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
+
+        page_tables.pml3t_ident[0].set_flags(
+            PageTableFlags::HUGE_PAGE | PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
```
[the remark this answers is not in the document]
connorkuehl: I think this helps us keep the page table structures smaller, since bigger pages would result in fewer entries.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl VirtualMachine {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        // Create a KVM context
+        let kvm = Kvm::new()?;
+        let fd = kvm.create_vm()?;
+
+        // Create the guest address space
+        let guest_addr_space = memmap::MmapOptions::new()
+            .len(DEFAULT_MEM_SIZE)
```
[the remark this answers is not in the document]
connorkuehl: I will start differentiating between these terms in future PRs once we have code that needs to deal with host addresses, guest physical addresses (host + offset), and guest virtual addresses. That is, I want to wait until those terms are "live", if that makes sense.

enarx-keep-sev/src/vm.rs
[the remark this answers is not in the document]
whitebrandy: Makes sense. I am not sure how this translates to the NPT mappings - does this map to an actual huge page on the host ? I guess at this stage, we  are not worried about that but this can also lead to a lot of wasted host memory (unless we are actually using that memory, ofcourse!)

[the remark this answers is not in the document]
connorkuehl: I don't know what NPT is, but all page table mappings here should be confined to the big block of host memory I mmap'd to serve as the guest's address space. I don't think the host will be impacted beyond having to supply that big block of memory for the address space.

enarx-keep-sev/src/vm.rs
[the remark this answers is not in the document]
npmccallum: It might be better to have separate newtypes for these. Perhaps we can expend the `memory` crate for this?

enarx-keep-sev/src/vm.rs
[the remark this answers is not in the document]
npmccallum: They should be allocated via `mmap()` with `MAP_HUGE_2MB` on the host side. Again, it would be good if we can enforce type safety for this.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+            .map_anon()?;
+        let region = MemoryRegion {
+            slot: 0,
+            flags: 0,
+            guest_phys_addr: 0,
+            memory_size: DEFAULT_MEM_SIZE as _,
+            userspace_addr: &*guest_addr_space as *const _ as *const u8 as u64,
+        };
```
npmccallum, line 38: I don't see any flags in this allocation asking for huge pages. I'm not sure if that is necessary or not.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        unsafe {
+            fd.set_user_memory_region(region)?;
+        }
+
+        // Set up the page tables
+        let mut page_tables = PageTables::default();
+        let pdpte_start =
+            PhysAddr::try_new(PDPTE_START).map_err(|_| PhysAddrNotValid(PDPTE_START))?;
```
npmccallum, line 47: Is it possible to avoid this wrapping here? Perhaps split this into multiple lines?

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+            pdpte_start,
+            PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
+
+        page_tables.pml3t_ident[0].set_flags(
+            PageTableFlags::HUGE_PAGE | PageTableFlags::WRITABLE | PageTableFlags::PRESENT,
+        );
+
+        let guest_pg_addr: *mut PageTables =
+            (region.userspace_addr + PML4_START) as *mut PageTables;
```
npmccallum, line 58:
```suggestion
        let guest_pg_addr = (region.userspace_addr + PML4_START) as *mut PageTables;
```

I think that avoids the wrap.

enarx-keep-sev/src/x86_64.rs
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use x86_64::structures::paging::page_table::PageTable;
+
+/// The *guest* physical address of the root page table structure.
+pub const PML4_START: u64 = 0x9000;
```
npmccallum, line 6:
```suggestion
pub const PML4_START: u64 = 0x9000;

```

enarx-keep-sev/src/vm.rs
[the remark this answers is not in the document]
connorkuehl: Yes that is my plan once these use cases materialize in the work that follows this PR

enarx-keep-sev/src/vm.rs
connorkuehl: The [guest's page tables enable huge pages for the guest](https://github.com/enarx/enarx/pull/536/files#diff-fd0caaaee80e2f801bb7595524d12efbR54). The keep installs those page tables into the guest's address space.

AIUI, the host's page topology is largely transparent here and I don't think it impacts the construction of our virtual machine nor should it color how the virtual machine uses the pages.

whitebrandy: So, the page isn't backed by an actual hugepage on the host but from what I understand, you just want the reduced complexity of setting up the page tables in the guest ?

npmccallum: I suspect the Linux kernel is likely going to use a huge page implicitly for any large-enough allocation that is a multiple of the huge page size. So I'm not sure if it is strictly necessary to have the flag. Nevertheless, there are several notes buried in KVM about having huge pages on both the guest and the host have performance benefits. So it is probably worth requiring huge pages from the kernel.

enarx-keep-sev/src/vm.rs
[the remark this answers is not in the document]
npmccallum: Yeah, `debt` is fine for this.

enarx-keep-sev/src/vm.rs
connorkuehl: > So, the page isn't backed by an actual hugepage on the host but from what I understand, you just want the reduced complexity of setting up the page tables in the guest ?

Yes, that and we want the systems-level code to be as tiny of a footprint in the guest address space as possible

> Nevertheless, there are several notes buried in KVM about having huge pages on both the guest and the host have performance benefits. So it is probably worth requiring huge pages from the host kernel.

Ah I see I did not know that. I'll research this and follow up on this in my next revision.

whitebrandy: KVM does EPT mappings lazily - so with a huge page, you minimize the number of EPT/NPT exits required. And, now you also need lesser number of TLB flushes since it maps to a bigger page size but you already get a little bit of this benefit even if only the guest knows about huge pages.

enarx-keep-sev/src/vm.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub struct VirtualMachine {
+    kvm: Kvm,
+    fd: VmFd,
+    address_space: Vec<MemoryRegion>,
+    mmap_handles: Vec<mmap::Unmap>,
+}
+
+impl VirtualMachine {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
```
npmccallum, line 22: Aren't all the errors `std::io::Error`?

connorkuehl: Many of them are. The KVM stuff in particular returns `kvm_ioctls::Error`

npmccallum: Ugh... That's why I used `std::io::Error` in `ketuvim`. Everything that comes from an `errno` should be `std::io::Error`, IMHO.

Please file a debt issue for this and I'll approve.

connorkuehl: Filed under https://github.com/enarx/enarx/issues/556

PR #537, opened 2020-05-15
Replace nolibc with libc

### LINKED ISSUE #500: Remove nolibc crate
The `nolibc` crate emerged because we needed `libc` constants and types in a `no_std` environment. However, our previous workspace setup made it impossible to link against the `libc` crate (since all features were coalesced across all subprojects). This has now been fixed in #420.

We should remove the `nolibc` crate and replace it with the upstream `libc` crate using `no_std`.

blazebissar:
Closes #500. 

`libc` utilizes `i32` and `i64` types for its various constants (syscalls, errnos, function argument constants) so this PR primarily addresses this with simple casting. There are times (notably in `event.rs`) where it employs `try_into` to avoid making too many assumptions. Additionally, the `libc::iovec` differs from the `nolibc::Iovec` but is largely compatible. One point of notable divergence is the use of `clone_from` to duplicate `libc::iovec`s due to the lack of support for `copy`, so there is a performance hit. Finally, `libc` does not appear to have any equivalent constants for the `ArchPrctlTask` `enum`, so it now resides in `handler.rs`.

Added some simple tests to confirm that the castings between `libc`s signed integers and the unsigned ones used by Enarx are consistent with the `nolibc` `from_syscall` and `into_syscalll` functions. Would it be better to perform these transformations in dedicated functions to better maintain consistency?

Eager to hear feedback.


enarx-keep-sgx-shim/Cargo.toml
```diff
@@ -9,9 +9,10 @@ license = "Apache-2.0"
 crt0stack = { path = "../crt0stack" }
 intel-types = { path = "../intel-types" }
 sgx-types = { path = "../sgx-types" }
-nolibc = { path = "../nolibc" }
 memory = { path = "../memory" }
 span = { path = "../span" }
+libc = "0.2"
```
haraldh, line 14: you might want for `no_std`
```suggestion
libc = { version = "0.2", features = [] }
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -2,15 +2,25 @@
...
-use nolibc::x86_64::error::Number as ErrNo;
-use nolibc::x86_64::syscall::Number as SysCall;
-use nolibc::{ArchPrctlTask, Iovec};
 use sgx_types::ssa::StateSaveArea;
 use span::{Contains, Line, Span};
 
 use core::{mem::size_of, slice::from_raw_parts_mut};
 
 const TRACE: bool = false;
+// The last 4095 numbers are errrnos.
```
haraldh, line 11: typo

connorkuehl:
Closes #500

npmccallum:
@blazebissar Please see the following for our standard practices:

https://chris.beams.io/posts/git-commit/

https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -32,32 +31,34 @@ pub extern "C" fn event(
...
-                        SysCall::BRK => h.brk(),
-                        SysCall::UNAME => h.uname(),
-                        SysCall::MPROTECT => h.mprotect(),
-                        SysCall::MMAP => h.mmap(),
-                        SysCall::MUNMAP => h.munmap(),
-                        SysCall::RT_SIGACTION => h.rt_sigaction(),
-                        SysCall::RT_SIGPROCMASK => h.rt_sigprocmask(),
-                        SysCall::SIGALTSTACK => h.sigaltstack(),
+                    aex.gpr.rax =
+                        Register::from_raw(match h.aex.gpr.rax.raw().try_into().unwrap() {
```
npmccallum, line 35: We can't `unwrap()` in the shim. It [causes an infinite loop when it fails](https://github.com/enarx/enarx/blob/master/enarx-keep-sgx-shim/src/main.rs#L17).

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -2,15 +2,25 @@
...
+// arch_prctl syscalls not available in the libc crate as of version 0.2.69
+enumerate::enumerate! {
+    #[derive(Copy, Clone)]
+    enum ArchPrctlTask: u64 {
+        ArchSetGs = 0x1001,
+        ArchSetFs = 0x1002,
+        ArchGetFs = 0x1003,
+        ArchGetGs = 0x1004,
+    }
+}
```
npmccallum, line 23: Maybe we should add this to `libc`? @connorkuehl Thoughts?

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -372,15 +390,15 @@ impl<'a> Handler<'a> {
         // array, measured in bytes.
         let mut offset = 0;
         for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-            u.base = ubuffer[offset..].as_mut_ptr();
-            u.size = t.size;
-            offset += t.size;
+            u.iov_base = ubuffer[offset..].as_mut_ptr() as *mut core::ffi::c_void;
```
npmccallum, line 393:
```suggestion
            u.iov_base = ubuffer[offset..].as_mut_ptr() as *mut _;
```

I think we can do this for all of the instances of `c_void` in this patch.

enarx-keep-sgx-shim/src/heap.rs
```diff
@@ -138,12 +137,12 @@ impl Heap {
         fd: isize,
         offset: usize,
     ) -> usize {
-        const RWX: u64 = nolibc::PROT_READ | nolibc::PROT_WRITE | nolibc::PROT_EXEC;
-        const PA: u64 = nolibc::MAP_PRIVATE | nolibc::MAP_ANONYMOUS;
+        const RWX: u64 = libc::PROT_READ as u64 | libc::PROT_WRITE as u64 | libc::PROT_EXEC as u64;
+        const PA: u64 = libc::MAP_PRIVATE as u64 | libc::MAP_ANONYMOUS as u64;
```
npmccallum, line 141: It might be easier to have these values be the same type as the `libc` constants and do a final cast later in the code.

enarx-keep-sgx-shim/src/handler.rs
connorkuehl: @npmccallum Agreed. I opened #540 for this.

blazebissar:
Thank you for the feedback, the latest push should address it with better commit messages too. I also added @npmccallum's links to the How to Contribute wiki page for future newcomers.

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -32,32 +31,42 @@ pub extern "C" fn event(
...
-                        SysCall::RT_SIGPROCMASK => h.rt_sigprocmask(),
-                        SysCall::SIGALTSTACK => h.sigaltstack(),
+                    aex.gpr.rax = Register::from_raw(
+                        match h
+                            .aex
+                            .gpr
+                            .rax
+                            .raw()
+                            .try_into()
+                            .unwrap_or(h.aex.gpr.rax.raw() as i64)
```
npmccallum, line 41: Just unconditionally cast here.

enarx-keep-sgx-shim/src/event.rs
```diff
@@ -32,32 +31,42 @@ pub extern "C" fn event(
...
-                        syscall => {
-                            h.print("unsupported syscall: ");
-                            h.print(&syscall);
-                            h.print("\n");
-                            ErrNo::ENOSYS.into_syscall()
-                        }
-                    });
+                            syscall => {
+                                h.print("unsupported syscall: ");
+                                // Restore the value after forcing cast to i64
```
npmccallum, line 63: The comment is not needed.

enarx-keep-sgx-shim/src/event.rs
npmccallum: I'll fix this in a follow up since I have to touch this line anyway.

npmccallum:
@haraldh I'd like to merge this. Any other objections?

PR #541, opened 2020-05-15
Implement additional Register conversions

npmccallum:


npmccallum:
@blazebissar FYI, this is the PR I mentioned to you. Once this is merged, we can re-type most of our raw integers that intend to represent registers (i.e. `u64`) into `Register<usize>` or `Register<u64>` and we should get free conversions.

PR #546, opened 2020-05-19
Fix verify false positive

### LINKED ISSUE #520: verify() false positives
Somewhere in the call chain amongst our `impl Verifiables` is a bug resulting in a false positive.

For example, I'm working on a feature for `sevctl` to enable the builtin Rome certificates. This feature will try to deduce which set of certificates to use with `sevctl`'s `verify` subcommand. To deduce which certificates to use, the current strategy is to iterate through each generation's certificates and see if it signs the SEV chain's CEK.

However, due to the false positive, this strategy does not work, because first the code will check if the Naples ASK signs the CEK, and the false positive tricks `sevctl` into thinking we're on Naples even if we're on Rome.

Here's a relevant snippet of code:

```rust
impl TryFrom<&sev::Chain> for Generation {
    type Error = ();
    fn try_from(schain: &sev::Chain) -> Result<Self, Self::Error> {
        let naples: ca::Chain = Generation::Naples.into();
        let rome: ca::Chain = Generation::Rome.into();

        Ok(if (&naples.ask, &schain.cek).verify().is_ok() {
            Generation::Naples
        } else if (&rome.ask, &schain.cek).verify().is_ok() {
            Generation::Rome
        } else {
            return Err(());
        })
    }
}
```

Run on Rome (note that it uses the Naples CA chain):

```
PDH EP384 D256 a19ef679477f2256d802ce9054ee6d1145dbeed52998c66ea97499a01d977c54
 ⮤ PEK EP384 E256 a37b51038cbaec1062f7b5f93e8caa55cc7798b1a813c26fae1cabc96320fa0a
   •⮤ OCA EP384 E256 deae2d45b8d1456c94b514fba0e6f6fb9f13afc9df9ba9670e9c9e9e54b4af6e
    ⮤ CEK EP384 E256 ce2cba2df8808d188f445192418e008ecf9bd1464db35ea64afa3b3b6966396c
       ⮤ ASK R2048 R256 5dff2a2a4dd59dc095d73a26af28be523192ade830bdd35bacdd81ee75584c45
         •⮤ ARK R2048 R256 8efddcaaf84990cf6354d93637be270220cf15797332b3d918c3b8cc77349f31
```

Run on Rome after switching the order of the if-statement branches (note that it now uses the Rome CA chain as that's the chain that's checked first):

```
PDH EP384 D256 a19ef679477f2256d802ce9054ee6d1145dbeed52998c66ea97499a01d977c54
 ⮤ PEK EP384 E256 a37b51038cbaec1062f7b5f93e8caa55cc7798b1a813c26fae1cabc96320fa0a
   •⮤ OCA EP384 E256 deae2d45b8d1456c94b514fba0e6f6fb9f13afc9df9ba9670e9c9e9e54b4af6e
    ⮤ CEK EP384 E256 ce2cba2df8808d188f445192418e008ecf9bd1464db35ea64afa3b3b6966396c
       ⮤ ASK R4096 R256 1f668d63e0a7cb120a41bfb1bcefca715564a187305237f058f6e62c3d4ba5b0
         •⮤ ARK R4096 R256 ae2456af6c6ad234badd6c75e58915be61568286766083853bf668e6e23977c3
```

Track down the code that results in the false positive and straighten it out.

connorkuehl:
Closes #520 

OpenSSL's verify() method will only return Err if it experiences some catastrophic error that prevents it from making a conclusive decision as to whether the public key's signature can be verified to be on the digest. It will *NOT* return Err if the key's signature is not on the digest; it will return Ok(false).

Our verification code only checks if OpenSSL returns `Ok`, which accidentally results in false positives. To fix this, we can inspect the interior boolean held by the `Ok` value. If it is `true`, the signature is verified. If it is false, the signature is not on the cert.

Relevant code from OpenSSL:

```rust
pub fn verify(&self, signature: &[u8]) -> Result<bool, ErrorStack> {
    unsafe {
        let r =
            EVP_DigestVerifyFinal(self.md_ctx, signature.as_ptr() as *mut _, signature.len());
        match r {
            1 => Ok(true),
            0 => {
                ErrorStack::get(); // discard error stack
                Ok(false)
            }
            _ => Err(ErrorStack::get()),
        }
    }
}
```

We have simply been assuming that if we get Ok from the verify method that the public key's signature has been found on the digest. We must inspect the interior `bool` to know whether the signature is verified.


whitebrandy:
Good find, looks good to me

PR #562, opened 2020-05-27
Clear extended CPU state on enclave entry and exit

lkatalin:
This relates to #334 , but the xCPU state will also need to be saved and restored before that issue can be closed. This is a preliminary PR to #427 , which I'll rebase onto this if this can be merged.

### DIFF
--- a/enarx-keep-sgx-shim/src/start.S
+++ b/enarx-keep-sgx-shim/src/start.S
@@ -40,6 +40,17 @@
     cld
 .endm
 
+# Clear the extended CPU state
+.macro zerox
+    push    %rax                                    # Save %rax
+    push    %rdx                                    # Save %rdx
+    movq    $~0,                    %rdx            # Set mask for xrstor in %rdx
+    movq    $~0,                    %rax            # Set mask for xrstor in %rax
+    xrstor  xsave(%rip)                             # Clear xCPU state with synthetic state
+    pop     %rdx                                    # Restore %rdx
+    pop     %rax                                    # Restore %rax
+.endm
+
 # Save preserved registers (except %rsp)
 .macro  savep
     push    %rbx
@@ -60,6 +71,21 @@
     pop     %rbx
 .endm
 
+    .section .rodata
+    .align 64
+xsave:                          # An initialized synthetic xsave area
+# Legacy
+    .fill   1, 4, 0x037F        # FCW
+    .fill   5, 4, 0
+    .fill   1, 4, 0x1F80        # MXCSR
+    .fill   1, 4, 0xFFFF        # MXCSR_MASK
+    .fill   60, 8, 0
+
+# Header
+    .fill   1, 8, 0             # XSTATE_BV
+    .fill   1, 8, 1 << 63       # XCOMP_BV (compaction mode)
+    .fill   6, 8, 0
+
 # This function is called during EENTER. Its inputs are as follows:
 #  %rax = The current SSA index. (i.e. %rbx->cssa)
 #  %rbx = The address of the TCS.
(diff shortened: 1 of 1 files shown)


enarx-keep-sgx-shim/src/start.S
```diff
@@ -76,9 +102,10 @@ enclave:
     cmp     $0,                     %rax            # If CSSA > 0...
     jne     .Levent                                 # ... restore stack from AEX[CSSA-1].
 
-    # Clear unused registers, set the stack pointer and jump to Rust
+    # Clear unused registers and xCPU state, set the stack pointer and jump to Rust
     zerop
     zerot
+    zerox
```
npmccallum, line 108: This doesn't work. `zerox` pushes `%rax` and `%rdx` to the stack. But you don't setup the secure stack until below. Therefore, you can't use the stack yet.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -163,6 +191,7 @@ syscall:
 
     xor     %rcx,                   %rcx            # Clear %rcx
     zerof   %rcx                                    # Clear CPU flags
+    zerox                                           # Clear xCPU state
```
npmccallum, line 194: This works, but dangerously. `zerox` pushes `%rdx` and `%rax` to the stack. But you just changed the position of the stack (`%rsp`) a few lines above. Therefore, when you push to the stack you are overwriting legitimate values and corrupting the stack.

enarx-keep-sgx-shim/src/start.S
lkatalin: @npmccallum Would it make sense to move this two lines down after `mov  STACK(%rcx),  %rsp`?

enarx-keep-sgx-shim/src/start.S
lkatalin: I see, `mov  0x48(%rsp), %rsp` set the stack pointer back to where the untrusted nonvolatile register states are saved.

enarx-keep-sgx-shim/src/start.S
npmccallum: I suspect it might be more readable to move the stack setup line up to the top and do all the zeroing afterwards.

lkatalin:
@npmccallum This should be using the correct stack, and `zerox` is invoked on entry as well as the two situations leading to an exit (pre- and post-syscall). I didn't consolidate those by putting `zerox` in `.Leexit` because the stack isn't necessarily available in `.Leexit`.

lkatalin:
@npmccallum Let me know what you think of this version; hopefully I understood your suggestions correctly.

enarx-keep-sgx-shim/src/start.S
```diff
@@ -156,13 +184,15 @@ enclave:
 syscall:
     savep                                           # Save preserved registers
     mov     %rsp,                   SRSP(%rcx)      # Save restoration stack pointer
+    
+    zerox                                           # Clear xCPU state
+    zerof   %rcx                                    # Clear CPU flags
+    xor     %rcx,                   %rcx            # Clear %rcx
```
npmccallum, line 190:
```suggestion
    xor     %rcx,                   %rcx            # Clear %rcx
    zerof   %rcx                                    # Clear CPU flags
    zerox                                           # Clear xCPU state
```

npmccallum: See the documentation for `zerof` for why.

lkatalin: `%rcx` must be zero ... got it.

PR #570, opened 2020-05-31
keep-runtime example improvements

### LINKED ISSUE #569: keep-runtime example improvements
**Description**
The existing example in keep-runtime/src/main.rs requires:

1. fixing - the example as is does not work, as it references a non-existent file (`fixtures/return_1.wasm`).  The available file (`fixtures/return_1.wat`) does actually work fine, as does the equivalent .wasm file (constructed using the WABT project's `wat2wasm` utility).  Plan: add the wasm file to the repo (and corresponding files for the other examples).
2. improving - it suggests a `cargo run` as the example.  While an initial `cargo build` is required, using the binary produced thereafter is simpler than the example given (and shouldn't require rust to be installed on the system).  Plan: provide further examples.

MikeCamel:
Added .wasm files in "fixtures" (generated with "wat2wasm")
Added example in src/main.rs to show use of generated binary

Fix #569

### DIFF
--- a/keep-runtime/src/main.rs
+++ b/keep-runtime/src/main.rs
@@ -5,10 +5,10 @@
 //! It can be used to run a Wasm file with given command-line
 //! arguments and environment variables.
 //!
-//! ## Example invocation
+//! ## Example build and invocation
 //!
 //! ```console
-//! $ RUST_LOG=keep_runtime=info RUST_BACKTRACE=1 cargo run target/debug/fixtures/return_1.wasm
+//! $ RUST_LOG=keep_runtime=info RUST_BACKTRACE=1 cargo run fixtures/return_1.wat
 //!    Compiling keep-runtime v0.1.0 (/home/steveej/src/job-redhat/enarx/github_enarx_enarx/keep-runtime)
 //!     Finished dev [unoptimized + debuginfo] target(s) in 4.36s
 //!      Running `target/debug/keep-runtime`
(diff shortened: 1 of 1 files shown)


ueno:
@MikeCamel we had the `.wasm` files in the initial version, but replaced them with the `.wat` files during the review, because wasmtime transparently handles both `.wasm` and `.wat` files.

MikeCamel:
> @MikeCamel we had the `.wasm` files in the initial version, but replaced them with the `.wat` files during the review, because wasmtime transparently handles both `.wasm` and `.wat` files. Maybe the example could directly run the `.wat` file?

I thought that might have been the case.  Having the example showing the .wat file is one option.  However, as it's .wasm files that we'll want to be supporting in the long run (it is the standard binary format, after all, and typically much smaller than the .wat format), I think we should definitely show that we support .wasm.

MikeCamel:
> @MikeCamel We definitely want to avoid committing things in git that not human readable. So perhaps a better approach would be to make the runtime accept either `.wat` or `.wasm` files as input. For example, the first four bytes of a `.wasm` contain the magic number: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure You could use that to detect a `.wasm` file and fall back to parsing `.wat`.

I think probably the easiest is to change the original example to use just the .wat examples, and remove the .wasm files.

Now that I have Rust compiling to .wasm, and writing to the local filesystem, I can write some simple examples in Rust, and add those to the repo.  I'm going to revert some of the changes in this PR and create a new issue + PR pair for that.

ueno:
> wasmtime transparently handles both .wasm and .wat files

This is no longer true with the latest wasmtime: the method that takes data now assumes that it is in the binary format. A couple of new methods are added to complement it, though they take `AsRef<Path>` and determine the file type based on the file extensions.

As it's unclear whether we will have filesystem abstraction in Keep, it might actually be a better idea to use the binary format extensively (for tests/examples, we can generate the files at build time).

keep-runtime/src/main.rs
```diff
@@ -18,6 +18,19 @@
 //!         ),
 //!     ]
 //! ```
+//! ## Example build and invocation with binary
+//!
+//! ```console
+//! $ cargo build --release
+//! Finished release [optimized] target(s) in 0.06s
+//! $ RUST_LOG=keep_runtime=info RUST_BACKTRACE=1 ./target/x86_64-unknown-linux-musl/release/
+//!    keep-runtime fixtures/wasi_snapshot1.wat
```
ueno, line 27: s/wat/wasm/ ? Note that the `.wasm` file is now generated under `OUT_DIR` at build time.

mbestavros:
Completed as part of other work.

PR #572, opened 2020-06-01
Load the SEV shim into the guest address space

connorkuehl:
Related: #313 

Unfortunately we can't execute the code that this PR is loading into the guest but PRs that build on top of this one will get us all the way there.

### DIFF
--- a/enarx-keep-sev/src/vm/mem.rs
+++ b/enarx-keep-sev/src/vm/mem.rs
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub use kvm_bindings::kvm_userspace_memory_region as KvmUserspaceMemoryRegion;
+use mmap::Unmap;
+use span::Span;
+use x86_64::VirtAddr;
+
+pub struct Region {
+    kvm_region: KvmUserspaceMemoryRegion,
+    _backing: Unmap,
+}
+
+impl Region {
+    pub fn new(kvm_region: KvmUserspaceMemoryRegion, backing: Unmap) -> Self {
+        Self {
+            kvm_region,
+            _backing: backing,
+        }
+    }
+
+    pub fn as_virt(&self) -> Span<VirtAddr, u64> {
+        Span {
+            start: VirtAddr::new(self.kvm_region.userspace_addr),
+            count: self.kvm_region.memory_size,
+        }
+    }
+}
--- a/enarx-keep-sev/src/vm/mod.rs
+++ b/enarx-keep-sev/src/vm/mod.rs
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pub mod builder;
+mod mem;
+
+pub use builder::Builder;
+
+use mem::Region;
+
+use kvm_ioctls::{Kvm, VmFd};
+
+pub struct VirtualMachine {
+    _kvm: Kvm,
+    _fd: VmFd,
+    address_space: Region,
+}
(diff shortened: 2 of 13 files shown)


enarx-keep-sev/src/vm/mem.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl AddressSpace {
+    pub fn new() -> Self {
+        Self { regions: vec![] }
+    }
+
+    pub fn add_region(&mut self, region: Region) {
+        self.regions.push(region);
+    }
```
npmccallum, line 35: I think the general strategy of Rust (maybe I'm wrong) is to just make `regions` public rather than focusing on encapsulation.

npmccallum: Which basically means that: `type AddressSpace = Vec<Region>`.

enarx-keep-sev/src/vm/load.rs
```diff
@@ -0,0 +1,57 @@
...
+                    0u8,
+                    ph.p_memsz as usize - ph.p_filesz as usize,
+                );
+            }
+        }
+    }
+
+    let virt_entry = address_space.to_host_virt_addr(PhysAddr::new(elf.entry))?;
+    Ok(virt_entry)
+}
```
npmccallum, line 57: Do you have interest in sharing [this code](https://github.com/enarx/enarx/blob/master/enarx-keep-sgx/src/component.rs)? It currently has one SGX data type, but we could probably fix that. I can think of lots of reasons why this might be beneficial:

1. Common validation
2. Common error types
3. Common user error messages

connorkuehl: I'm experimenting with this right now, thanks!

loader/src/lib.rs
```diff
@@ -1,43 +1,59 @@
...
+    fn from_header(ph: &ProgramHeader) -> Self;
+}
+
+impl FromHeader for () {
+    fn from_header(_ph: &ProgramHeader) -> Self {}
+}
+
 /// A loadable segment of code
-pub struct Segment {
+pub struct Segment<T: FromHeader> {
```
npmccallum, line 40: Rather than this approach, I think we should make two separate segment types. One for the `Builder` and one for the `loader`. Then `enarx-keep-sgx-shim` can carry a function to convert between the two. In particular, I want to avoid a dependency between the `loader` and the `Builder`. This way when we merge the `sgx` libraries, I can move the `Builder` code from here to there without adding a dependency.

enarx-keep-sev/src/vm/mem.rs
```diff
@@ -0,0 +1,47 @@
...
+impl Region {
+    pub fn new(kvm_region: KvmUserMemoryRegion, unmap: Unmap) -> Self {
+        Self {
+            _kvm_region: kvm_region,
+            _unmap: unmap,
+        }
+    }
+}
+
+pub type AddressSpace = Vec<Region>;
```
npmccallum, line 24: I have a sneaking suspicion that this type really belongs as a sub-feature of a `VirtualMachine` type instead.

enarx-keep-sev/src/vm/mod.rs
```diff
@@ -70,11 +73,30 @@ impl VirtualMachine {
                 .write(page_tables);
         }
 
+        let mut address_space = AddressSpace::new();
+        address_space.push(Region::new(region, unmap));
+
```
haraldh, line 78:
```suggestion
        let mut address_space = AddressSpace::new();
        address_space.push(Region::new(region, unmap));

        // Install the page tables into the guest address space
        unsafe {
            mem::to_host_virt_addr(&address_space, PML4_START)?
                .as_mut_ptr::<PageTables>()
                .write(page_tables);
        }
```

Issue #525, opened 2020-05-12
sevctl: status - unreadable char


haraldh:
```console
$ cargo run -q -- verify
PDH EP384 D256 a19ef679477f2256d802ce9054ee6d1145dbeed52998c66ea97499a01d977c54
 ⮤ PEK EP384 E256 a37b51038cbaec1062f7b5f93e8caa55cc7798b1a813c26fae1cabc96320fa0a
   •⮤ OCA EP384 E256 deae2d45b8d1456c94b514fba0e6f6fb9f13afc9df9ba9670e9c9e9e54b4af6e
    ⮤ CEK EP384 E256 ce2cba2df8808d188f445192418e008ecf9bd1464db35ea64afa3b3b6966396c
       ⮤ ASK R4096 R256 1f668d63e0a7cb120a41bfb1bcefca715564a187305237f058f6e62c3d4ba5b0
         •⮤ ARK R4096 R256 ae2456af6c6ad234badd6c75e58915be61568286766083853bf668e6e23977c3
```

I haven't found a font, which displays `⮤` correctly.. 

https://github.com/enarx/enarx/blob/d9d5c24f2ebc802b906c210193935bed4edf82cc/sevctl/src/main.rs#L288

ueno:
According to `fc-list :charset=2BA4`, the following fonts seem to have the coverage:
```console
/usr/share/fonts/vlgothic/VL-Gothic-Regular.ttf: VL Gothic,VL ゴシック:style=regular
/usr/share/fonts/gdouros-symbola/Symbola.ttf: Symbola:style=Regular
```

haraldh:
* vlgothic-fonts: is in group `@Fonts` but marked optional
* gdouros-symbola-fonts: is in group `@Fonts` but marked optional

haraldh:
hmm... still no clue, what the `•` means here in this context

haraldh:
@connorkuehl , @npmccallum even if we only have linux machines, there are people using other OSes to ssh into those machines, so we should probably care about the default MacOS fonts.

npmccallum:
@haraldh ACK. Patches welcome.

PR #588, opened 2020-06-03
sevctl: better output for `verify`

haraldh:
`⮤` looks nice, but is not included in many fonts, yet.
`⬑` seems to be included in way more fonts.

This patch also adds a legend and a `slash-through` for invalid
signatures (for the color blinded).

Fixes: https://github.com/enarx/enarx/issues/525

### DIFF
--- a/sevctl/src/main.rs
+++ b/sevctl/src/main.rs
@@ -275,7 +275,7 @@ mod verify {
         err |= status("   ", &schain.cek, &schain.pek, quiet);
         err |= status("      ", &cchain.ask, &schain.cek, quiet);
         err |= status("         ", &cchain.ark, &cchain.ask, quiet);
-
+        println!("\n • = self signed, ⬑ = signs, •̷ = invalid self sign, ⬑̸ = invalid signs");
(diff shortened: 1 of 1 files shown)


connorkuehl:
I agree that we have an accessibility issue here especially since we are using the colors red and green; however, I'm wondering if the way forward is switching to an ASCII text string rather than changing glyphs around.

We don't make any guarantees about our output on `stdout`, but using regular text might provide more opportunities for interoperability in the event someone tries to parse sevctls output programmatically.

We could probably still keep the tree structure, but have some short text for each entry saying something like "ARK signs ASK... ok" "ASK signs CEK... ok" -- I'm just thinking out loud here.

What do you guys think?

haraldh:
> We don't make any guarantees about our output on `stdout`, but using regular text might provide more opportunities for interoperability in the event someone tries to parse sevctls output programmatically.

For that we should have a machine parseable output format, like json.

haraldh:
> @haraldh We need a rebase for tests to pass.

@npmccallum done

PR #590, opened 2020-06-04
Use certificate size to select the correct message digest

### LINKED ISSUE #548: Add unit tests for Rome certificate chains
The `sev` crate already has a number of unit tests for encoding/decoding and verifying the Naples generation chain under `sev/tests/`

Add similar tests for Rome certificate chains.

### LINKED ISSUE #581: Verify Rome ARK and ASK correctness
#509 and #546 introduce the Rome ARK and ASK certificates, respectively.

However, `sevctl verify` output suggests the ARK is not self-signed and that the ARK does not sign the ASK. Both of those entries are colored red.

`sevctl verify` works correctly for Naples.

connorkuehl:
Fixes #581
Closes #548

Edit: I have also tested `sevctl verify` on both Naples and Rome and now the certificate chains are all verifying correctly with this patch.


sev/src/certs/crypto.rs
```diff
@@ -110,22 +110,34 @@ where
...
+            // Common case
+            sign::Verifier::new(sig.hash, &self.key)?,
+            // Rome CA certificate quirk
+            // The Naples ARK/ASK use a SHA256 digest whereas the Rome
+            // ones use a SHA384 digest. These certificates do not explicitly
+            // advertise which one we should use here, and there is no convenient
+            // way to deduce which kind of certificate we have here except to see
+            // if we can successfully verify the certificate with SHA384.
+            sign::Verifier::new(openssl::hash::MessageDigest::sha384(), &self.key)?,
+        ];
```
npmccallum, line 123: This is, unfortunately, precisely what we **MUST NOT** do. You've now just given an attacker two different ways to get a false positive for every certificate.

Instead, you need to change the values [here](https://github.com/enarx/enarx/blob/master/sev/src/certs/ca/cert/v1.rs#L235) and [here](https://github.com/enarx/enarx/blob/master/sev/src/certs/ca/cert/v1.rs#L260) based on the **size** of the RSA key in the certificate. 2k RSA always uses SHA256 and 4k RSA always uses SHA384.

PR #591, opened 2020-06-04
Add running SGX payload to test suite

### LINKED ISSUE #522: Run payload during cargo make test
We can currently run `cargo make test` to run our tests. When we run these on an SGX enabled system, we also want to run an integration test that executes the payload binary in an SGX enclave.

lkatalin:
It runs when ~~`cargo make test`~~ `cargo make integration` or ~~`cargo make ci-flow`~~ `cargo make integration-ci` are invoked.

Resolves #522 .

This PR does the following:
- Creates a new ~~`tests/`~~ `integration-tests/` crate for top-level tests for our emulated workspace. Currently there is one file included `sgx_payload.rs`, which checks for correct binary paths and runs the SGX payload.

- Adds new tasks to `Makefile.toml` to ensure that crates are built before the tests in ~~`tests/`~~ `integration-tests/` are run. Currently this means *all* tests run after all builds, as per [this discussion](https://github.com/sagiegurari/cargo-make/issues/429). ~~Some default `cargo make` tasks are overwritten here to ensure that `cargo make test` and `cargo make ci-flow` actually call our versions of those tasks, since our versions ensure that crates build first.~~

I tested this by:
- Adding the ~~`tests/`~~ `integration-tests/` crate as one of the first crates listed in the emulated workspace in `Makefile.toml` and running ~~`cargo make test`~~ `cargo make integration` and ~~`cargo make ci-flow`~~ `cargo make integration-ci` after a `cargo make clean`. 
- I also tested that if the `cfg` for the test itself in `sgx_payload.rs` is changed to `has_sgx` instead of `not(has_sgx)`, the test will in fact be ignored. I did not test on a non-SGX machine.

### DIFF
--- a/integration-tests/Makefile.toml
+++ b/integration-tests/Makefile.toml
@@ -1,0 +1,10 @@
+# `cargo make test` is disabled for this crate so that 
+# integration tests are not run with unit tests; specifically, tests
+# in this crate should only be run after builds have completed in all
+# other crates in the emuated workspace. 
+#
+# Tests in this crate can be invoked instead with `cargo make 
+# integration` at the top level, or simply by running `cargo test` 
+# inside this crate.
+[tasks.test]
+disabled = true
--- a/integration-tests/src/main.rs
+++ b/integration-tests/src/main.rs
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests
+
+#![deny(missing_docs)]
+#![deny(clippy::all)]
+
+fn main() {}
--- a/integration-tests/tests/sgx_payload.rs
+++ b/integration-tests/tests/sgx_payload.rs
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains integration tests, including the SGX payload test.
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+/// Finds an absolute file path for a file in a directory.
+#[cfg_attr(not(has_sgx), ignore)]
+fn find_filepath(dir: PathBuf, name: &str) -> Result<PathBuf, std::io::Error> {
+    let file = Command::new("find")
+        .current_dir(&dir)
+        .arg("-name")
+        .arg(name)
+        .output()
+        .expect("could not find file")
+        .stdout;
+
+    let path = dir.join(String::from_utf8(file).unwrap().trim());
+    Ok(std::fs::canonicalize(&path)?)
+}
+
+/// This test runs the payload in the SGX keep using the SGX shim.
+#[cfg_attr(not(has_sgx), ignore)]
+#[test]
+fn sgx_payload() {
+    // Define directories
+    let wksp_root = Path::new(env!("CARGO_MANIFEST_DIR")).parent().unwrap();
+    let sgx_keep_dir = wksp_root.join("enarx-keep-sgx/target/");
+    let payload_dir = wksp_root.join("payload/target/x86_64-unknown-linux-musl/");
+    let sgx_shim_dir = wksp_root.join("enarx-keep-sgx-shim/target/x86_64-unknown-linux-musl/");
+
+    // Find the current SGX Keep, payload, and SGX shim
+    let keep = find_filepath(sgx_keep_dir, "enarx-keep-sgx").unwrap();
+    let payload = find_filepath(payload_dir, "payload").unwrap();
+    let shim = find_filepath(sgx_shim_dir, "enarx-keep-sgx-shim").unwrap();
+
+    // Run the test
+    let mut payload_test = Command::new(keep)
+        .current_dir(wksp_root)
+        .arg("--code")
+        .arg(payload)
+        .arg("--shim")
+        .arg(shim)
+        .spawn()
+        .expect("failed to run sgx payload test");
+
+    let ecode = payload_test.wait().expect("failed to wait on child");
+
+    assert!(ecode.success());
+}
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -8,6 +8,7 @@ CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = [
 	"enarx-keep-sgx",
 	"enarx-keep-sgx-shim",
 	"enumerate",
+	"integration-tests",
 	"intel-types",
 	"iocuddle",
 	"iocuddle-sgx",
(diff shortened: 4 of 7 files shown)


lkatalin:
My understanding is that the non-passing CI isn't on my end currently.

lkatalin:
@connorkuehl 
I actually can't remember why we wanted to move the payload test into `src/main.rs`. It's there now but succeeds silently, which is ... maybe okay, but not very rewarding. A failure would probably be more difficult to understand than if it were a `#[test]` also.

`cargo make test` calls all tests not in `integration-tests/`.
`cargo make integration` calls `build` and `run` in `integration-tests/`
`cargo make ci-flow` does what it did before
`cargo make integration-ci` does the `ci-flow` and then `integration`... it can't be put in `post-ci` without invoking `integration` on every single crate. It is impossible.

My thought is to probably restore the payload test as a `test`, but otherwise this should work.

integration-tests/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+
+/// This test runs the payload in the SGX keep using the SGX shim.
+#[cfg_attr(not(has_sgx), ignore)]
+fn sgx_payload() {
+    // Define directories
+    let wksp_root = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .unwrap()
+        .to_str()
+        .unwrap();
```
connorkuehl: I think if we remove the `.to_str()`, the following lines of code could be more like:

```rust
let sgx_keep_dir = wksp_root.join("enarx-keep-sgx/target/");
let payload_dir = wksp_root.join("payload/target/x86_64-unknown-linux-musl/");
let sgx_shim_dir = wksp_root.join("enarx-keep-sgx-shim/target/");
```

this would require removing some of the later `.to_owned()` below.

integration-tests/tests/sgx_payload.rs
```diff
@@ -0,0 +1,76 @@
...
+    let payload = payload_dir.to_owned() + &find_payload[2..find_payload.len() - 1];
+    let sgx_shim = sgx_shim_dir.to_owned() + &find_shim[2..find_shim.len() - 1];
+
+    let result = Command::new(sgx_keep)
+        .current_dir(wksp_root.to_owned())
+        .arg("--code")
+        .arg(payload)
+        .arg("--shim")
+        .arg(sgx_shim)
+        .output()
```
connorkuehl, line 72: How about `.spawn()` instead of `.output()`? This gives us a handle to the child process, so we can still see its [stdout](https://doc.rust-lang.org/std/process/struct.Child.html#structfield.stdout) and [stderr](https://doc.rust-lang.org/std/process/struct.Child.html#structfield.stderr), but also we can [kill](https://doc.rust-lang.org/std/process/struct.Child.html#method.kill) it. This way if the shim panics we can still kill the child process with a timer and the test fails.

integration-tests/tests/sgx_payload.rs
```diff
@@ -0,0 +1,76 @@
...
+/// This test runs the payload in the SGX keep using the SGX shim.
+#[cfg_attr(not(has_sgx), ignore)]
+#[test]
+fn sgx_payload() {
+    // Define directories
+    let wksp_root = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .unwrap()
+        .to_str()
+        .unwrap();
```
connorkuehl, line 20: I wonder if we remove the `.to_str()` then just below turns into something like this:

```rust
let sgx_keep_dir = wksp_root.join("enarx-keep-sgx/target");
let payload_dir = wksp_root.join("payload/target/x86_64-unknown-linux-musl");
let sgx_shim_dir = wksp_root.join("enarx-keep-sgx-shim/target");
```

then I think we can do away with the `.to_owned()`s further down.

integration-tests/tests/sgx_payload.rs
```diff
@@ -0,0 +1,76 @@
...
+            .output()
+            .expect("failed")
+            .stdout,
+    )
+    .unwrap();
+
+    // Trim off newline and current dir from results strings; join file name to path
+    let sgx_keep = sgx_keep_dir.to_owned() + &find_keep[2..find_keep.len() - 1];
+    let payload = payload_dir.to_owned() + &find_payload[2..find_payload.len() - 1];
+    let sgx_shim = sgx_shim_dir.to_owned() + &find_shim[2..find_shim.len() - 1];
```
connorkuehl, line 64: Let's use absolute file paths here with [fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html).

I think something like this:

```rust
let sgx_keep = std::fs::canonicalize(sgx_keep_dir);
```

lkatalin:
Very helpful comments @connorkuehl , thank you!

connorkuehl:
@lkatalin This is what I had in mind (this requires the logic to be moved out of the `#[test]` function and back into `main.rs`). I've tested it locally and it executes the payload only once:

```
[tasks.run]
command = "cargo"
args = ["run"]

[tasks.integration]
condition = { env = {"CARGO_MAKE_CRATE_NAME" = "integration-tests"} }
run_task = { name = "run" }

[tasks.post-ci-flow]
dependencies = [
	"integration"
]
```

integration-tests/tests/sgx_payload.rs
lkatalin: This is very cool and I did not know it existed. Unfortunately doesn't seem to be able to handle the `\n` at the end, but using it is still shorter.

lkatalin:
@connorkuehl Thanks for the tips on formatting those paths. I made a separate function for this, which I think makes the code shorter and cleaner. If you have any ideas on how to further shorten `find_filepath()`, I'm all ears.

Re: the Makefile logic above, I tried it out but I'm unsure what the advantages are over the current version aside from being able to call `ci-flow` directly to incorporate the tests (and we had already decided we don't care much about having to call custom commands). In addition, I find it has these disadvantages over the current form:
- Calling only the integration tests with `cargo run` instead of `cargo test` doesn't sit well with me from a consistency standpoint if we can avoid it.
- The version of `[tasks.integration]` does recurse into all directories, which is unnecessary and makes the output more verbose without including extra information.
- I found that the "build" dependency did not help ensure that those crates are built before the payload test is called; I tried it after running a `cargo make clean` and it didn't work for me.

So unless there is something I missed, I think keeping it the way it is now makes more sense.

integration-tests/tests/sgx_payload.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut payload_test = Command::new(keep)
+        .current_dir(wksp_root)
+        .arg("--code")
+        .arg(payload)
+        .arg("--shim")
+        .arg(shim)
+        .spawn()
+        .expect("failed to run sgx payload test");
+
+    let ecode = payload_test.wait().expect("failed to wait on child");
```
connorkuehl, line 51: No action required here, but just a note that I filed #609 that will build on this nicely.

integration-tests/tests/sgx_payload.rs
```diff
@@ -0,0 +1,54 @@
...
+    let file = Command::new("find")
+        .current_dir(&dir)
+        .arg("-name")
+        .arg(name)
+        .output()
+        .expect("could not find file")
+        .stdout;
+
+    let path = dir.to_str().unwrap().to_owned() + &String::from_utf8(file).unwrap();
+    Ok(std::fs::canonicalize(&path[..path.len() - 1])?)
```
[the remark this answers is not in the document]
lkatalin: I was really wanting some kind of functionality like `trim()`. Thanks!

integration-tests/Makefile.toml
```diff
@@ -0,0 +1,2 @@
+[tasks.test]
```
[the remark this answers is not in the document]
lkatalin: Good idea.

integration-tests/tests/sgx_payload.rs
lkatalin: Good point, thanks for filing that!

PR #595, opened 2020-06-05
Update tests to use merged Enarx dev container

mbestavros:
Following #594.


npmccallum:
@mbestavros It looks like a test is failing. I'm not sure why.

PR #607, opened 2020-06-08
Inform users of `cargo make` when `cargo build/test` is used

### LINKED ISSUE #598: Block regular "cargo x" (i.e., "cargo build") at the root of the workspace
Our repo _looks_ like a Rust workspace but it is not. Running `cargo build` or some other target at the root is insufficient. We use `cargo-make` for this.

Since this goes slightly against the grain of most Rust project workflows only using "cargo x" commands, the current suggestion is to add a `Cargo.toml` to the root of our repo with a `build.rs` that vacuously fails and emits a message reminding the user to use `cargo-make`.

Ideally, this will not conflict with our regular usage of `cargo-make`.

mbestavros:
Display a quick error message whenever `cargo build`, `cargo test` or similar is invoked. Links back to the `cargo-make` repo and lists a few basic `cargo make` commands.

Resolves #598.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,24 @@
+// Let people using the standard `cargo x` syntax 
+fn main() {
+    
+    panic!("
+    
+    Enarx uses `cargo-make` to build and run its code:
+    https://github.com/sagiegurari/cargo-make
+
+    To build Enarx:
+
+    1. Install cargo-make.
+    $ cargo install cargo-make
+
+    2. Build Enarx.
+    $ cargo make build
+
+    3. You can optionally run tests.
+    $ cargo make test
+
+    For more information, see:
+    https://github.com/enarx/enarx/wiki/How-to-contribute-code#enarx-development-environment-set-up
+    
+    ");
+}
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -0,0 +1,14 @@
+// Let people using the standard `cargo x` syntax 
+fn main() {
+    
+    panic!("
+    
+    Enarx uses `cargo-make` to build and run its code:
+    https://github.com/sagiegurari/cargo-make
+
+    You can try `cargo make build` or `cargo make test`.
+
```
connorkuehl, line 10: I don't know where my last comment went :sweat_smile: 

but my suggestion would be to explicitly state here, near the end:

"To install cargo-make, run: cargo install cargo-make"

mbestavros: Done!

npmccallum: Perhaps something like:
```
To build Enarx:

1. Install cargo-make.
$ cargo install cargo-make

2. Build Enarx.
$ cargo make build

3. You can optionally run tests.
$ cargo make test

For more information, see:
https://github.com/enarx/wiki/...
```

connorkuehl:
@mbestavros please rebase

Issue #363, opened 2020-03-23
crt0stack should take [usize]


npmccallum:
Currently, `crt0stack` works in terms of `&[u8]` buffers. However, this almost guarantees that alignment will be incorrect, wasting space in the buffer. We should switch to `&[usize]` buffers instead.

FYI: @haraldh

haraldh:
yeah

npmccallum:
It appears that the alignment requirement is 16 bytes for x86_64. But `usize` only aligns to 8 bytes. `u128` only aligns to 8 bytes too. So it might not be straightforward.

haraldh:
Do you think this is still an issue?

npmccallum:
@haraldh We should probably do a sensible `align_to()` inside the constructor.

haraldh:
@npmccallum 

Hmm, thinking about this, we have a stack where we fill the bottom of the slice.
And we already align there. And the resulting handle is somewhere near the bottom.

npmccallum:
I think we only align to 8 there. We need to align to at least 16. But our SGX code crashes if we don't align to 32 bytes, I don't know why.

haraldh:
This is fixed now, but we might want to wait to close it until #613 is merged

PR #613, opened 2020-06-15
Add crt0 stack alignment test

jyotsna-penumaka:
Resolves: #363

### DIFF
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -959,4 +959,43 @@ mod tests {
         assert_eq!(reader.next(), None); // terminator
         assert_eq!(reader.next(), None); // don't overrun
     }
+
+    #[test]
+    fn crt0_stack_alignment() {
+        // Prepare the crt0 stack.
+        #[repr(C, align(32))]
+        struct Aligned<T>(T);
+
+        let mut aligned = Aligned([0u8; 1024]);
+        let mut error = false;
+        for i in 0..32 {
+            let mut builder = Builder::new(&mut aligned.0[i..]);
+            builder.push("arg").unwrap();
+            // Set the environment
+            let mut builder = builder.done().unwrap();
+            builder.push("LANG=C").unwrap();
+
+            // Set the aux vector
+            let mut builder = builder.done().unwrap();
+            builder.push(&Entry::ExecFilename("/init")).unwrap();
+            builder.push(&Entry::Platform("x86_64")).unwrap();
+            builder.push(&Entry::Uid(1000)).unwrap();
+            builder.push(&Entry::EUid(1000)).unwrap();
+            builder.push(&Entry::Gid(1000)).unwrap();
+            builder.push(&Entry::EGid(1000)).unwrap();
+            builder.push(&Entry::PageSize(4096)).unwrap();
+            builder.push(&Entry::Secure(false)).unwrap();
+            builder.push(&Entry::ClockTick(100)).unwrap();
+            builder.push(&Entry::Flags(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap2(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::Random([0u8; 16])).unwrap();
+
+            let handle = builder.done().unwrap();
+            let alignment = (handle.start_ptr() as *const () as usize) % STACK_ALIGNMENT;
+            eprintln!("offset: {}, alignment: {}", i, alignment);
+            error |= alignment != 0;
+        }
+        assert!(!error);
+    }
 }


crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+
+    #[test]
+    #[should_panic]
+    fn crt0_stack_alignment() {
+        // Prepare the crt0 stack.
+        #[repr(C, align(32))]
+        struct Aligned(u8);
+        let mut crt0 = [0u8; 1024];
+        let crt0 = unsafe { crt0.align_to_mut::<Aligned>().1 };
+        let crt0 = unsafe { crt0.align_to_mut::<u8>().1 };
```
npmccallum, line 937:
```suggestion
        // Prepare the crt0 stack.
        #[repr(C, align(32))]
        struct Aligned<T>(T);

        let mut aligned = Aligned([u8; 1024]);
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+        for i in 0..32 {
+            let crt0 = &mut crt0[i..];
+            let rand = unsafe { core::mem::transmute([0u8; 16]) };
+            //Set the arguments
+
+            let mut builder = Builder::new(crt0);
```
npmccallum, line 943:
```suggestion
            let mut builder = Builder::new(&mut aligned.0[i..]);
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+            builder.push(&Entry::EUid(1000)).unwrap();
+            builder.push(&Entry::Gid(1000)).unwrap();
+            builder.push(&Entry::EGid(1000)).unwrap();
+            builder.push(&Entry::PageSize(4096)).unwrap();
+            builder.push(&Entry::Secure(false)).unwrap();
+            builder.push(&Entry::ClockTick(100)).unwrap();
+            builder.push(&Entry::Flags(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap2(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::Random(rand)).unwrap();
```
npmccallum, line 964:
```suggestion
            builder.push(&Entry::Random([0u8; 16])).unwrap();
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+
+            let _handle = builder.done().unwrap();
```
npmccallum, line 966:
```suggestion
            let handle = builder.done().unwrap();
            assert_eq!(handle.start_ptr() as usize % 32, 0);
```

connorkuehl: Following this suggestion: I think this would require the removal of the `#[should_panic]` attribute at the top of this unit test.

npmccallum: @connorkuehl Correct. I thought I had added a suggestion to that effect, but it didn't make it in the review for some reason.

jyotsna-penumaka: Thank you @npmccallum and @connorkuehl for the detailed suggestions.

@npmccallum, Lily and I  were discussing the other day about the pointer returned by start_ptr() function. 
`handle.start_ptr()` is returning a &(), resulting in an error when I cast it to usize.
I tried using `*(&handle.0[handle.1])` instead, which is always returning 1. 

@npmccallum and @connorkuehl, since we are misaligning the stack, `handle.start_ptr() as usize % 32` would never be 0. Our test would be considered a success only if it panics right?
Shouldn't we have  `#[should_panic]` at the top?

npmccallum: > @npmccallum, Lily and I were discussing the other day about the pointer returned by start_ptr() function.
> `handle.start_ptr()` is returning a &(), resulting in an error when I cast it to usize.
> I tried using `*(&handle.0[handle.1])` instead, which is always returning 1.

You probably just need to cast it to a pointer first. So: `handle.start_ptr() as *() as usize`

> @npmccallum and @connorkuehl, since we are misaligning the stack, `handle.start_ptr() as usize % 32` would never be 0.

This about this process as inputs and output. The inputs are:
1. a buffer of memory that will be used as the crt0 stack
2. arguments
3. environment variables
4. auxiliary vectors

The process writes inputs 2-4 into the buffer of memory (1) and returns a pointer to the "bottom" of the data written.

What we want to ensure is that no matter how the buffer of memory (1) is aligned, the output pointer is always correctly aligned.

> Our test would be considered a success only if it panics right?
> Shouldn't we have `#[should_panic]` at the top?

`assert_eq!(handle.start_ptr() as *() as usize % 32, 0);` verifies that the stack is properly aligned and panics otherwise. Therefore, the test will not panic when it is successful.

As a note: I think the actual stack alignment that we need is 16 bytes. But We're starting at 32 bytes to make sure the test fails and inspect the results.

jyotsna-penumaka: Thank you @npmccallum
`(handle.start_ptr() as *const ()` worked.
Do you think the correct stack alignment is 8 instead?
`(handle.start_ptr() as *const () as usize) % 32` always returns 8.
When I changed ` #[repr(C, align(32))]` --> ` #[repr(C, align(8))]` and 
`handle.start_ptr() as *const () as usize) % 32,` --> `handle.start_ptr() as *const () as usize) % 8`, the test passes.

npmccallum: 8 does not seem correct to me. @haraldh What are your thoughts here?

I'm rather surprised that it always returns 8.

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,40 @@ mod tests {
...
         assert_eq!(reader.next(), None); // don't overrun
     }
+
+    #[test]
+    fn crt0_stack_alignment() {
+        // Prepare the crt0 stack.
+        #[repr(C, align(32))]
+        struct Aligned<T>(T);
+
+        let mut aligned = Aligned([0u8; 1024]);
```
npmccallum, line 935:
```suggestion
        let mut aligned = Aligned([0u8; 1024]);
        let mut error = false;
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,40 @@ mod tests {
...
+            builder.push(&Entry::PageSize(4096)).unwrap();
+            builder.push(&Entry::Secure(false)).unwrap();
+            builder.push(&Entry::ClockTick(100)).unwrap();
+            builder.push(&Entry::Flags(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap2(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::Random([0u8; 16])).unwrap();
+
+            let handle = builder.done().unwrap();
+            assert_eq!((handle.start_ptr() as *const () as usize) % 32, 0);
```
npmccallum, line 961:
```suggestion
            let handle = builder.done().unwrap();
            let alignment = (handle.start_ptr() as *const () as usize) % 32;
            eprintln!("offset: {}, alignment: {}", i, alignment);
            error |= alignment != 0;
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,40 @@ mod tests {
...
+        }
```
npmccallum, line 962:
```suggestion
        }
        
        assert!(!error);
```

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+            builder.push(&Entry::PageSize(4096)).unwrap();
+            builder.push(&Entry::Secure(false)).unwrap();
+            builder.push(&Entry::ClockTick(100)).unwrap();
+            builder.push(&Entry::Flags(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::HwCap2(0)).unwrap(); // TODO: https://github.com/enarx/enarx/issues/386
+            builder.push(&Entry::Random([0u8; 16])).unwrap();
+
+            let handle = builder.done().unwrap();
+            let alignment = (handle.start_ptr() as *const () as usize) % 32;
```
haraldh, line 962: should be `STACK_ALIGNMENT` right now

crt0stack/src/lib.rs
```diff
@@ -925,4 +925,45 @@ mod tests {
...
+    fn crt0_stack_alignment() {
+        // Prepare the crt0 stack.
+        #[repr(C, align(32))]
+        struct Aligned<T>(T);
+
+        let mut aligned = Aligned([0u8; 1024]);
+        let mut error = false;
+        for i in 0..32 {
+            let mut builder = Builder::new(&mut aligned.0[i..]);
+            builder.push("/init").unwrap();
```
haraldh, line 939: As the number of arguments also plays a role, one can add some variation here by
```rust
 builder.push("arg").unwrap();
```
in a variable number.

jyotsna-penumaka:
Thank you @haraldh and @npmccallum
I made the requested changes to my code.

Issue #385, opened 2020-03-30
Investigate crt0stack GPF


npmccallum:
Currently, unless we manually align the crt0stack array to 32 bytes, we get a GPF in SGX when attempting to add `Entry::Random` to the stack. We need to debug this issue because it might be a bigger problem. For now, I'm aligning the crt0stack to 32 bytes.

PR #616, opened 2020-06-16
Fix a crasher

npmccallum:
The stack was properly aligned. However, Rust presumed that the stack
was aligned before the call instruction was issued. Because we were
using jmp instead of call, this meant that the output from Rust made
incorrect assumptions about the stack alignment, leading to crashes
later. We solve this by using call rather than jmp so that the Rust
assumptions are correct. This removes the need for a manually aligned
crt0stack which worked around this issue.

There are still some other issues in the crt0stack that need to be
resolved (Harald has fixes for them). However, they don't currently
trigger crashes.

Fixes: #385

### DIFF
--- a/enarx-keep-sgx-shim/src/entry.rs
+++ b/enarx-keep-sgx-shim/src/entry.rs
@@ -10,14 +10,6 @@ extern "C" {
     fn exit(code: u8) -> !;
 }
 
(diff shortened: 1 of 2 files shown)


haraldh:
no credit? :-(

npmccallum:
@haraldh Sorry! Thanks to @haraldh for pointing out this might be the issue.

PR #617, opened 2020-06-16
removed the Vendor type and added a test to verify the values in the Author struct

jyotsna-penumaka:


npmccallum:
@jyotsna-penumaka It looks like some tests don't pass. You'll need to make sure all tests pass before we review it.

PR #618, opened 2020-06-17
Stack alignment, safe guards and small stacks

haraldh:
Make use of `slice::copy_within()` to support small stack sizes.

Guard against OutOfSpace situations with checked integer arithmetic.

Align the final stack to `const STACK_ALIGNMENT = 16`.

### DIFF
--- a/crt0stack/src/lib.rs
+++ b/crt0stack/src/lib.rs
@@ -32,12 +32,15 @@
 #![cfg_attr(not(test), no_std)]
 #![deny(missing_docs)]
 #![deny(clippy::all)]
+#![deny(clippy::integer_arithmetic)]
 
 pub mod auxv;
 pub use auxv::Entry;
 
 use auxv::Key;
 use core::marker::PhantomData;
+use core::mem::{self, size_of};
+const STACK_ALIGNMENT: usize = 16;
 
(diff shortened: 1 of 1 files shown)


npmccallum:
@jyotsna-penumaka When this is merged, you should rebase your test on top of it.

crt0stack/src/lib.rs
```diff
@@ -300,24 +318,49 @@ impl<'a> Builder<'a, Aux> {
...
+            let align_offset = (&self.stack[dst_start_idx] as *const u8 as usize) % STACK_ALIGNMENT;
+
+            let dst_start_idx = dst_start_idx.checked_sub(align_offset).ok_or(OutOfSpace)?;
+
+            // Align the source start index
+            #[allow(clippy::integer_arithmetic)]
+            let src_start_idx = self.items % core::mem::size_of::<usize>();
+
+            self.stack
+                .copy_within(src_start_idx..self.items, dst_start_idx);
```
haraldh, line 354: Strange.. this has a line break, in contrast to line 345, which is longer!

haraldh:
> This mostly looks good. I'm not a fan of all the wrapping from `rustfmt`, I think it makes it harder to read. Could you split up some of those into separate lines with less wrapping?

done

PR #621, opened 2020-06-18
Fix SGX syscalls readv/writev

npmccallum:


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -418,13 +418,15 @@ impl<'a> Handler<'a> {
 
             let mut offset = 0;
             for (t, u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-                if u.iov_base != ubuffer[offset..].as_mut_ptr() as *mut _ || u.iov_len != t.iov_len
-                {
+                let us = &mut ubuffer[offset..][..t.iov_len];
```
lkatalin, line 421: This makes sense... the offset is updated each time and used as start pointer, and length of the iovec is the end pointer.

Issue #469, opened 2020-04-28
Improve logging facilities in the SGX shim


npmccallum:
Currently we have a trait `Print` which has minimal functionality. It would be better if we could do something like `impl core::fmt::Write for Handler`, that way we could use the `write!()`-family of macros. However, attempts to do this revealed problems that we didn't have time to debug. See: https://github.com/enarx/enarx/pull/408#discussion_r410233879

PR #623, opened 2020-06-18
Implement core::fmt::Write for Handler

npmccallum:
This lets us dump the janky Print trait that I made. This is more Rust ergonomic.

Fixes: #469

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -5,11 +5,13 @@ use crate::Layout;
 use sgx_types::ssa::StateSaveArea;
 use span::{Contains, Line, Span};
 
+use core::fmt::Write;
 use core::mem::size_of;
 use core::ptr::copy_nonoverlapping;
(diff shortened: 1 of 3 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -51,20 +49,23 @@ pub struct Handler<'a> {
...
         let map = match self.ualloc(len) {
-            Err(_) => return,
+            Err(_) => return Err(core::fmt::Error),
             Ok(map) => map,
         };
 
         unsafe {
             // Copy the encrypted input into unencrypted memory.
-            copy_nonoverlapping(bytes.as_ptr(), map, bytes.len());
+            core::ptr::copy_nonoverlapping(bytes.as_ptr(), map, bytes.len());
```
lkatalin, line 68: Why did this prefix come back?

npmccallum: I think I just missed one.

PR #629, opened 2020-06-23
Use cargo-make global envs as root dir instead of git root directory

### LINKED ISSUE #599: Use builtin cargo-make environment variable for finding the workspace root
Right now we use [`git` to deduce where the root of the workspace is](https://github.com/enarx/enarx/blob/master/Makefile.toml#L29) to construct absolute paths to some of our CI scripts.

However, this is made redundant by `cargo-make`'s builtin environment variable that already does this:

> `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` - The original working directory of the workspace. Enables workspace members access to the workspace level `CARGO_MAKE_WORKING_DIRECTORY`.

Furthermore, our current approach has a problem because it depends on `git`. This isn't an issue for the common case, but GitHub does allow people to download a copy of the repo _without_ any of the `.git/` data. This dependency on `git` will prevent people from building the Enarx repo if they have downloaded a copy of the source without `git`.

```bash
$ curl -LOk https://github.com/enarx/enarx/archive/master.zip
$ cd enarx-master
$ cargo make
[cargo-make] INFO - cargo make 0.30.6
[cargo-make] ERROR - Error while executing command, exit code: 128
[cargo-make] WARN - Build Failed.
$ git rev-parse --show-toplevel
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
$ echo $?
128
```

Switch our `Makefile.toml` to use the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` to replace `CARGO_WORKSPACE_ROOT`.

ziyi-yan:
`git` command and `.git` directory will not be required for building enarx.

Resolves #599

Signed-off-by: Ziyi Yan <ziyi.yan@foxmail.com>

After trying and reading `cargo-make` documentation, I found out that `CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY` is the one we need to replace the original `CARGO_WORKSPACE_ROOT`, which is the directory containing the `Makefile.toml`.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -27,7 +27,6 @@ CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = [
 	"units",
 	"vdso"
 ]
-CARGO_WORKSPACE_ROOT = { script = ["git rev-parse --show-toplevel"] }
 
 [tasks.deny]
 install_crate = "cargo-deny"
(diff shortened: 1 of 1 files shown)


npmccallum:
@ziyi-yan Thanks for the contribution! @connorkuehl will take it from here.

connorkuehl:
Hi @ziyi-yan! Thanks for the pull request! I wanted to reply to [your comment](https://github.com/enarx/enarx/issues/599#issuecomment-648035016) on #599 here:

> I tried out use CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY for replacement. But it turns out that it refers to the root directory of each member, instead of the root directory of the project. I think it's because there is no real workspace defined in the root directory. It is an emulated workspace by the option CARGO_MAKE_WORKSPACE_EMULATION from cargo-make. So I use CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY instead and it works.

I think that you have found a bug in cargo-make's workspace emulation feature and so I [filed a bug upstream](https://github.com/sagiegurari/cargo-make/issues/434). I'd like to wait to approve this pull request until we hear back from Sagie regarding this behavior so we don't have to come back and change it from CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY to CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY in the event a fix is pushed upsream.

npmccallum:
@connorkuehl It looks like this was fixed upstream. Do we still need to merge this PR?

ziyi-yan:
Got it replaced and it does pass the CI. 🥳 @connorkuehl PTAL

PR #632, opened 2020-06-24
Protocol work

MikeCamel:
**OLD - IGNORE - see comment from 2020-07-03.**
This is a first draft, and there are a number of questions to be answered: it may be that @npmccallum already has answers for some of them. 

Two major outstanding questions are 
a) around the establishment and use of the session-key; and 
b) authentication of the Tenant to the Keep.

The "key-material" sent to the Tenant from the CPU/firmware via Keep Manager is not confidential, as the Keep Manager is untrusted.  We can be sure of its integrity, as it is part of the signed attestation bundle from the CPU/firmware.  If it is a public key, generated by the Keep, it can be used as part of a Diffie-Hellman key exchange, but otherwise cannot be used as a key itself (as it is not confidential, and could be spoofed by a party in league with the Keep Manager).  On the other hand, as long as the communications between the Keep Loader and the Tenant are confidentiality-protected, the key material can be used as authentication by the Tenant to the KeepLoader, as Keep Manager is part of the host's domain, and there is no trust benefit to the host in allowing other Tenants to use the Keep (at least in the default trust model).  There are possible charging implications, however, if the Keep Manager creates a Keep and then lets another tenant use it.  One way to address this is for the Tenant to send the Keep Manager a copy of its (the Tenant's) public key, for this to be loaded into the Keep with the Keep runtime image, and for the KeepLoader to perform a challenge-response (with a Nonce, to avoid replay attacks) before loading a workload, or, simpler, to use this as client authentication in the DH exchange discussed below.

Assuming that sending a self-signed cert from the Keep to the CPU/firmware as the key material is implausible (given amount of data), these seem to be the options to allow the Tenant to communicate with the Keeploader:
1. plain socket with in-band DH to establish a session key which is used to encrypt specific data.  Downside: requires hand-crafted crypto handshake and protocol.
2. TLS socket.  This requires set-up, and we have a public key, not a certificate.  UNKNOWN: are there easily-available TLS modes to accept public keys, not certs?  Downside: requires hand-crafted protocol.
3. HTTP with in-band DH to establish a session key which is used to encrypt specific data.  Downside: requires hand-crafted crypto handshake.
4. HTTPS.  This is tricky because the standard http servers require a cert at start-time, so performing the DH in-band becomes difficult.
5. plain socket with in-band DH to establish a cert, followed by HTTP, using that cert.  Downside: requires hand-crafted crypto handshake, connection drop and reestablishment.

If we can place a Tenant-signed cert into the Keep, and send a Keep-signed cert to the CPU/firmware, HTTPS with client authentication becomes a very plausible option.~~
OLD-IGNORE


MikeCamel:
Realised I'd made some mistakes.  Hopefully fixed.

axelsimon:
Here's a copy of the diagram as a viewable png file, as of currently [latest commit](https://github.com/enarx/enarx/pull/632/commits/afc906d129b74178fc03de1751a5bdda11b200af):
protocol

MikeCamel:
A further comment: the final method, "Workload deletion", requires the use another thread, as it assumes that the AppLoader is waiting for connections on a port.  Given the cost of thread - and their availability at all in some cases - this final method is likely to be left unimplemented in early versions.

MikeCamel:
This new version is a major re-write after a conversation with @npmccallum and @ambaxter.  I hope I've remembered all the pieces correctly!

DH key exchange is now part of the attestation, and binding to a URI comes much later: initial conversations from the Tenant to the Keep are proxied via the Keepmanager on a local UNIX socket.

I've also separated out some of the steps to manage state better (and also to prepare for "pre-warmed" Keeps).

MikeCamel:
protocol

lkatalin:
I have a lot of questions about this protocol diagram. TBH I think the diagram needs a narrative explanation to accompany it because I'm not sure what all of the data going back and forth is doing. Some of the actions here don't seem to relate to anything I've heard about in our architecture or the RFCs I've read.

Do these steps all happen for both SEV and SGX (and whatever other technology)? Could there be different-colored lines showing steps that are only for a specific technology?

What role would contracts have in an SGX Keep (if any)?

What is `T-pub`?

What is an attestation reset?

How does the challenge / response differ from the attestation (which at least in part appears to be happening locally on the host)?

Etc.

MikeCamel:
This work has been overtaken by events, and therefore superseded.  We will create a new version separately.

Issue #627, opened 2020-06-19
Test SGX syscall handling


lkatalin:
We need to test that the [syscalls](https://github.com/enarx/enarx/blob/master/enarx-keep-sgx-shim/src/event.rs#L43) we've implemented in the Handler are working correctly. These tests should run as top-level integration tests. An example of this type of test is our [SGX payload test](https://github.com/enarx/enarx/blob/master/integration-tests/tests/sgx_payload.rs).

At a high level, the following should be done for each syscall in the Handler:
- [ ] Using `main.rs` in the `payload` crate as an example, create a payload that will make the syscall being tested. This file can go somewhere in the `integration-tests` crate. (Alternatively we could put all our payloads in the `payload` crate and be more specific about which one we're calling in the tests and demos.)
- [ ] Create a file in the `integration-tests` crate that will call the payload created above. This will be much the same as the SGX payload test except for [this line](https://github.com/enarx/enarx/blob/master/integration-tests/tests/sgx_payload.rs#L45), which will contain the new payload file, and the [earlier](https://github.com/enarx/enarx/blob/master/integration-tests/tests/sgx_payload.rs#L33) [lines](https://github.com/enarx/enarx/blob/master/integration-tests/tests/sgx_payload.rs#L38) related to finding that payload.

Issue #609, opened 2020-06-09
Kill payload integration test after a period of time


connorkuehl:
The SGX shim hangs when it panics. Our integration tests should kill the SGX test if it takes an unreasonable amount of time to complete. We can hand-tune the timer, a safe bet is probably 30s-60s right now.

This is less of an issue on local developer machines, but will be useful since we will run PRs on CI machines.

connorkuehl:
Closed by #707

PR #633, opened 2020-06-24
Add a test for exit system call

jyotsna-penumaka:
Related: #627 and #609

### DIFF
--- a/integration-tests/src/bin/exit_one.rs
+++ b/integration-tests/src/bin/exit_one.rs
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::exit;
+
+fn main() {
+    unsafe {
+        exit(1);
+    }
+}
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains the common code for integration tests
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+use std::io::{Error, ErrorKind};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::time::Duration;
+use wait_timeout::ChildExt;
+use walkdir::WalkDir;
+
+pub fn keep_type() -> &'static str {
+    #[cfg(has_sgx)]
+    return "enarx-keep-sgx";
+
+    #[cfg(has_sev)]
+    return "enarx-keep-sev";
+
+    #[cfg(not(any(has_sgx, has_sev)))]
+    compile_error!("Need either SGX or SEV!");
+}
+
+pub struct IntegrationTest {
+    pub wksp_root: PathBuf,
+    pub keep: PathBuf,
+    pub payload: PathBuf,
+    pub shim: PathBuf,
+}
+
+impl IntegrationTest {
+    pub fn new(bin: &str) -> Self {
+        let enarx_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+        let enarx_dir = enarx_dir.parent().unwrap();
+        let keep_type = keep_type();
+
+        let wksp_root = enarx_dir.to_path_buf();
+        let keep_dir = enarx_dir
+            .join(format!("{}/target/", keep_type))
+            .to_path_buf();
+        let payload_dir = enarx_dir
+            .join("integration-tests/target/x86_64-unknown-linux-musl/")
+            .to_path_buf();
+        let shim_dir = enarx_dir
+            .join(format!(
+                "{}-shim/target/x86_64-unknown-linux-musl/",
+                keep_type
+            ))
+            .to_path_buf();
+        let shim_filename = format!("{}-shim", keep_type);
+        let shim_filename = shim_filename.as_str();
+
+        let keep = find_filepath(keep_dir, keep_type).unwrap();
+        let payload = find_filepath(payload_dir, bin).unwrap();
+        let shim = find_filepath(shim_dir, shim_filename).unwrap();
+
+        Self {
+            wksp_root,
+            keep,
+            payload,
+            shim,
+        }
+    }
+
+    pub fn run(self: Self, timeout: u64, exit_status: i32) -> () {
+        let seconds = Duration::from_secs(timeout);
+
+        let mut cmd = Command::new(self.keep)
+            .current_dir(self.wksp_root)
+            .arg("--code")
+            .arg(self.payload)
+            .arg("--shim")
+            .arg(self.shim)
+            .spawn()
+            .expect("failed to run the test");
+
+        let ecode = match cmd.wait_timeout(seconds).unwrap() {
+            Some(status) => status.code(),
+            None => {
+                cmd.kill().unwrap();
+                panic!("killed by watchdog!");
+            }
+        };
+
+        assert_eq!(exit_status, ecode.unwrap());
+    }
+}
+
+/// Finds an absolute file path for a file in a directory.
+#[cfg_attr(not(has_sgx), ignore)]
+pub fn find_filepath(dir: impl AsRef<Path>, name: &str) -> Result<PathBuf, Error> {
+    for entry in WalkDir::new(&dir).into_iter().filter_map(|e| e.ok()) {
+        if entry.path().file_name().unwrap() == name {
+            return Ok(entry.path().to_path_buf());
+        }
+    }
+    Err(Error::new(ErrorKind::Other, "file not found"))
+}
--- a/integration-tests/.cargo/config
+++ b/integration-tests/.cargo/config
@@ -1,0 +1,6 @@
+[build]
+target = "x86_64-unknown-linux-musl"
+rustflags = [
+    "-C", "relocation-model=pic",
+    "-C", "linker=../cc",
+]
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -8,3 +8,6 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+libc = { version = "0.2.0", features = [] }
+wait-timeout = "0.2.0"
+walkdir = "2"
--- a/integration-tests/link.json
+++ b/integration-tests/link.json
@@ -1,0 +1,18 @@
+{
+    "build": {
+        "prepend": [
+            "-static-pie"
+        ],
+
+        "replace": {
+            ".*/crt1.o": ["/usr/x86_64-linux-musl/lib64/rcrt1.o"],
+            "-no-pie": []
+        },
+
+        "debug": false
+    },
+
+    "test": {
+        "debug": false
+    }
+}
--- a/integration-tests/src/bin/exit_zero.rs
+++ b/integration-tests/src/bin/exit_zero.rs
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::exit;
+
+fn main() {
+    unsafe {
+        exit(0);
+    }
+}
--- a/integration-tests/src/bin/watchdog.rs
+++ b/integration-tests/src/bin/watchdog.rs
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    loop {}
+    unreachable!();
+}
--- a/integration-tests/tests/syscall_exit.rs
+++ b/integration-tests/tests/syscall_exit.rs
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains integration tests, including the exit syscall payload test.
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the exit syscall payload in the SGX keep using the SGX shim.
+
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn exit_zero() {
+    IntegrationTest::new("exit_zero").run(5, 0);
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn watchdog() {
+    IntegrationTest::new("watchdog").run(5, 0);
+}
+
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn exit_one() {
+    IntegrationTest::new("exit_one").run(5, 1);
+}


npmccallum:
@jyotsna-penumaka If you rebase and force push, the CI error will go away. It has been fixed on master.

npmccallum:
@jyotsna-penumaka I was able to fix the last test blocker. All tests now pass. Please squash your commits into a logical series and we can proceed with review.

jyotsna-penumaka:
Thanks @npmccallum
I am trying to add code to kill the test in case we don't get an exit code.
I will squash all my commit as soon as I am done.

jyotsna-penumaka:
@npmccallum and @connorkuehl 

Thanks for the suggestions @connorkuehl  :)
I made few changes to my code.
In order to resolve code duplication I now have a common/mod.rs that contains a struct, which initialises the required binary paths. Before each syscall test, I just instantiate the struct. 
Like @connorkuehl suggested, I kill the child process if it takes more than 60 secs, therefore we will always have an exit code. 
I moved the binaries to the integration-tests crate, I also copied the linker script and a .cargo config file from the payload crate to the integration-tests crate.
I think my naming scheme is still a little confusing.
I plan to have a separate file for each syscall, so that we would be able to specify the name of the syscall to test.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+            shim: find_filepath(
+                (PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().unwrap())
+                    .join("enarx-keep-sgx-shim/target/x86_64-unknown-linux-musl/")
+                    .to_path_buf(),
+                "enarx-keep-sgx-shim",
+            )
+            .unwrap(),
+        }
+    }
+}
```
connorkuehl, line 44: This solution does improve code reuse, but we can take it a couple of steps further.

1. We can more seamlessly generate code that works for all of our keeps (not just sgx) using conditional compilation. I'm thinking of a free-standing function like this:

```rust
#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
pub const fn keep_under_test() -> &'static str {
    #[cfg(feature = "sgx")]
    "enarx-keep-sgx"
    
    #[cfg(feature = "sev")]
    "enarx-keep-sev"
}
```

can help us fill in some of the setup code in a more flexible way without having to change any of our code:

```rust
            keep: find_filepath(
                (PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().unwrap())
                    .join(format!("{}/target/", keep_under_test())
                    .to_path_buf(),
                keep_under_test(),
            )
```

furthermore, since we've been consistently naming our crates `enarx-keep-blah` and `enarx-keep-blah-shim` this would let us do:

```rust
            shim: find_filepath(
                (PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().unwrap())
                    .join(format!("{}-shim/target/x86_64-unknown-linux-musl/", keep_under_test()))
                    .to_path_buf(),
                format!("{}-shim", keep_under_test()),
            )
```

There are lots of calls to `PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().unwrap()`, I'd suggest factoring them all out and storing this as a variable at the beginning of the function.

As far as selecting the payload directory:

```rust             
payload_dir: (PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().unwrap())
                .join("integration-tests/target/x86_64-unknown-linux-musl/")
                .to_path_buf(),
```

I think this is an artifact from these binaries living in the `payload` crate in a previous revision. This could probably be closer to

```rust
payload_dir: PathBuf::from(env!("CARGO_MANIFEST_DIR"))
                .join("target/x86_64-unknown-linux-musl/")
                .to_path_buf(),
```

After all these changes, the `Setup` struct is beginning to look like a constructor function for a `Command`, which we can use to avoid having to instantiate a `Command` struct by hand in each test.

I think `struct Setup` should be changed to `struct IntegrationTest(Command)`.

Then its constructor function could end with something like:

```rust
pub fn new(bin: &str) -> Self {
    // snipped setup
    // "bin" is the name of the payload we find during the setup stage,
    // ultimately its full filepath should wind up in the `code` variable referred to below

    let cmd = Command::new(keep)
        .current_dir(wksp_root)
        .arg("--shim")
        .arg(shim)
        .arg("--code")
        .arg(code)

    Self(cmd)
}
```

Then it can offer a method called something like:

```rust
pub fn exec(&self, timeout: Option<Duration>) -> ExitStatus {
    let timeout = time.unwrap_or(Duration::from_secs(60));
    let status = self.0.spawn().unwrap();
    match status.wait_timeout(timeout).unwrap() {
        None => {
            self.0.kill.unwrap();
            panic!("killed by watchdog");
        },
        Some(exit_status) => exit_status
       }
    }
}
```

So after all that, ideally running an integration test should be as easy as:

```rust
#[test]
fn test_sys_exit() {
    let test = IntegrationTest::new("test_sys_exit");
    assert!(test.exec(None).success());
}
```

Regardless of what keep we are trying to test. I think the above suggestions should accommodate most of our integration testing needs for now.

Also just a disclaimer that I haven't tried compiling _any_ of my code snippets above; I'm just thinking out loud. So if you decide to incorporate some of the ideas above, some elbow grease might be needed to keep them to compile :slightly_smiling_face:

integration-tests/tests/exit_syscall_payload.rs
```diff
@@ -0,0 +1,68 @@
...
+    let mut payload_test = Command::new(setup.keep)
+        .current_dir(setup.wksp_root)
+        .arg("--code")
+        .arg(payload)
+        .arg("--shim")
+        .arg(setup.shim)
+        .spawn()
+        .expect("failed to run exit payload test");
+
+    let ecode = match payload_test.wait_timeout(Duration::from_secs(60)).unwrap() {
```
connorkuehl, line 60: Also, for testing the watchdog/autokill functionality, I think a duration of 1-5 seconds is probably sufficient. Give the keep enough time to setup and enter the loop for a little bit. This would hold up our CI for another entire minute.

npmccallum:
@connorkuehl @jyotsna-penumaka Just thinking out loud here. I'm thinking we probably need some kind of test harness something like this:

```
fn run_test(payload: &str, timeout: usize, exit_status: u8, stdin: &str, stdout: &str, stderr: &str);
```

We should be able to declaratively say that under a given payload these are all the invariants we expect. This should work for all keeps/shims but with different payloads.

connorkuehl:
> @connorkuehl @jyotsna-penumaka Just thinking out loud here. I'm thinking we probably need some kind of test harness something like this:
> 
> ```
> fn run_test(payload: &str, timeout: usize, exit_status: u8, stdin: &str, stdout: &str, stderr: &str);
> ```
> 
> We should be able to declaratively say that under a given payload these are all the invariants we expect. This should work for all keeps/shims but with different payloads.

Agreed. I like this better than my earlier ramblings. I don't think we're far off from that as long as we keep a handle to the [Child](https://doc.rust-lang.org/std/process/struct.Child.html) around for messing with stdin/stdout/stderr.

jyotsna-penumaka:
Thank you @connorkuehl and @npmccallum 
I have made changes to my code by following your suggestions, they were very helpful.
I didn't make use of stdin, stdout or stderr in the run_test function since we are only testing exit syscall at the moment.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+            .wait_timeout(Duration::from_secs(num_sec as u64))
+            .unwrap()
+        {
+            Some(status) => status.code(),
+            None => {
+                cmd.kill().unwrap();
+                panic!("killed by watchdog!");
+            }
+        };
+        assert_eq!(exit_status, ecode.unwrap());
```
npmccallum, line 99:
```suggestion
        let seconds = Duration::from_secs(timeout.unwrap_or(5) as u64);
        let exit_status = exit_status.unwrap_or(0);

        let mut cmd = Command::new(self.keep)
            .current_dir(self.wksp_root)
            .arg("--code")
            .arg(self.payload)
            .arg("--shim")
            .arg(self.shim)
            .spawn()
            .expect("failed to run the test");
        
        let ecode = match cmd.wait_timeout(seconds).unwrap() {
            Some(status) => status.code(),
            None => {
                cmd.kill().unwrap();
                panic!("killed by watchdog!");
            }
        };

        assert_eq!(exit_status, ecode.unwrap());
```

1. I used vertical whitespace to group related blocks of code.
2. I replaced `match` with `.unwrap_or()` where applicable.
3. I moved `Duration::from_secs()` earlier, which removed unnecessary wrapping.
4. I moved input argument processing to the top of the function.

It still might be possible to simplify the remaining match statement if I looked more closely. But it may not be worth it.

Notice that the groups of code are now:
1. Process input arguments.
2. Prepare the command for execution.
3. Execute the command.
4. Evaluate the results.

This should be very easy to read now.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+/// Finds an absolute file path for a file in a directory.
+#[cfg_attr(not(has_sgx), ignore)]
+pub fn find_filepath(dir: PathBuf, name: &str) -> Result<PathBuf, std::io::Error> {
```
npmccallum, line 105:
```suggestion
pub fn find_filepath(dir: impl AsRef<Path>, name: &str) -> Result<PathBuf, std::io::Error> {
```

You can see the same strategy [here](https://doc.rust-lang.org/std/fs/fn.read.html).

Basically, any time you're going to take a path as an input, you want to instead take "any type that can be expressed as a `Path`." This allows you to use things like string literals.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+    pub fn run_test(
+        self: Self,
+        timeout: Option<usize>,
+        exit_status: Option<i32>,
+        _stdin: Option<&str>,
+        _stdout: Option<&str>,
+        _stderr: Option<&str>,
+    ) -> () {
```
npmccallum, line 72:
```suggestion
    }

    pub fn run_test(
        self: Self,
        timeout: usize,
        exit_status: i32,
    ) -> () {
```

1. Vertical space between functions.
2. Remove unused parameters. Just add them later.
3. Require all arguments.

In particular, we will *always* test for all the test criteria. So force the caller to think carefully about sensible values.

integration-tests/tests/common/mod.rs
npmccallum: Note that I incuded `.unwrap_r()` even though it isn't needed anymore. I just wanted to let you know that it exists.

integration-tests/tests/exit_syscall_payload.rs
```diff
@@ -0,0 +1,24 @@
...
+    test.run_test(Some(5), Some(0), None, None, None);
+}
+
+#[cfg_attr(not(has_sgx), ignore)]
+#[test]
+#[should_panic]
+fn timeout_sys_exit() {
+    let test = common::IntegrationTest::new("test_watchdog");
+    test.run_test(Some(5), Some(0), None, None, None);
+}
```
npmccallum, line 24: We currently have a `watchdog` and `exit_zero` test. We need an `exit_one` test. This is trivial to add, but verifies that changing the input to `exit()` actually has an effect.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        .unwrap();
+        Self {
+            wksp_root: local_wksp_root,
+            keep: local_keep,
+            payload: local_payload,
+            shim: local_shim,
+        }
+    }
+
+    pub fn run_test(self: Self, timeout: usize, exit_status: i32) -> () {
```
npmccallum, line 70:
```suggestion
    pub fn run(self: Self, timeout: usize, exit_status: i32) -> () {
```

Since the type is `IntegrationTest` we know it is a test. Hence `run_test()` is redundant.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    let file = Command::new("find")
+        .current_dir(&dir)
+        .arg("-name")
+        .arg(name)
+        .output()
+        .expect("could not find file")
+        .stdout;
+
+    let path = dir.as_ref().join(String::from_utf8(file).unwrap().trim());
+    Ok(std::fs::canonicalize(&path)?)
```
npmccallum, line 105: Use the `walkdir` crate here instead.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        let local_shim = find_filepath(
+            enarx_dir
+                .join(format!(
+                    "{}-shim/target/x86_64-unknown-linux-musl/",
+                    keep_type
+                ))
+                .to_path_buf(),
+            format!("{}-shim", keep_type).as_str(),
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
npmccallum: @jyotsna-penumaka This request got missed in the last round of changes.

integration-tests/src/bin/test_watchdog.rs
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    loop {}
+    unreachable!();
+}
```
npmccallum, line 6: Rename this file: `git mv test_watchdog.rs watchdog.rs`

integration-tests/tests/exit_syscall_payload.rs
```diff
@@ -0,0 +1,30 @@
...
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn exit_zero() {
+    IntegrationTest::new("exit_zero").run(5, 0);
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn watchdog() {
+    IntegrationTest::new("test_watchdog").run(5, 0);
```
npmccallum, line 23:
```suggestion
    IntegrationTest::new("watchdog").run(5, 0);
```

integration-tests/tests/exit_syscall_payload.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
```
npmccallum, line 1: Rename this file: `git mv exit_syscall_payload.rs syscall_exit.rs`

The reason for this is that the name of this file will be output in the test runs. It will be left-justified. This way all the syscalls line up and it is easy to spot which syscall is failing.

integration-tests/src/bin/exit_one.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
```
npmccallum, line 1: Right now, all these binaries will be built and installed during a `cargo install`. We probably don't want that. I think we can stop that from happening by putting `#![cfg(test)]` here.

integration-tests/tests/common/mod.rs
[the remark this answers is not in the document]
npmccallum: Don't forget to fix this.

integration-tests/src/bin/exit_one.rs
npmccallum: After @jyotsna-penumaka tried it, I confirmed this doesn't work. Resolved.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+            .join("integration-tests/target/x86_64-unknown-linux-musl/")
+            .to_path_buf();
+        let shim_dir = enarx_dir
+            .join(format!(
+                "{}-shim/target/x86_64-unknown-linux-musl/",
+                keep_type
+            ))
+            .to_path_buf();
+        let shim_filename = format!("{}-shim", keep_type);
+        let shim_filename = shim_filename.as_str();
```
connorkuehl, line 53: I don't think there's anything technically wrong with this but just for the sake of Rust edification it would be sufficient to just borrow with `&shim_filename` for function signatures that require `&str`.

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+                panic!("killed by watchdog!");
+            }
+        };
+
+        assert_eq!(exit_status, ecode.unwrap());
+    }
+}
+
+/// Finds an absolute file path for a file in a directory.
+#[cfg_attr(not(has_sgx), ignore)]
```
connorkuehl, line 92: This needs to be updated since `has_sgx` is not the only `cfg`.

I wonder if we should build on top of our `has_blah` `cfg`s so that we can more conveniently ignore code without repeating the list of `has_blah` `cfg`s.

I think something like:

```rust
#[cfg_attr(any(has_sgx, has_sev), has_tee)]
```

should work if put in the root of the crate. Either `integration-tests/src/main.rs` or maybe `integration-tests/tests/common/mod.rs`

Then it's more convenient because there are some functions like `keep_type()` that need to query for a specific `has_sgx` or `has_sev`, but there are other blocks of code that don't care _which_ TEE we're running under, just that there is one.

This will be especially nice as once we have even more TEEs we'll just need to remember to add their entries to `build.rs` and that `cfg_attr`

integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        Self {
+            wksp_root,
+            keep,
+            payload,
+            shim,
+        }
+    }
+
+    pub fn run(self: Self, timeout: u64, exit_status: i32) -> () {
```
connorkuehl, line 67: This signature is technically correct, but I think it's a little more idiomatic to present this as:

```rust
pub fn run(self, timeout: u64, exit_status: i32) {
```

PR #634, opened 2020-06-25
enarx-keep-sev-shim: fix const `ALIGN_SECTION` for release

haraldh:
The IDE refactor rename didn't catch the release version of `ALIGN_SECTION`.


haraldh:
And because the CI didn't catch this, I also opened https://github.com/enarx/enarx/issues/635

PR #639, opened 2020-06-26
memory: add conf fn Page::zeroed()

haraldh:
Page::default() is not const


haraldh:
yeah.. `openssl` makes it fail the CI

PR #653, opened 2020-06-30
makefile changes

sagiegurari:
i looked at the workspace emulation usage and saw few things that can be improved in the makefile.

* bool env vars can be just true/false, no need to make them strings
* new env var for test scripts dir location, so make tasks smaller and more readable.
* removed ```install_crate = "cargo-deny"``` as cargo make knows that automatically 
* replaced shell script 
```install_script = [''' which toml || cargo install toml-cli ''']```
with toml
```toml
install_crate = { crate_name = "toml-cli", binary = "toml", test_arg = "--help" }
```
which is multi platform

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -1,5 +1,5 @@
 [env]
-CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = "true"
+CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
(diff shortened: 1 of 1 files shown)


sagiegurari:
Done, sorry making a crappy message there :)
I have tested it on gitpod env, however some tests did fail when i ran the integration-ci:
cc: error: /usr/x86_64-linux-musl/lib64/rcrt1.o: No such file or directory

which is more of an env issue and not related. but just be aware

PR #656, opened 2020-07-01
Make TSS and GDT part of .data rather than allocate on the stack

haraldh:
If the kernel stack later gets changed, we want no allocated objects on the old stack.

### DIFF
--- a/enarx-keep-sev-shim/src/gdt.rs
+++ b/enarx-keep-sev-shim/src/gdt.rs
@@ -15,8 +15,11 @@ use x86_64::structures::gdt::{
 use x86_64::structures::tss::TaskStateSegment;
 use x86_64::{PrivilegeLevel, VirtAddr};
 
-static mut TSS: Option<TaskStateSegment> = None;
-pub(crate) static mut GDT: Option<(GlobalDescriptorTable, Selectors)> = None;
+const STACK_NUM_PAGES: usize = 5;
+static mut STACKS: [[Page; STACK_NUM_PAGES]; 7] = [[Page::zeroed(); STACK_NUM_PAGES]; 7];
+
+pub(crate) static mut TSS: TaskStateSegment = TaskStateSegment::new();
+pub(crate) static mut GDT: GlobalDescriptorTable = GlobalDescriptorTable::new();
 
 /// The user data segment
 ///
@@ -32,24 +35,13 @@ const USER_DATA_SEGMENT: u64 = (USER_DATA_SEGMENT_INDEX << 3) | (PrivilegeLevel:
 const USER_CODE_SEGMENT_INDEX: u64 = 4;
 const USER_CODE_SEGMENT: u64 = (USER_CODE_SEGMENT_INDEX << 3) | (PrivilegeLevel::Ring3 as u64);
 
-const STACK_NUM_PAGES: usize = 5;
-
-/// The selectors used for the GDT
-pub struct Selectors {
-    /// The code selector of the shim
-    pub code_selector: SegmentSelector,
-    /// The data selector of the shim
-    pub data_selector: SegmentSelector,
-    /// The data selector of the payload
-    pub user_data_selector: SegmentSelector,
-    /// The code selector of the payload
-    pub user_code_selector: SegmentSelector,
-    /// The tss selector
-    pub tss_selector: SegmentSelector,
-}
-
 /// Initialize the GDT
-pub fn init() {
+///
+/// # Safety
+///
+/// `unsafe` because the caller has to ensure it is only called once
+/// and in a single-threaded context.
+pub unsafe fn init(level_0_stack: VirtAddr) {
     #[cfg(debug_assertions)]
     eprintln!("init_gdt");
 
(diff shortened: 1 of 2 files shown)


enarx-keep-sev-shim/src/main.rs
```diff
@@ -52,11 +53,18 @@ macro_rules! entry_point {
 
 entry_point!(shim_main);
 
+/// FIXME: will be replaced by a dynamically allocated stack with safe guards
+pub static mut LEVEL_0_STACK: [Page; 5] = [Page::zeroed(); 5];
```
npmccallum, line 57: Should `STACK_NUM_PAGES` be used here?

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -15,8 +15,11 @@ use x86_64::structures::gdt::{
 use x86_64::structures::tss::TaskStateSegment;
 use x86_64::{PrivilegeLevel, VirtAddr};
 
-static mut TSS: Option<TaskStateSegment> = None;
-pub(crate) static mut GDT: Option<(GlobalDescriptorTable, Selectors)> = None;
+const STACK_NUM_PAGES: usize = 5;
+static mut STACKS: [[Page; STACK_NUM_PAGES]; 7] = [[Page::zeroed(); STACK_NUM_PAGES]; 7];
```
npmccallum, line 19:
```suggestion
static mut STACKS: [[Page; STACK_NUM_PAGES]; 6] = [[Page::zeroed(); STACK_NUM_PAGES]; 6];
```

?

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -59,93 +51,59 @@ pub fn init() {
...
 
-            // The stacks for the exceptions and interrupts
-            for i in 0..7 {
-                tss.interrupt_stack_table[i] =
-                    stack_end(VirtAddr::from_ptr(&STACKS[i + 1]), STACK_NUM_PAGES);
-            }
-            tss
-        });
+    // The stacks for the exceptions and interrupts
+    for i in 0..7 {
```
npmccallum, line 57: It might be cleaner to use zip on iterators here.

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -59,93 +51,59 @@ pub fn init() {
...
-        SFMask::write(RFlags::INTERRUPT_FLAG | RFlags::TRAP_FLAG);
-
-        // Set the kernel gs base to the TSS to be used in `_syscall_enter`
-        KernelGsBase::write(VirtAddr::new(TSS.as_ref().unwrap() as *const _ as u64));
-    }
+    // `syscall` loads segments from STAR MSR assuming a data_segment follows `kernel_code_segment`
+    // so the ordering is crucial here. Star::write() will panic otherwise later.
+    let code_selector = GDT.add_entry(Descriptor::kernel_code_segment());
+    let flags =
+        DescriptorFlags::USER_SEGMENT | DescriptorFlags::PRESENT | DescriptorFlags::WRITABLE;
```
npmccallum, line 65: Can we avoid this wrap somehow?

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -59,93 +51,59 @@ pub fn init() {
...
+    load_gs(SegmentSelector(0));
+
+    // Set the selectors to be set when userspace uses `syscall`
+    Star::write(
+        user_code_selector,
+        user_data_selector,
+        code_selector,
+        data_selector,
+    )
+    .unwrap();
```
npmccallum, line 99: Can we get rid of this wrap somehow?

enarx-keep-sev-shim/src/gdt.rs
connorkuehl: I don't follow how the number of stacks is chosen here

enarx-keep-sev-shim/src/main.rs
haraldh: No, because this is the main stack and it will be changed in later patches anyway.

enarx-keep-sev-shim/src/gdt.rs
haraldh: > ?

Huh? why 6?

```rust
pub struct TaskStateSegment {
[…]
    pub interrupt_stack_table: [VirtAddr; 7],
[…]
```

enarx-keep-sev-shim/src/gdt.rs
haraldh: ok

enarx-keep-sev-shim/src/gdt.rs
haraldh: yes

enarx-keep-sev-shim/src/gdt.rs
haraldh: yes

enarx-keep-sev-shim/src/gdt.rs
haraldh: I added
```rust
debug_assert_eq!(STACKS.len(), TSS.interrupt_stack_table.len());
```

enarx-keep-sev-shim/src/gdt.rs
haraldh: Because zip() ends, when one if the iterators is short on elements, I added:
```rust
debug_assert_eq!(STACKS.len(), TSS.interrupt_stack_table.len());
```

Issue #614, opened 2020-06-15
Use sallyport with SGX to communicate with host


lkatalin:
Use the sallyport crate's functionality of message passing between enclave and host (with `Request` and `Reply`) to proxy syscalls:

- [x] Allocate a sallyport::Block on the stack in main()
- [x] Pass a reference to it to the enclave in the first argument.
- [x] Change the enclave to write the syscall registers into the sallyport::Request
- [x] Fill in block buffer in case of write
- [x] Return ~~`ENOBUFS`~~ `EMSGSIZE` if inputs are larger than the block
- [ ] Remove syscall args except aex, ctx
- [ ] Don't preserve `%rbx` when calling VDSO
- [x] Remove `ualloc()` and `ufree()`
- [x] Remove the handler function (pass null to the vdso).
- [x] Handle normal EEXIT in the main loop of main().
- [x] Parse Reply data from syscall

lkatalin:
Note to self: the `%rdi` from sgx-keep's `enter()` is still the same `%rdi` in sgx-shim's `entry()`; that is, `_rdi` [here](https://github.com/enarx/enarx/blob/master/enarx-keep-sgx-shim/src/entry.rs#L72) becomes `block: &mut sallyport::Block`.

PR #668, opened 2020-07-07
Prepare for using Sallyport in SGX

lkatalin:
This series of commits:
- Passes a sallyport Block into the enclave and shim's handler
- Creates a new syscall proxy function that reads arguments out of the handler's Block, but otherwise works the same way as the previous function

Related to #614

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -103,6 +103,25 @@ impl<'a> Handler<'a> {
         }
     }
 
+    #[inline(never)]
+    unsafe fn proxy(&mut self) -> u64 {
+        let ret = syscall(
+            self.block.msg.req.arg[0].raw() as u64, // rdi
+            self.block.msg.req.arg[1].raw() as u64, // rsi
+            self.block.msg.req.arg[2].raw() as u64, // rdx
+            self.aex,
+            self.block.msg.req.arg[4].raw() as u64, // r8
+            self.block.msg.req.arg[5].raw() as u64, // r9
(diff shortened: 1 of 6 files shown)


npmccallum:
@lkatalin Let's do just the first four commits in the first PR.

lkatalin:
Sounds good @npmccallum , I'll re-push.

sallyport/src/lib.rs
```diff
@@ -82,7 +82,8 @@ impl Request {
 #[repr(C)]
 #[derive(Copy, Clone, Default, PartialEq, Debug)]
 pub struct Reply {
-    ret: [Register<usize>; 2],
+    /// Return values from the syscall: rax:rdx
+    pub ret: [Register<usize>; 2],
```
npmccallum, line 86: Definitely don't do this.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -103,6 +103,25 @@ impl<'a> Handler<'a> {
...
+            self.block.msg.req.arg[2].raw() as u64, // rdx
+            self.aex,
+            self.block.msg.req.arg[4].raw() as u64, // r8
+            self.block.msg.req.arg[5].raw() as u64, // r9
+            self.block.msg.req.arg[3].raw() as u64, // r10
+            self.block.msg.req.num.raw() as u64,    // rax
+            self.ctx,
+        );
+
+        self.block.msg.rep.ret[0] = ret.into();
```
npmccallum, line 120:
```suggestion
        self.block.msg.rep = Ok([ret.into(), 0]).into();
```

PR #669, opened 2020-07-07
Use syscall proxy for exit

lkatalin:
Related to #614

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -215,8 +215,9 @@ impl<'a> Handler<'a> {
             .into()
             .map(|x| x.into())
             .unwrap_or_else(|| self.aex.gpr.rdi.raw());
+
         loop {
-            unsafe { self.syscall(libc::SYS_exit as u64, code, 0, 0, 0, 0, 0) };
+            unsafe { self.proxy(Request::new(libc::SYS_exit as usize, &[code.into()])) };
         }
     }
 
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -3,7 +3,8 @@
 use crate::Layout;
 
 use bounds::{Contains, Line, Span};
-use sallyport::Block;
+use memory::Register;
+use sallyport::{Block, Request};
 use sgx_types::ssa::StateSaveArea;
 
 use core::fmt::Write;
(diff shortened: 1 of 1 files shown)


lkatalin:
@npmccallum I should have thought of this earlier, but: would it make more sense to pass parameters into `handler.proxy()` from each implemented syscall and then have `proxy()` update the block? It might make the syscalls themselves slightly more readable and possibly lessen the chance that `proxy()` will be called on a `block` that holds unknown state, but I'm not sure. Is there a reason this way is better?

npmccallum:
@lkatalin If it were me, I'd do `fn proxy(&mut self, Request) -> Result<[Register<usize>; 2], libc::c_int>;`. I would also update all the syscalls to return the same thing. Then you can handle the syscall return values all the same way in `entry.rs`.

lkatalin:
@npmccallum Thanks. I changed the `proxy()` signature. I'll have to update the return types of all the syscalls together in order for it to compile properly. So I plan to do that last.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -117,9 +120,11 @@ impl<'a> Handler<'a> {
             self.ctx,
         );
 
-        self.block.msg.rep = Ok([ret.into(), 0usize.into()]).into();
+        let rep = Ok([ret.into(), 0usize.into()]);
 
-        ret
+        self.block.msg.rep = rep.into();
+
+        rep
```
npmccallum, line 127: This is very close, but unfortunately incorrect.

The `ret` variable contains both the return contents and the errno. So we can't just return `Ok` for every syscall, because one might fail. You can see the logic we use in `sallyport` to handle this [here](https://github.com/enarx/enarx/blob/master/sallyport/src/lib.rs#L89-L117).

It may be simple enough just to convert the `Ok([ret, 0])` into a `Reply` and back again:

```suggestion
        self.block.msg.rep = rep.into();
        self.block.msg.rep.into()
```

... so long as the compiler doesn't try to optimize anything out incorrectly. I *think* we'll be okay.

lkatalin: @npmccallum Happy to do it that way, though if there's some chance the compiler could introduce an error, maybe I should use `match` statements the way we do in `sallyport`. It's a bit more verbose but also probably more readable than converting twice. What do you think?

npmccallum: Let's use the double conversion.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -117,9 +120,10 @@ impl<'a> Handler<'a> {
             self.ctx,
         );
 
-        self.block.msg.rep = Ok([ret.into(), 0usize.into()]).into();
+        let rep = Ok([ret.into(), 0usize.into()]);
 
-        ret
+        self.block.msg.rep = rep.into();
```
npmccallum, line 125: No need for a temporary variable here.

Issue #636, opened 2020-06-25
keep-runtime: adjust preallocated memory size based on Keep implementations


ueno:
wasmtime reserves a memory region at startup that can be seen as a linear memory from WebAssembly code. By default, it is quite large (6GB on 64-bit platforms), though it can be configured through [an option](https://docs.rs/wasmtime/0.18.0/wasmtime/struct.Config.html#method.static_memory_maximum_size). I guess it should be adjusted accordingly to the limits in the Keep implementations.

ueno:
@lkatalin @connorkuehl this is what I mentioned in the standup; since the memory is `mmap`ed with `MAP_PRIVATE`, it is not really allocated, but nevertheless fails the `mmap` call if the size exceeds the heap size allocated for the shim.

npmccallum:
Just a note: we shouldn't presume that Enarx's implementation of `mmap()` has all of Linux's invariants. For example, on sgx today we have to pre-allocate all possible memory for the guest.

platten:
We are focusing efforts towards dynamic memory mapping in the shims vs. tuning this variable.

PR #676, opened 2020-07-09
keep-runtime: Prefer dynamic memory allocation in wasmtime

ueno:
By default, wasmtime internally allocates 4GiB of memory once at the startup, which does not fit well in our SGX heap.  This configures wasmtime to only do dynamic memory allocation.

Related to #636.


ueno:
@npmccallum sorry, there was a stupid logic error in the previous commit; should be fixed now.

PR #678, opened 2020-07-09
Update read syscall to use proxy

lkatalin:
Related to #614 .

Not sure if we still need `ret <= ERRNO_BASE` since `proxy()` now returns an `Ok()` or `Err`?

Edit: I thought about it more and the `ERRNO` check doesn't make sense, so I removed it.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -9,7 +9,6 @@ use sgx_types::ssa::StateSaveArea;
 
 use core::fmt::Write;
 use core::mem::size_of;
-use core::ptr::copy_nonoverlapping;
 use core::slice::{from_raw_parts, from_raw_parts_mut};
 
 const TRACE: bool = false;
@@ -254,31 +253,39 @@ impl<'a> Handler<'a> {
     pub fn read(&mut self) -> u64 {
         self.trace("read", 3);
 
-        let fd = self.aex.gpr.rdi.raw();
-        let buf = self.aex.gpr.rsi.raw() as *mut u8;
-        let size = self.aex.gpr.rdx.raw();
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
+        // Allocate some unencrypted memory from the Block.
+        let cursor = self.block.cursor();
+        let slice: &mut [u8] = match unsafe { cursor.alloc(self.aex.gpr.rdx.into()) } {
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
         };
 
-        unsafe {
-            // Do the syscall; replace encrypted memory with unencrypted memory.
-            let ret = self.syscall(libc::SYS_read as u64, fd, map as _, size, 0, 0, 0);
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let req = request!(libc::SYS_read => self.aex.gpr.rdi, slice, slice.len());
+        let res = unsafe { self.proxy(req) };
 
-            // Copy the unencrypted input into encrypted memory.
-            if ret <= ERRNO_BASE {
-                if ret > size {
+        match res {
+            Ok(res) => {
+                if usize::from(res[0]) > self.aex.gpr.rdx.into() {
                     self.attacked();
                 }
 
-                copy_nonoverlapping(map, buf, ret as _);
+                let tbuf = unsafe { self.aex.gpr.rsi.as_slice_mut(self.aex.gpr.rdx.into()) };
+
+                // Reallocate some unencrypted memory from the Block.
+                let cursor = self.block.cursor();
+                let slice: &mut [u8] = match unsafe { cursor.alloc(self.aex.gpr.rdx.into()) } {
+                    Ok(slice) => slice,
+                    Err(_) => return -libc::EMSGSIZE as u64,
+                };
+
+                // Copy the unencrypted memory into encrypted memory.
+                tbuf.copy_from_slice(&slice[..tbuf.len()]);
+
+                res[0].into()
             }
 
-            self.ufree(map, size);
-            ret
+            Err(code) => -code as u64,
         }
     }
 


lkatalin:
I accidentally closed this.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -252,31 +252,25 @@ impl<'a> Handler<'a> {
...
-        let buf = self.aex.gpr.rsi.raw() as *mut u8;
-        let size = self.aex.gpr.rdx.raw();
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(request!(libc::SYS_read => self.aex.gpr.rdi, &self.block.buf, usize::from(self.aex.gpr.rdx)))
```
npmccallum, line 257: What happens if `self.aex.gpr.rdx` is longer than `self.block.buf.len()`?

lkatalin: Nothing good. I've added a check for this. Thanks.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,29 @@ impl<'a> Handler<'a> {
...
+            return -libc::EMSGSIZE as u64;
+        }
 
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(request!(libc::SYS_read => self.aex.gpr.rdi, &self.block.buf, usize::from(self.aex.gpr.rdx)))
```
npmccallum, line 262:
```suggestion
            self.proxy(request!(libc::SYS_read => self.aex.gpr.rdi, &self.block.buf, self.aex.gpr.rdx))
```

This should work. If not, we probably have a bug in `Register` conversions. `self.aex.gpr.rdx` should be `Register<u64>`, which should infallibly convert to `Register<usize>` on 64-bit systems.

lkatalin:
@npmccallum I'm not sure if there is some way to do this without having to allocate twice.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,40 @@ impl<'a> Handler<'a> {
...
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
         };
 
-        unsafe {
-            // Do the syscall; replace encrypted memory with unencrypted memory.
-            let ret = self.syscall(libc::SYS_read as u64, fd, map as _, size, 0, 0, 0);
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let req =
+            request!(libc::SYS_read => self.aex.gpr.rdi, slice, usize::from(self.aex.gpr.rdx));
```
npmccallum, line 265:
```suggestion
        let req = request!(libc::SYS_read => self.aex.gpr.rdi, slice, slice.len());
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,40 @@ impl<'a> Handler<'a> {
...
-        let buf = self.aex.gpr.rsi.raw() as *mut u8;
-        let size = self.aex.gpr.rdx.raw();
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
+        // Allocate some unencrypted memory from the Block.
+        let mut cursor = self.block.cursor();
+        let slice: &mut [u8] = match unsafe { cursor.alloc(usize::from(self.aex.gpr.rdx)) } {
```
npmccallum, line 258:
```suggestion
        let slice: &mut [u8] = match unsafe { cursor.alloc(self.aex.gpr.rdx.into()) } {
```

enarx-keep-sgx-shim/src/handler.rs
npmccallum: This is defensive coding. We allocate the size of `slice` from `rdx`. But what if someone introduces a bug in the future? The size we pass to the proxy should *always* be the size of the valid memory. Otherwise we crash.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,40 @@ impl<'a> Handler<'a> {
...
+                };
 
-            self.ufree(map, size);
-            ret
+                // Copy the unencrypted memory into encrypted memory.
+                tbuf.copy_from_slice(&slice[..tbuf.len()]);
+
+                res[0].into()
+            }
+            Err(code) => -code as u64,
```
npmccallum, line 289:
```suggestion
            }

            Err(code) => -code as u64,
```

Please always have a vertical space after a closing bracket.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,40 @@ impl<'a> Handler<'a> {
...
-            }
+                let tbuf = unsafe { self.aex.gpr.rsi.as_slice_mut(self.aex.gpr.rdx.into()) };
+
+                // Reallocate some unencrypted memory from the Block.
+                let mut cursor = self.block.cursor();
+                let slice: &mut [u8] = match unsafe { cursor.alloc(usize::from(self.aex.gpr.rdx)) }
+                {
+                    Ok(slice) => slice,
+                    Err(_) => return -libc::EMSGSIZE as u64,
+                };
```
npmccallum, line 282:
```suggestion
                let slice: &mut [u8] = match unsafe { cursor.alloc(self.aex.gpr.rdx.into()) } {
                    Ok(slice) => slice,
                    Err(_) => return -libc::EMSGSIZE as u64,
                };
```

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Appreciate this!

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,39 @@ impl<'a> Handler<'a> {
...
-            // Do the syscall; replace encrypted memory with unencrypted memory.
-            let ret = self.syscall(libc::SYS_read as u64, fd, map as _, size, 0, 0, 0);
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let req = request!(libc::SYS_read => self.aex.gpr.rdi, slice, slice.len());
+        let res = unsafe { self.proxy(req) };
 
-            // Copy the unencrypted input into encrypted memory.
-            if ret <= ERRNO_BASE {
-                if ret > size {
+        match res {
```
kubkon, line 267: You could probably combine these two into one-liner:

```rust
match unsafe { self.proxy(req) } {
...
}
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,39 @@ impl<'a> Handler<'a> {
...
+            Ok(res) => {
+                if usize::from(res[0]) > self.aex.gpr.rdx.into() {
```
kubkon, line 269: Is `res` always guaranteed to be non-empty? If not, index access into an empty container will result in a panic. Either way, I think it is almost always better to use `res.get(0).expect("some msg")` than `res[0]` since it aids in auditing the code.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Had suggested I separate them; however, the match statements will be going away shortly in a separate PR once this returns a `Result` type.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: This is a good point; I'll update it. Thanks @kubkon !

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -254,31 +253,39 @@ impl<'a> Handler<'a> {
...
 
-                copy_nonoverlapping(map, buf, ret as _);
+                let tbuf = unsafe { self.aex.gpr.rsi.as_slice_mut(self.aex.gpr.rdx.into()) };
+
+                // Reallocate some unencrypted memory from the Block.
+                let cursor = self.block.cursor();
+                let slice: &mut [u8] = match unsafe { cursor.alloc(self.aex.gpr.rdx.into()) } {
+                    Ok(slice) => slice,
+                    Err(_) => return -libc::EMSGSIZE as u64,
+                };
```
kubkon, line 280: This pattern of:

```rust
let something = match ... {
   Ok(x) => x,
   Err(_e) => return -libc::EMSGSIZE as u64,
}
```

seems to be repeating a lot in this file. Would it make sense to wrap up the implementation of a syscall in a fallible function such as

```rust
fn read(&mut self) -> u64 {
    fn read_impl(&mut self) -> Result<usize, Errno> {
         // the actual impl where you can use '?'
    }

    match self.read_impl() {
        Ok(x) => x,
        Err(e) => u64::from(e),
    }
}
```

Oh, and this might be done in a future PR. I just thought I'd mention it here for posterity.

kubkon: The `Errno` type could be implemented using the `thiserror` crate I guess.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum The above approach lets us avoid panics, but I think it doesn't work with the `Register` conversions so I'm having to do casts again if I use it - thoughts?

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @kubkon You're completely right, and this is going away.

npmccallum: @kubkon Yes, that is precisely where we are going. See my latest PR which implements this.

enarx-keep-sgx-shim/src/handler.rs
npmccallum: @kubkon The `res` value is a `[Register<usize>; 2]` which conforms to the Linux syscall ABI. It will never be empty.

lkatalin:
@kubkon I think your suggestions will be addressed in a separate PR shortly.

enarx-keep-sgx-shim/src/handler.rs
kubkon: @npmccallum Ah, OK! Thanks for the explanation!

PR #679, opened 2020-07-09
Update write syscall to use proxy

lkatalin:
Related to #614

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -284,29 +284,28 @@ impl<'a> Handler<'a> {
     pub fn write(&mut self) -> u64 {
         self.trace("write", 3);
 
-        let fd = self.aex.gpr.rdi.raw();
-        let buf = self.aex.gpr.rsi.raw() as *const u8;
-        let size = self.aex.gpr.rdx.raw();
+        let input = unsafe { self.aex.gpr.rsi.as_slice(self.aex.gpr.rdx.into()) };
 
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
+        if input.len() > self.block.buf.len() {
+            return -libc::EMSGSIZE as u64;
+        }
+
+        // Copy the encrypted input into unencrypted memory.
+        self.block.buf[..input.len()].copy_from_slice(input);
+
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(request!(libc::SYS_write => self.aex.gpr.rdi, &self.block.buf, input.len()))
         };
 
-        unsafe {
-            // Copy the encrypted input into unencrypted memory.
-            copy_nonoverlapping(buf, map, size as _);
-
-            // Do the syscall; replace encrypted memory with unencrypted memory.
-            let ret = self.syscall(libc::SYS_write as u64, fd, map as _, size, 0, 0, 0);
-            self.ufree(map, size);
-
-            if ret <= ERRNO_BASE && ret > size {
-                self.attacked();
+        match res {
+            Ok(res) => {
+                if usize::from(res[0]) > input.len() {
+                    self.attacked();
+                }
+                res[0].into()
             }
-
-            ret
+            Err(code) => -code as u64,
         }
     }
 
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -216,6 +216,7 @@ impl<'a> Handler<'a> {
             .map(|x| x.into())
             .unwrap_or_else(|| self.aex.gpr.rdi.raw());
 
+        #[allow(unused_must_use)]
         loop {
             unsafe { self.proxy(request!(libc::SYS_exit => code)) };
         }
@@ -233,6 +234,7 @@ impl<'a> Handler<'a> {
             .into()
             .map(|x| x.into())
             .unwrap_or_else(|| self.aex.gpr.rdi.raw());
+        #[allow(unused_must_use)]
         loop {
             unsafe { self.proxy(request!(libc::SYS_exit_group => code)) };
         }


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,25 +285,27 @@ impl<'a> Handler<'a> {
...
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
-        };
-
         unsafe {
             // Copy the encrypted input into unencrypted memory.
-            copy_nonoverlapping(buf, map, size as _);
+            copy_nonoverlapping(buf, &self.block.buf as *const _ as *mut u8, size as _);
```
npmccallum, line 290: Instead of converting the `buf` to a pointer, let's convert the input to a slice and use the slice copy methods. Hoorah for safe code!

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,25 +285,27 @@ impl<'a> Handler<'a> {
...
-            if ret <= ERRNO_BASE && ret > size {
-                self.attacked();
+            match self.proxy(Request::new(
+                libc::SYS_write as usize,
+                &[
+                    fd.into(),
+                    (&self.block.buf as *const _ as usize).into(),
+                    size.into(),
+                ],
+            )) {
```
npmccallum, line 300: Don't wrap the match statement. Break this up.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Do you mean something other than `let input = from_raw_parts(buf, size as usize);`? I have to also wrap that as unsafe.

npmccallum: Yes, that's what I mean.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -288,25 +288,31 @@ impl<'a> Handler<'a> {
...
         let size = self.aex.gpr.rdx.raw();
 
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
-        };
+        // Copy the encrypted input into unencrypted memory.
+        let input = unsafe { from_raw_parts(buf, size as usize) };
+        let ubuf = unsafe { from_raw_parts_mut(&self.block.buf as *const _ as _, size as usize) };
```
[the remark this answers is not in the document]
lkatalin: @npmccallum Is this to give `ubuf` the same alignment as `input`? I think the alignment of `input` is `1` since it's a `&[u8]`.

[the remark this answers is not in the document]
npmccallum: If `buf` is already the correct type, why are you using `from_raw_parts_mut()` at all?

npmccallum, line 293: I think what you want is something like this:

```suggestion
        let input: &[u8] = unsafe { from_raw_parts(self.aex.gpr.rsi.into(), self.aex.gpr.rdx.into()) };
        
        if input.len() > self.block.buf.len() { return -EMSGSIZE; }
        self.block.buf[..input.len()].copy_from_slice(input);
};
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -288,25 +288,31 @@ impl<'a> Handler<'a> {
...
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(Request::new(
+                libc::SYS_write as usize,
+                &[
+                    fd.into(),
+                    (&self.block.buf as *const _ as usize).into(),
+                    size.into(),
+                ],
+            ))
```
npmccallum, line 305: Use `request!()`. That should remove all the boilerplate.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -288,25 +288,31 @@ impl<'a> Handler<'a> {
...
+        };
 
-            if ret <= ERRNO_BASE && ret > size {
-                self.attacked();
+        match res {
+            Ok(res) => {
+                if res[0].raw() as u64 > size {
```
npmccallum, line 310:
```suggestion
                if res[0].into() > size {
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -288,25 +288,31 @@ impl<'a> Handler<'a> {
...
+                    self.attacked();
+                }
+                res[0].raw() as u64
```
npmccallum, line 313:
```suggestion
                res[0].into()
```

enarx-keep-sgx-shim/src/handler.rs
lkatalin: I used this except I kept `size` as a variable since it's used several times.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: The compiler couldn't infer a type for this, so I have it as `.raw()` (but not a `u64` anymore since `size` is a `usize` now).

lkatalin:
I'm totally unsure if handling the Result in `exit` in this way is okay. It's strange to think about error handling in a function that doesn't return.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -217,7 +217,10 @@ impl<'a> Handler<'a> {
             .unwrap_or_else(|| self.aex.gpr.rdi.raw());
 
         loop {
-            unsafe { self.proxy(request!(libc::SYS_exit => code)) };
+            match unsafe { self.proxy(request!(libc::SYS_exit => code)) } {
+                Ok(_) => (),
+                Err(_) => (),
+            }
```
npmccallum, line 223: Instead of trying to consume the result, which will leave a later reader of this code wondering what it does, use the `#[allow(unused_must_use)]` attribute which makes it clear that we are ignoring the return value on purpose.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -234,7 +237,10 @@ impl<'a> Handler<'a> {
             .map(|x| x.into())
             .unwrap_or_else(|| self.aex.gpr.rdi.raw());
         loop {
-            unsafe { self.proxy(request!(libc::SYS_exit_group => code)) };
+            match unsafe { self.proxy(request!(libc::SYS_exit_group => code)) } {
+                Ok(_) => (),
+                Err(_) => (),
+            }
```
npmccallum, line 243: See above.

enarx-keep-sgx-shim/src/handler.rs
npmccallum: Also, if you did want to consume a variable, using `drop(value)` would be much cleaner.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,28 +285,30 @@ impl<'a> Handler<'a> {
         self.trace("write", 3);
 
         let fd = self.aex.gpr.rdi.raw();
-        let buf = self.aex.gpr.rsi.raw() as *const u8;
-        let size = self.aex.gpr.rdx.raw();
+        let size = self.aex.gpr.rdx.raw() as usize;
```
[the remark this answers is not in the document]
lkatalin: Interesting, I was trying to avoid calling `len()` so often - good to know!

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,28 +285,30 @@ impl<'a> Handler<'a> {
...
 
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size) {
-            Err(errno) => return errno as u64,
-            Ok(map) => map,
-        };
+        let input: &[u8] = unsafe { from_raw_parts(self.aex.gpr.rsi.raw() as *const u8, size) };
```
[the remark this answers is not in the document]
lkatalin: @npmccallum I've got
```
let input: &[u8] = unsafe { from_raw_parts(self.aex.gpr.rsi.into(), self.aex.gpr.rdx.into()) };
    |                                                                     ^^^^ the trait `core::convert::From<memory::register::Register<u64>>` is not implemented for `*const u8`

```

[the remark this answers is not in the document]
lkatalin: And it is rebased on the new changes to `Register`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,28 +285,30 @@ impl<'a> Handler<'a> {
...
-            }
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(request!(libc::SYS_write => fd, &self.block.buf as *const _ as usize, size))
+        };
 
-            ret
+        match res {
+            Ok(res) => {
+                if res[0].raw() > size {
```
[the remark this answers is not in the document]
lkatalin: `into()` gives the error `cannot infer type for type parameter T declared on the trait Into`. But `usize::from(res[0])` works fine, so I'll use that.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -285,28 +285,30 @@ impl<'a> Handler<'a> {
...
-            self.ufree(map, size);
+        // Copy the encrypted input into unencrypted memory.
+        self.block.buf[..input.len()].copy_from_slice(input);
 
-            if ret <= ERRNO_BASE && ret > size {
-                self.attacked();
-            }
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let res = unsafe {
+            self.proxy(request!(libc::SYS_write => fd, &self.block.buf as *const _ as usize, size))
```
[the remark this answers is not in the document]
lkatalin: This one did not work as suggested either.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -284,29 +284,28 @@ impl<'a> Handler<'a> {
...
+        match res {
+            Ok(res) => {
+                if usize::from(res[0]) > input.len() {
+                    self.attacked();
+                }
+                res[0].into()
             }
-
-            ret
+            Err(code) => code as u64,
```
npmccallum, line 310: `Err(code)` contains the positive `errno`. But `code as u64` needs the negative `errno`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -208,6 +208,7 @@ impl<'a> Handler<'a> {
     /// Proxy an exit() syscall
     ///
     /// The optional `code` parameter overrides the value from `aex`.
+    #[allow(unused_must_use)]
```
npmccallum, line 211: We don't want to allow this on the whole function. Just the `self.proxy()` invocation. If for some reason you can't put the attribute on the invocation itself, put it on the `loop {}`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -226,6 +227,7 @@ impl<'a> Handler<'a> {
     /// The optional `code` parameter overrides the value from `aex`.
     /// TODO: Currently we are only using one thread, so this will behave the
     /// same way as exit(). In the future, this implementation will change.
+    #[allow(unused_must_use)]
```
npmccallum, line 230: See above.

PR #680, opened 2020-07-09
Update readv syscall to use proxy

lkatalin:
Related to #614 

The lines
```
let ptr: Register<usize> = uiovec.as_ptr().into();
let len: Register<usize> = uiovec.len().into();
```

are necessary to satisfy the borrow checker.

Edit: The `match` statements and `unwrap()`s are going away in a subsequent PR.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -351,79 +351,67 @@ impl<'a> Handler<'a> {
     pub fn readv(&mut self) -> u64 {
         self.trace("readv", 3);
 
-        let fd = self.aex.gpr.rdi.raw();
-        let trusted = unsafe {
-            from_raw_parts_mut(
-                self.aex.gpr.rsi.raw() as *mut libc::iovec,
-                self.aex.gpr.rdx.raw() as usize,
-            )
+        let tiovec: &mut [libc::iovec] =
+            unsafe { self.aex.gpr.rsi.as_slice_mut(self.aex.gpr.rdx.into()) };
+
+        // Allocate the correct number of iovecs in untrusted memory.
+        let cursor = self.block.cursor();
+        let uiovec: &mut [libc::iovec] = match unsafe { cursor.alloc::<libc::iovec>(tiovec.len()) }
+        {
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
         };
 
-        // Add up total size of buffers and size of iovec array.
-        let bufsize = trusted.iter().fold(0, |a, e| a + e.iov_len);
-        let iovecsize = size_of::<libc::iovec>() * trusted.len();
-        let size = bufsize + iovecsize;
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size as u64) {
-            Err(errno) => return errno as u64,
-            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
-        };
-
-        // Split allocated memory into that used by the iovec struct array and that used by its buffers.
-        let (uiovec, ubuffer) = map.split_at_mut(iovecsize);
-
-        // Convert the prefix from a byte slice into an iovec slice.
-        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<libc::iovec>() };
-        if untrusted.len() != trusted.len() {
-            self.attacked();
-        }
-
-        // Set pointers in unencrypted iovec slice to use the rest of the allocated memory.
-        // The offset is into the buffer area allocated immediately after the iovec struct
-        // array, measured in bytes.
-        let mut offset = 0;
-        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-            u.iov_base = ubuffer[offset..].as_mut_ptr() as *mut _;
+        // Update pointers from iovecs to buffers in untrusted memory.
+        let mut size = 0;
+        for (t, u) in tiovec.iter().zip(uiovec.iter_mut()) {
+            size += t.iov_len;
             u.iov_len = t.iov_len;
-            offset += t.iov_len;
+            u.iov_base = match unsafe { cursor.alloc(t.iov_len) } {
+                Ok(base) => base.as_mut_ptr(),
+                Err(_) => return -libc::EMSGSIZE as u64,
+            };
         }
 
         // Do the syscall; replace encrypted memory with unencrypted memory.
-        let ret = unsafe {
-            self.syscall(
-                libc::SYS_readv as u64,
-                fd,
-                untrusted.as_ptr() as _,
-                untrusted.len() as u64,
-                0,
-                0,
-                0,
-            )
-        };
+        let req = request!(libc::SYS_readv => self.aex.gpr.rdi, uiovec, uiovec.len());
+        let res = unsafe { self.proxy(req) };
 
-        // Copy the unencrypted input into encrypted memory.
-        if ret <= ERRNO_BASE {
-            if ret > size as u64 {
-                self.attacked();
-            }
-
-            let mut offset = 0;
-            for (t, u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-                let us = &mut ubuffer[offset..][..t.iov_len];
-                offset += t.iov_len;
-
-                if u.iov_base != us.as_mut_ptr() as *mut _ || u.iov_len != t.iov_len {
+        match res {
+            Ok(res) => {
+                let mut read = res[0].into();
+                if size < read {
                     self.attacked();
                 }
 
-                let ts = unsafe { from_raw_parts_mut(t.iov_base as *mut u8, t.iov_len) };
-                ts.copy_from_slice(us);
-            }
-        }
+                // Allocate the correct number of iovecs in untrusted memory.
+                // We reallocate again here to keep the Rust compiler happy.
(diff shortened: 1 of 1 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -45,6 +45,35 @@ extern "C" {
...
+            padding.len() + size_of::<T>() * count
+        };
+
+        let (data, next) = buf.split_at_mut(mid);
+
+        self.0.replace(next);
+
+        Ok(unsafe { data.align_to_mut::<T>().1 })
+    }
+}
```
npmccallum, line 75: See my comments about the cursor in the other PR. Note that cursor should probably be its own PR (and should probably have tests).

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -348,78 +377,58 @@ impl<'a> Handler<'a> {
...
-        // The offset is into the buffer area allocated immediately after the iovec struct
-        // array, measured in bytes.
-        let mut offset = 0;
-        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-            u.iov_base = ubuffer[offset..].as_mut_ptr() as *mut _;
-            u.iov_len = t.iov_len;
-            offset += t.iov_len;
-        }
+        let ptr: Register<usize> = uiovec.as_ptr().into();
+        let len: Register<usize> = uiovec.len().into();
```
npmccallum, line 399: No longer needed.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -348,78 +377,58 @@ impl<'a> Handler<'a> {
...
 
-        // Convert the prefix from a byte slice into an iovec slice.
-        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<libc::iovec>() };
-        if untrusted.len() != trusted.len() {
-            self.attacked();
+        // Update pointers from iovecs to buffers in untrusted memory.
+        let mut size = 0;
+        for (t, u) in tiovec.iter().zip(uiovec.iter_mut()) {
+            size += t.iov_len;
+            u.iov_base = c.alloc(t.iov_len).unwrap().as_mut_ptr();
```
npmccallum, line 395:
```suggestion
            u.iov_len = t.iov_len;
            u.iov_base = c.alloc(t.iov_len).unwrap().as_mut_ptr();
```

Don't forget to initialize the length!

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -348,78 +377,58 @@ impl<'a> Handler<'a> {
...
-                0,
-                0,
-                0,
-            )
+        let res = unsafe {
+            self.proxy(Request::new(
+                libc::SYS_readv as usize,
+                &[fd.into(), ptr, len],
+            ))
         };
```
npmccallum, line 407: Use `request!()`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -348,78 +377,58 @@ impl<'a> Handler<'a> {
...
+                let mut c = Cursor::new(&mut self.block.buf);
+                let uiovec: &mut [libc::iovec] = c.alloc(tiovec.len()).unwrap();
+
+                for (t, u) in tiovec.iter_mut().zip(uiovec.iter_mut()) {
+                    u.iov_base = c.alloc(t.iov_len).unwrap().as_mut_ptr();
+                    // Grab data from untrusted buffers and copy them to trusted buffers
+                    let ubuf = unsafe { from_raw_parts(u.iov_base as *mut u8, u.iov_len) };
+                    let tbuf = unsafe { from_raw_parts_mut(t.iov_base as *mut u8, t.iov_len) };
+                    tbuf.copy_from_slice(ubuf);
+                }
```
npmccallum, line 426: We probably need to be more defensive in this section. `res[0]` returns the number of bytes read. We should copy no more than this number of bytes into the trusted memory.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Good catch, thanks.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: I've added a check for this.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -351,79 +351,67 @@ impl<'a> Handler<'a> {
...
 
-        let fd = self.aex.gpr.rdi.raw();
-        let trusted = unsafe {
-            from_raw_parts_mut(
-                self.aex.gpr.rsi.raw() as *mut libc::iovec,
-                self.aex.gpr.rdx.raw() as usize,
-            )
-        };
+        let tiovec: &mut [libc::iovec] =
+            unsafe { self.aex.gpr.rsi.as_slice_mut(self.aex.gpr.rdx.into()) };
```
npmccallum, line 355: In my version of this function, I avoid the wrap by inferring the type and passing the `tiovec` to `cursor.copy_slice()`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -351,79 +351,67 @@ impl<'a> Handler<'a> {
...
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size as u64) {
-            Err(errno) => return errno as u64,
-            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
+        // Allocate the correct number of iovecs in untrusted memory.
+        let cursor = self.block.cursor();
+        let uiovec: &mut [libc::iovec] = match unsafe { cursor.alloc(tiovec.len()) } {
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
         };
```
npmccallum, line 362: Using `copy_slice()` here makes the loop below smaller because you don't need to set `u.iov_len`. You just patch up the `u.iov_base`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -351,79 +351,67 @@ impl<'a> Handler<'a> {
...
+        // Update pointers from iovecs to buffers in untrusted memory.
+        let mut size = 0;
+        for (t, u) in tiovec.iter().zip(uiovec.iter_mut()) {
+            size += t.iov_len;
             u.iov_len = t.iov_len;
-            offset += t.iov_len;
+            u.iov_base = match unsafe { cursor.alloc(t.iov_len) } {
+                Ok(base) => base.as_mut_ptr(),
+                Err(_) => return -libc::EMSGSIZE as u64,
+            };
```
npmccallum, line 372: There is a very subtle bug here. Relying on type inference means `cursor.alloc()` returns a `&mut [c_void]`. But `size_of::<c_void>() == 0`. So you're actually allocating zero bytes here, regardless of the value of `t.iov_len`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -351,79 +351,67 @@ impl<'a> Handler<'a> {
...
+                        Ok(base) => base.as_mut_ptr(),
+                        Err(_) => return -libc::EMSGSIZE as u64,
+                    };
+
+                    // Grab data from untrusted buffers and copy to trusted buffers.
+                    let ubuf = unsafe { from_raw_parts(u.iov_base as *mut u8, u.iov_len) };
+                    let tbuf = unsafe { from_raw_parts_mut(t.iov_base as *mut u8, t.iov_len) };
+                    if bytes_copied < res[0].into() {
+                        tbuf.copy_from_slice(ubuf);
+                    }
```
npmccallum, line 406: I'd invite you to compare this implementation with my version.

1. Rather than starting at zero and counting up I start at the bytes read and count down. This allows me to efficiently use `core::cmp::min()` to get the bytes to read, which reduces to zero once we've copied all the read bytes.

2. I never use the value of the untrusted iovecs. Because we handed this memory to the host, we'd have to validate it all for malicious modifications. Instead, I simply reproduce the slices as we did before the syscall. Because this is deterministic, we get the same results but don't have to validate the values.

lkatalin: I like this approach.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Thank you for pointing this out. I hope I fixed it with `alloc::<usize>`.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Re: this and the above comment, I tried to implement this but ran into a ball-of-yarn problem. Because of how my code is organized, I can't get rid of the explicit typing for the `iovec` array; and trying to introduces other borrowing errors. Seems to make sense to solve this problem with your PR?

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Please see below.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Actually, I think part of your PR may fix this problem for me (returning `mut` from `copy_slice()`).

lkatalin:
Superseded by #735

PR #681, opened 2020-07-09
Add syscall clock_gettime

### LINKED ISSUE #664: enarx-keep-sgx-shim: clock_gettime syscall
wasmtime's codegen [gathers timing information](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/timing.rs) for each compilation pass using `clock_gettime(CLOCK_MONOTNIC, ...)`. While on the host this is handled through vDSO, this turns into a normal syscall inside a Keep.

I'm not sure how to handle time information within a Keep in general, but maybe the shim could provide a stub.

ziyi-yan:
Closes #664

I implemented syscall `clock_gettime` according to the [Usage section](https://github.com/enarx/enarx/tree/master/sallyport#usage) from sallyport. I used `self.block.cursor()` to allocate a buffer memory for host to write the result and then copy it back to the address in enclave.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -448,4 +448,26 @@ impl<'a> Handler<'a> {
 
         Ok([trusted.len().into(), Default::default()])
     }
+
+    // Do clock_gettime syscall
+    pub fn clock_gettime(&mut self) -> sallyport::Result {
+        self.trace("clock_gettime", 2);
+
+        let clk_id = self.aex.gpr.rdi;
+        let trusted = unsafe { self.aex.gpr.rsi.into_slice_mut(1usize) };
+
+        let cursor = self.block.cursor();
+        let untrusted = unsafe { cursor.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))? };
+        let req = request!(libc::SYS_clock_gettime => clk_id, untrusted);
+        let res = unsafe { self.proxy(req)? };
+
+        if 0usize != res[0].into() {
+            self.attacked();
+        }
+
+        let cursor = self.block.cursor();
+        let untrusted = unsafe { cursor.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))? };
+        trusted.copy_from_slice(untrusted);
+        Ok(res)
+    }
 }
--- a/integration-tests/src/bin/clock_gettime.rs
+++ b/integration-tests/src/bin/clock_gettime.rs
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::clock_gettime;
+
+fn main() {
+    let mut ts = libc::timespec {
+        tv_nsec: Default::default(),
+        tv_sec: Default::default(),
+    };
+
+    unsafe {
+        clock_gettime(libc::CLOCK_MONOTONIC, &mut ts as *mut _);
+    }
+}
--- a/integration-tests/tests/syscall_clock_gettime.rs
+++ b/integration-tests/tests/syscall_clock_gettime.rs
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::all)]
+
+mod common;
+use common::IntegrationTest;
+
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn clock_gettime() {
+    IntegrationTest::new("clock_gettime").run(5, 0, "", "", "");
+}
--- a/enarx-keep-sgx-shim/src/event.rs
+++ b/enarx-keep-sgx-shim/src/event.rs
@@ -49,7 +49,7 @@ pub extern "C" fn event(
                         libc::SYS_rt_sigprocmask => h.rt_sigprocmask(),
                         libc::SYS_sigaltstack => h.sigaltstack(),
                         libc::SYS_getrandom => h.getrandom(),
-
+                        libc::SYS_clock_gettime => h.clock_gettime(),
                         syscall => {
                             debugln!(h, "unsupported syscall: 0x{:x}", syscall as u64);
                             Err(libc::ENOSYS).into()


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -641,6 +641,33 @@ impl<'a> Handler<'a> {
...
+
+    // Do clock_gettime syscall
+    pub fn clock_gettime(&mut self) -> u64 {
+        self.trace("clock_gettime", 2);
+
+        let clk_id = self.aex.gpr.rdi.raw();
+        let ts = self.aex.gpr.rsi.raw() as *mut libc::timespec;
+
+        unsafe {
+            let uaddr: *mut libc::timespec = core::mem::transmute(&mut self.block.buf[0]);
```
npmccallum, line 653: Don't use `transmute` here. We don't know that the alignment of the buffer is correct. Instead, use https://doc.rust-lang.org/std/primitive.slice.html#method.align_to in order to get a reference to a `timespec`. If you do this, much more of this code will be safe.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -644,4 +644,28 @@ impl<'a> Handler<'a> {
...
+        let clk_id = self.aex.gpr.rdi.raw();
+        let ts = self.aex.gpr.rsi.raw() as *mut libc::timespec;
+
+        unsafe {
+            let (_, untrusted, _) = self.block.buf.align_to::<libc::timespec>();
+            let uaddr = untrusted.as_ptr();
+            let result = self.proxy(request!(libc::SYS_clock_gettime => clk_id, uaddr));
+            match result {
+                Ok([ret, _]) => {
+                    let ret = ret.raw() as u64;
```
kubkon, line 661: If I may offer some word of advice here, it is almost always better to use safe wrappers around explicit casts such as `From/Into` or `TryFrom/TryInto` rather than the `as` keyword. In this particular case, as far as I can see, `ret.raw() as u64` is fine here, however rewriting is as `let ret: u64 = ret.raw().into()` will force the compiler to trigger a compile error if it is impossible to do a safe cast (i.e., if we were to cast an integer of higher resolution into a smaller one such as `u64` into `u32`).

ziyi-yan: Thanks! Got it fixed in the latest change.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -644,4 +644,28 @@ impl<'a> Handler<'a> {
 
         trusted.len() as u64
     }
+
+    // Do clock_gettime syscall
+    pub fn clock_gettime(&mut self) -> u64 {
+        self.trace("clock_gettime", 2);
+
+        let clk_id = self.aex.gpr.rdi.raw();
+        let ts = self.aex.gpr.rsi.raw() as *mut libc::timespec;
```
npmccallum, line 653:
```suggestion
        let ts: *mut libc::timespec = self.aex.gpr.rsi.into();
```

npmccallum: FYI: `raw()` is deprecated in https://github.com/enarx/enarx/pull/735

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -644,4 +644,28 @@ impl<'a> Handler<'a> {
...
+
+        unsafe {
+            let (_, untrusted, _) = self.block.buf.align_to::<libc::timespec>();
+            let uaddr = untrusted.as_ptr();
+            let result = self.proxy(request!(libc::SYS_clock_gettime => clk_id, uaddr));
```
npmccallum, line 658: Since you wrote this, we added the `Cursor` type which should help with allocations like this.

```suggestion
            let cursor = self.block.cursor();
            let mut untrusted = cursor.copy_slice(&[*ts]).or(Err(libc::EINVAL))?;
            let result = self.proxy(request!(libc::SYS_clock_gettime => clk_id, &mut untrusted[0]))?;
            if result[0] != 0 { self.attacked(); }
            *ts = untrusted[0];
            Ok(result)
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -448,4 +448,32 @@ impl<'a> Handler<'a> {
...
+            match res {
+                Ok([ret, _]) => {
+                    let ret: usize = ret.into();
+                    if ret == 0 {
+                        trusted.copy_from_slice(untrusted);
+                    }
+                }
+                Err(_) => {},
+            };
+            res
```
npmccallum, line 476:
```suggestion
            let res = self.proxy(req)?;

            // 0 is the only valid success value.
            if 0usize != res[0].into() {
                self.attacked();
            }

            let cursor = self.block.cursor();
            let untrusted = cursor.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))?;
            trusted.copy_from_slice(untrusted);
            Ok(res)
```

integration-tests/tests/syscall_clock_gettime.rs
```diff
@@ -0,0 +1,12 @@
...
+#![deny(clippy::all)]
+
+mod common;
+use common::IntegrationTest;
+
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn clock_gettime() {
+    IntegrationTest::new("clock_gettime").run(5, 0, "", "", "");
+}
```
npmccallum, line 12: Please ensure the file has a newline at the end.

integration-tests/src/bin/clock_gettime.rs
```diff
@@ -0,0 +1,14 @@
...
+use libc::clock_gettime;
+
+fn main() {
+    let mut ts = libc::timespec {
+        tv_nsec: Default::default(),
+        tv_sec: Default::default(),
+    };
+
+    unsafe {
+        clock_gettime(libc::CLOCK_MONOTONIC, &mut ts as *mut _);
```
connorkuehl, line 12: Is there any kind of assertion we can make here to confirm our assumptions about this test and determine pass/fail?

npmccallum: @connorkuehl Not yet. See #747.

npmccallum: What we have currently is that this payload executed the syscall correctly and didn't crash and didn't hang. #747 will confirm that the actual outputs are correct.

connorkuehl: Makes sense. Thanks!

PR #682, opened 2020-07-09
Update writev syscall to use proxy

lkatalin:
Related to #614 

Needs to be rebased after #678 and #680 are merged.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -415,72 +415,44 @@ impl<'a> Handler<'a> {
     pub fn writev(&mut self) -> u64 {
         self.trace("writev", 3);
 
-        let fd = self.aex.gpr.rdi.raw();
-        let trusted = unsafe {
-            from_raw_parts_mut(
-                self.aex.gpr.rsi.raw() as *mut libc::iovec,
-                self.aex.gpr.rdx.raw() as usize,
-            )
+        let tiovec: &[libc::iovec] = unsafe { self.aex.gpr.rsi.as_slice(self.aex.gpr.rdx.into()) };
+
+        // Allocate the correct number of iovecs in untrusted memory.
+        let c = self.block.cursor();
+        let uiovec: &mut [libc::iovec] = match unsafe { c.alloc(tiovec.len()) } {
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
         };
 
-        // Add up total size of buffers and size of iovec array.
-        let bufsize = trusted.iter().fold(0, |a, e| a + e.iov_len);
-        let iovecsize = size_of::<libc::iovec>() * trusted.len();
-        let size = bufsize + iovecsize;
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size as u64) {
-            Err(errno) => return errno as u64,
-            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
-        };
-
-        // Split allocated memory into that used by the iovec struct array
-        // and that used by its buffers.
-        let (uiovec, ubuffer) = map.split_at_mut(iovecsize);
-
-        // Convert the prefix from a byte slice into an iovec slice.
-        let (_, untrusted, _) = unsafe { uiovec.align_to_mut::<libc::iovec>() };
-        if untrusted.len() != trusted.len() {
-            self.attacked();
-        }
-
-        // Set pointers in unencrypted iovec slice to use the rest
-        // of the allocated memory, then copy the encrypted input
-        // into unencrypted memory. The offset is into the buffer
-        // area allocated immediately after the iovec struct array,
-        // measured in bytes.
-        let mut offset = 0;
-        for (t, mut u) in trusted.iter_mut().zip(untrusted.iter_mut()) {
-            let ts = unsafe { from_raw_parts(t.iov_base as *const u8, t.iov_len) };
-            let us = &mut ubuffer[offset..][..t.iov_len];
-            offset += t.iov_len;
-
-            us.copy_from_slice(ts);
-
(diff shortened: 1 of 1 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -45,6 +45,35 @@ extern "C" {
...
+            padding.len() + size_of::<T>() * count
+        };
+
+        let (data, next) = buf.split_at_mut(mid);
+
+        self.0.replace(next);
+
+        Ok(unsafe { data.align_to_mut::<T>().1 })
+    }
+}
```
npmccallum, line 75: I know I wrote this code. But it has some problems.

1. We should try to implement this without the `Option`. I forget why I added it. But it should be possible. This will remove the cases of `.unwrap()`.

2. `alloc()` either needs to take `T: Default` and use `Default::default()` to initialize all the elements of the returned array or it needs to be made `unsafe`. As it stands now, the function is safe but it returns an array to uninitialized elements. I think I'd go for the `unsafe` option.

3. The cursor should probably store a return value in case allocation fails. This means we can specify the `Err` return type as a type parameter and it will simplify error handling.

4. This probably belongs in the `sallyport` crate. If so, we can make the `buf` field private. The `Block` type would just grow a `fn cursor<'a, T>(err: T) -> Cursor<'a, T>` which would replace `Cursor::new()`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -430,71 +459,49 @@ impl<'a> Handler<'a> {
...
+        let mut c = Cursor::new(&mut self.block.buf);
+        let uiovec: &mut [libc::iovec] = c.alloc(tiovec.len()).unwrap();
+
+        // Update pointers from iovecs to buffers in untrusted memory.
+        // Copy encrypted input to unencrypted memory
+        let mut size = 0;
+        for (t, u) in tiovec.iter().zip(uiovec.iter_mut()) {
+            size += t.iov_len;
+            u.iov_base = c.alloc(t.iov_len).unwrap().as_mut_ptr();
+            let ubuf = unsafe { from_raw_parts_mut(u.iov_base as *mut u8, u.iov_len) };
```
npmccallum, line 479: What is the value of `u.iov_len` here?

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -430,71 +459,49 @@ impl<'a> Handler<'a> {
...
-            let us = &mut ubuffer[offset..][..t.iov_len];
-            offset += t.iov_len;
-
-            us.copy_from_slice(ts);
-
-            u.iov_base = us.as_mut_ptr() as *mut _;
-            u.iov_len = us.len();
-        }
+        let ptr = Register::from_raw(uiovec.as_ptr() as usize);
+        let len = Register::from_raw(uiovec.len() as usize);
```
npmccallum, line 485: No longer needed.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -430,71 +459,49 @@ impl<'a> Handler<'a> {
...
-                untrusted.len() as u64,
-                0,
-                0,
-                0,
-            )
+        let res = unsafe {
+            self.proxy(Request::new(
+                libc::SYS_writev as usize,
+                &[fd.into(), ptr.into(), len.into()],
+            ))
```
npmccallum, line 492: Use `request!()`.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @npmccallum Are you sure about making the `buf` private? We can't use `if input.len() > self.block.buf.len()` in the `handler` if that field is private.

lkatalin: I created #706 for this.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -415,72 +415,44 @@ impl<'a> Handler<'a> {
...
-
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(size as u64) {
-            Err(errno) => return errno as u64,
-            Ok(map) => unsafe { from_raw_parts_mut(map, size as usize) },
+        // Allocate the correct number of iovecs in untrusted memory.
+        let c = self.block.cursor();
+        let uiovec: &mut [libc::iovec] = match unsafe { c.alloc(tiovec.len()) } {
+            Ok(slice) => slice,
+            Err(_) => return -libc::EMSGSIZE as u64,
```
npmccallum, line 424: If you use `copy_slice()` here then you don't need unsafe code and you only need to patch up the `iov_base`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -415,72 +415,44 @@ impl<'a> Handler<'a> {
...
+        for (t, u) in tiovec.iter().zip(uiovec.iter_mut()) {
+            size += t.iov_len;
+            u.iov_len = t.iov_len;
+            u.iov_base = match unsafe { c.alloc(t.iov_len) } {
+                Ok(base) => base.as_mut_ptr(),
+                Err(_) => return -libc::EMSGSIZE as u64,
+            };
+            let ubuf = unsafe { from_raw_parts_mut(u.iov_base as *mut u8, t.iov_len) };
+            let tbuf = unsafe { from_raw_parts(t.iov_base as *mut u8, t.iov_len) };
+            ubuf.copy_from_slice(tbuf);
```
npmccallum, line 439: There's a lot of code here that could  be simplified by using `copy_slice()`.

lkatalin: @npmccallum I believe this only works when `copy_slice()` returns `mut`, which it does not do yet as your PR is scheduled to be merged after this.

lkatalin:
Superseded by #735

PR #685, opened 2020-07-10
Update handler's implementation of Write to use proxy

lkatalin:
Related to #614 

Not sure if `copy_nonoverlapping` should be changed to `copy_from_slice` since it was already there.


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -55,33 +55,28 @@ pub struct Handler<'a> {
...
-                map as _,
-                len,
-                0,
-                0,
-                0,
+            core::ptr::copy_nonoverlapping(
+                bytes.as_ptr(),
+                self.block.buf.as_mut_ptr(),
+                bytes.len(),
             );
```
npmccallum, line 69: Use the safe slice copy functions instead of pointer copies.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -55,33 +55,28 @@ pub struct Handler<'a> {
...
-            self.ufree(map, len);
+            // Do the syscall; replace encrypted memory with unencrypted memory.
+            self.proxy(Request::new(
+                libc::SYS_write as usize,
+                &[
+                    (libc::STDERR_FILENO as usize).into(),
+                    (&self.block.buf as *const _ as usize).into(),
+                    len.into(),
+                ],
+            ));
```
npmccallum, line 79: Use `request!()`.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -54,37 +54,24 @@ pub struct Handler<'a> {
...
         }
 
-        // Allocate some unencrypted memory.
-        let map = match self.ualloc(len) {
-            Err(_) => return Err(core::fmt::Error),
-            Ok(map) => map,
+        let mut cursor = self.block.cursor();
+        let untrusted = match cursor.copy_slice(s.as_bytes()).or(Err(core::fmt::Error)) {
+            Ok(slice) => slice,
+            Err(e) => return Err(e),
```
connorkuehl, line 64: How about the `?` operator here?

```diff
diff --git a/enarx-keep-sgx-shim/src/handler.rs b/enarx-keep-sgx-shim/src/handler.rs
index b8bc7f1..0be3cbd 100644
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -59,10 +59,7 @@ impl<'a> Write for Handler<'a> {
         }
 
         let mut cursor = self.block.cursor();
-        let untrusted = match cursor.copy_slice(s.as_bytes()).or(Err(core::fmt::Error)) {
-            Ok(slice) => slice,
-            Err(e) => return Err(e),
-        };
+        let untrusted = cursor.copy_slice(s.as_bytes()).or(Err(core::fmt::Error))?;
 
         // Do the syscall; replace encrypted memory with unencrypted memory.
         let req = request!(libc::SYS_write => libc::STDERR_FILENO, untrusted, s.as_bytes().len());
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -54,37 +54,24 @@ pub struct Handler<'a> {
...
-                libc::SYS_write as u64,
-                libc::STDERR_FILENO as u64,
-                map as _,
-                len,
-                0,
-                0,
-                0,
-            );
+        // Do the syscall; replace encrypted memory with unencrypted memory.
+        let req = request!(libc::SYS_write => libc::STDERR_FILENO, untrusted, s.as_bytes().len());
```
connorkuehl, line 68: ~~`s.as_bytes().len()`~~ -> `untrusted.len()`

enarx-keep-sgx-shim/src/handler.rs
lkatalin: @connorkuehl When I had tried to use `?`, it told me it couldn't convert the error type from `copy_slice`, which is `()`, to a `core::fmt::Error` (which is required by the `Write` trait).

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -54,37 +54,24 @@ pub struct Handler<'a> {
...
+        let res = unsafe { self.proxy(req) };
 
-            self.ufree(map, len);
+        match res {
+            Ok(_) => Ok(()),
+            Err(_) => Err(core::fmt::Error),
```
connorkuehl, line 73: `write` might return a number of bytes that is less than the number you set out to write. Does the shim care in this case? I know it's just a shim-print routine.

enarx-keep-sgx-shim/src/handler.rs
connorkuehl: Did you leave on the `.or(Err(core::fmt::Error))`? That diff in my previous comment compiles for me.

lkatalin: Let me give it a shot.

lkatalin: You were right! :) I think when I tried it maybe my code was structured differently.

enarx-keep-sgx-shim/src/handler.rs
lkatalin: Good point, I've included a check for this. It returns an error if the two aren't equal, because I figure if it writes more bytes than we expect then that's also bad. I wasn't able to use `untrusted.len()` there because of the borrow checker.

lkatalin:
Thanks for the helpful comments, @connorkuehl .

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -54,37 +54,26 @@ pub struct Handler<'a> {
...
 
-            self.ufree(map, len);
+        match res {
+            Ok(res) => {
+                if usize::from(res[0]) != s.bytes().len() {
+                    return Err(core::fmt::Error);
+                }
+                Ok(())
+            }
+            Err(_) => Err(core::fmt::Error),
```
npmccallum, line 75:
```suggestion
            Ok(res) if usize::from(res[0]) > s.bytes().len() => self.attacked(),
            Ok(res) if usize::from(res[0]) == s.bytes().len() => Ok(()),
            _ => Err(core::fmt::Error),
```

lkatalin: This is much cleaner.

PR #697, opened 2020-07-10
Proxy syscalls with sallyport

### LINKED ISSUE #313: Implement enarx-keep-sev

connorkuehl:
Closes #313 

Last patchset before I incorporate it on top of the backend framework

```
$ ./enarx-keep-sev/target/debug/enarx-keep-sev \
    --shim ./enarx-keep-sev-shim/target/x86_64-unknown-linux-musl/debug/enarx-keep-sev-shim \
    --code ./payload/target/x86_64-unknown-linux-musl/debug/payload
init_gdt
Hello World!
```

### DIFF
--- a/enarx-keep-sev/src/vm/syscall.rs
+++ b/enarx-keep-sev/src/vm/syscall.rs
@@ -22,10 +22,13 @@ mod param {
     pub const R9: usize = 5;
 }
 
+use param::RSI;
+
 /// Get the fixup table for a given syscall number, if any.
 #[inline]
 fn syscall_fixup_table(syscall: libc::c_long) -> Option<SyscallFixup> {
     match syscall {
+        libc::SYS_write => Some(&[RSI]),
         _ => None,
     }
 }
--- a/enarx-keep-sev/src/vm/syscall.rs
+++ b/enarx-keep-sev/src/vm/syscall.rs
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/// Some system call parameters contain pointers to userspace memory (such as the `write` syscall).
+/// These values are *guest physical addresses*, so they must be "fixed up" to be *host virtual
+/// addresses* so that the host can transparently proxy the syscall.
+///
+/// The `SyscallFixup` type contains a slice of integers that correspond to an index within the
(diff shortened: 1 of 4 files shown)


enarx-keep-sev/src/vm/syscall.rs
```diff
@@ -0,0 +1,43 @@
...
+    if let Some(fixups) = syscall_fixup_table(req.num.raw() as _) {
+        for fixup in fixups {
+            let guest = req.arg[*fixup];
+            let host = guest.raw() + (offset as usize);
+            req.arg[*fixup] = memory::Register::from_raw(host);
+        }
+    }
+
+    unsafe { req.syscall() }
+}
```
npmccallum, line 46: Presuming that you have newtypes for `PhysAddr` (meaning guest phys addr), why don't we make `sallyport::Block` take `T`? Then we can do:

```rust
impl From<sallyport::Request<PhysAddr>> for sallyport::Request<usize> {}
```

Thoughts?

connorkuehl: I think this would require moving the `offset` parameter into global scope which may be harder to track when we move to the `Thread` abstraction (and eventually, having more than 1 of them).

connorkuehl: Furthermore, this code is going to change almost immediately once your backend-trait patches are merged.

connorkuehl:
Superceded by #722

PR #698, opened 2020-07-10
Create diagrams with cargo make

### LINKED ISSUE #487: Port the Makefile in docs/ to use 'cargo-make'
I think off the top of my head the preference would be for these tasks to be defined in the root `Makefile.toml` to keep things simple.

The `misc-diagrams` test script will need to be updated accordingly if it is not outright superceded by the `cargo-make' task that results from this issue.

ambaxter:
Closes #487

Signed-off-by: Adam Baxter <ambaxter@users.noreply.github.com>


ambaxter:
Added the check function to cargo make, also. It's not failing when it should. I'm asking about that. After that we should be able to finish this up.

ambaxter:
Ready for review.

docs/Makefile.toml
```diff
@@ -0,0 +1,90 @@
...
+  diagram_date = exec git log -1 --date=unix --format="%cd" -- "${file}.png"
+  diagram_date_out = trim ${diagram_date.stdout}
+  if less_than ${diagram_date_out} ${source_date_out}
+    echo "This diagram is out-of-date: ${file}. Please generate an updated diagram and commit it to include it with your pull request."
+    fail = set true
+  end
+end
+if ${fail}
+  assert_fail "Some diagrams are out of date"
+end
```
sagiegurari, line 74: you can replace these 3 lines with:

assert_false ${fail} "Some diagrams are out of date"

assert false validates the value is false. if its true it will fail and break the build.

docs/Makefile.toml
```diff
@@ -0,0 +1,90 @@
...
+
+[tasks.clean-diagrams]
+script_runner = "@duckscript"
+script = [
+'''
+files_array = split ${ALL_FILES} ";"
+for file in ${files_array}
+  if is_path_exists "${file}.png"
+    rm "${file}.png"
+  end
```
sagiegurari, line 86: no need for the 'if' condition there.
rm won't break if the path doesn't exist. it would just do nothing.

.tests/misc-diagrams
```diff
@@ -1,20 +1,4 @@
...
-	diagram_date=$(git log -1 --date=unix --format="%cd" -- "$diagram_dir$diagram")
-
-	if [[ $diagram_date -lt $source_date ]]; then
-		echo "This diagram is out-of-date: $diagram_dir$diagram. Please generate an"`
-		      `" updated diagram and commit it to include it with your pull request."
-		status=1
-	fi
-done
-exit $status
+cargo make --cwd "$diagram_dir" check-diagrams
```
sagiegurari, line 4: if i remember this script is invoked from another cargo-make flow.
so its cargo-make->script->cargo-make
might want to replace that to cargo-make->cargo-make which is really simply to do

```toml
[tasks.call-another-cargo-make]
command = "cargo"
args = ["make", "check-diagrams"]
cwd = "CHANGE TO RIGHT PATH"
```

ambaxter: I'm having an issues

I've forced the script to fail. If I call the makefile from the docs directory it exits with an error as expected.
From the main makefile the build doesn't fail. 

The main task looks like:
```
command = "cargo"
args = ["make", "check-diagrams"]
cwd = "${ENARX_DOCS_DIR}"
```

```
[peaches@deepwell enarx]$ cargo make misc-diagrams
[cargo-make] INFO - cargo make 0.32.0
[cargo-make] INFO - Project: enarx
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: misc-diagrams
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: misc-diagrams
[cargo-make] INFO - Execute Command: "cargo" "make" "check-diagrams"
[cargo-make][1] INFO - Project: enarx
[cargo-make][1] INFO - Build File: Makefile.toml
[cargo-make][1] INFO - Task: check-diagrams
[cargo-make][1] INFO - Profile: development
[cargo-make][1] INFO - Running Task: workspace
[cargo-make][1] INFO - Build Done in 0 seconds.
[cargo-make] INFO - Build Done in 0 seconds.
```

ambaxter: And oddly enough the automatic testing found errors where I expected them to be. Weird

sagiegurari: oh, its getting the workspace emulation env vars from parent process so it tries to run it as a workspace.
such a complex setup.
you can set in env block of your new makefile workspace emulation false.

ambaxter: Ok! It works as expected now. Thank you!

ambaxter:
@connorkuehl Done

PR #706, opened 2020-07-13
Add Cursor to sallyport Block

### LINKED ISSUE #705: Add a cursor to sallyport Block
The cursor contains a method to sequentially allocate chunks of the Block of a specified size; it also contains a pointer to the next allocatable chunk of the Block and and error type passed in at instantiation.

TODO:
- [x] Implement Cursor in sallyport
- [x] Cursor field should not be an `Option`
- [x] `alloc()` should be `unsafe` or initialize the returned array
- [x] Cursor should store an error value to return in case of `alloc()` failure
- [x] `sallyport::Block` should have a function `fn cursor<'a, T>(err: T) -> Cursor<'a, T>` that replaces `Cursor::new()`
- [ ] `Block.buf` should become a private field
- [x] It should have tests

First referenced in comments on #680 and #682 .

lkatalin:
Resolves #705 

@npmccallum This doesn't have tests yet. But I am wondering if it's in the general direction of what you're thinking. Also: making the `Block.buf` field private would mean we can't check whether the input / output is longer than the available buffer in the `handler`.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -286,19 +286,19 @@ impl<'a> Handler<'a> {
     pub fn write(&mut self) -> u64 {
         self.trace("write", 3);
 
-        let input = unsafe { self.aex.gpr.rsi.as_slice(self.aex.gpr.rdx.into()) };
-
-        if input.len() > self.block.buf.len() {
-            return -libc::EMSGSIZE as u64;
-        }
+        let input: &[u8] = unsafe { self.aex.gpr.rsi.as_slice(self.aex.gpr.rdx.into()) };
 
         // Copy the encrypted input into unencrypted memory.
-        self.block.buf[..input.len()].copy_from_slice(input);
+        let mut cursor = self.block.cursor();
+        let untrusted = match cursor.copy_slice(input).or(Err(libc::EMSGSIZE)) {
+            Ok(slice) => slice,
+            Err(e) => return -e as u64,
+        };
+
+        let req = request!(libc::SYS_write => self.aex.gpr.rdi, untrusted, input.len());
 
         // Do the syscall; replace encrypted memory with unencrypted memory.
-        let res = unsafe {
-            self.proxy(request!(libc::SYS_write => self.aex.gpr.rdi, &self.block.buf, input.len()))
-        };
+        let res = unsafe { self.proxy(req) };
 
         match res {
             Ok(res) => {
--- a/sallyport/src/lib.rs
+++ b/sallyport/src/lib.rs
@@ -171,6 +171,46 @@ impl Block {
     pub const fn buf_capacity() -> usize {
         Page::size() - size_of::<Message>()
     }
+
+    /// Returns a Cursor for the Block
+    pub fn cursor<'a>(&'a mut self) -> Cursor<'a> {
+        Cursor(&mut self.buf)
+    }
+}
+
+/// Helper for allocation of untrusted memory in a Block.
+pub struct Cursor<'a>(&'a mut [u8]);
+
+impl<'a> Cursor<'a> {
+    /// Allocates an array, containing count number of T items. The result is uninitialized.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it returns an uninitialized array.
+    pub unsafe fn alloc<T>(&'a mut self, count: usize) -> Result<&'a mut [T], ()> {
+        let mid = {
+            let (padding, data, _) = self.0.align_to_mut::<T>();
+
+            if data.len() < count {
+                return Err(());
+            }
+
+            padding.len() + size_of::<T>() * count
+        };
+
+        let (data, next) = self.0.split_at_mut(mid);
+
+        self.0 = next;
+
+        Ok(data.align_to_mut::<T>().1)
+    }
+
+    /// Copies data from a value into a slice using self.alloc().
+    pub fn copy_slice<T: Copy>(&'a mut self, value: &[T]) -> Result<&'a [T], ()> {
+        let slice: &mut [T] = unsafe { self.alloc(value.len())? };
+        slice.copy_from_slice(value);
+        return Ok(slice);
+    }
 }
 
 #[cfg(test)]
@@ -249,5 +289,20 @@ mod tests {
         assert_eq!(req.arg[4], Register::<usize>::from(0));
         assert_eq!(req.arg[5], Register::<usize>::from(0));
         assert_eq!(req.arg[6], Register::<usize>::from(0));
+    }
+
+    #[test]
+    fn cursor() {
+        let mut block = Block::default();
+
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(4096usize) }, Err(()));
+
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(42usize) }.unwrap().len(), 42);
+
+        let mut c = block.cursor();
+        let slice = c.copy_slice(&[87, 2, 3]).unwrap();
+        assert_eq!(&slice, &[87, 2, 3]);
     }
 }


sallyport/src/lib.rs
```diff
@@ -171,6 +171,41 @@ impl Block {
...
+}
+
+/// Helper for allocation of untrusted memory in a Block.
+pub struct Cursor<'a, E> {
+    ptr: &'a mut [u8],
+    err: E,
+}
+
+impl<'a, E> Cursor<'a, E> {
+    /// Allocates a `count`-sized chunk of a buffer for use. And updates the buffer's pointer.
```
npmccallum, line 191: This documentation is incorrect. It allocates an array, containing `count` number of T items. The result is uninitialized. This needs to be documented in a `# Safety` section.

npmccallum:
> Also: making the `Block.buf` field private would mean we can't check whether the input / output is longer than the available buffer in the `handler`.

This check happens in `allocate()` already.

npmccallum:
I've changed my mind on the error. Let's revert to the tuple.

lkatalin:
@npmccallum I'm currently having these problems with this PR:

1. I'm not sure if the `cursor.copy_slice()` function works correctly. The test I have currently checks the return value from that function, and it passes. However, I had another test that actually checks the contents of `block.buf` and it would seem that only the first value from the slice was copied into `buf`. I'm not sure how that's possible since the return value should `==` the contents of `block.buf`. Maybe the test itself was incorrect?

2. Making the `block.buf` field private is proving to be difficult. For example, trying to call the syscall on a `cursor` instead of a `block.buf` in the handler results in borrow checker errors again. Also, parts of the `sev` crates would have to be updated as well.

sallyport/src/lib.rs
```diff
@@ -171,6 +171,49 @@ impl Block {
...
+/// Helper for allocation of untrusted memory in a Block.
+pub struct Cursor<'a>(&'a mut [u8]);
+
+impl<'a> Cursor<'a> {
+    /// Allocates an array, containing count number of T items. The result is uninitialized.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it returns an uninitialized array.
+    pub unsafe fn alloc<T>(&'a mut self, count: usize) -> Result<&'a mut [T], libc::c_int> {
```
npmccallum, line 190:
```suggestion
    pub unsafe fn alloc<T>(&'a mut self, count: usize) -> Result<&'a mut [T], ()> {
```

sallyport/src/lib.rs
```diff
@@ -171,6 +171,49 @@ impl Block {
...
+        let mid = {
+            let (padding, data, _) = self.0.align_to_mut::<T>();
+
+            if data.len() < count {
+                return Err(libc::EMSGSIZE);
```
npmccallum, line 195:
```suggestion
                return Err(());
```

sallyport/src/lib.rs
npmccallum: Callers may want different errors in different contexts. So we'll just let the caller do `Result::or()`.

sallyport/src/lib.rs
```diff
@@ -250,4 +293,17 @@ mod tests {
         assert_eq!(req.arg[5], Register::<usize>::from(0));
         assert_eq!(req.arg[6], Register::<usize>::from(0));
     }
+
+    #[test]
+    fn cursor() -> Result<(), libc::c_int> {
```
npmccallum, line 298:
```suggestion
    fn cursor() {
```

sallyport/src/lib.rs
```diff
@@ -250,4 +293,17 @@ mod tests {
...
+        let mut block = Block::default();
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(4096usize) }, Err(libc::EMSGSIZE));
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(42usize) }.unwrap().len(), 42);
+        let mut c = block.cursor();
+        let slice = c.copy_slice(&[87, 2, 3])?;
+        assert_eq!(&slice, &[87, 2, 3]);
+        Ok(())
```
npmccallum, line 307:
```suggestion
```

sallyport/src/lib.rs
```diff
@@ -250,4 +293,17 @@ mod tests {
...
+    #[test]
+    fn cursor() -> Result<(), libc::c_int> {
+        let mut block = Block::default();
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(4096usize) }, Err(libc::EMSGSIZE));
+        let mut c = block.cursor();
+        assert_eq!(unsafe { c.alloc::<usize>(42usize) }.unwrap().len(), 42);
+        let mut c = block.cursor();
+        let slice = c.copy_slice(&[87, 2, 3])?;
+        assert_eq!(&slice, &[87, 2, 3]);
```
npmccallum, line 306:
```suggestion
        let mut block = Block::default();

        let mut c = block.cursor();
        assert_eq!(unsafe { c.alloc::<usize>(4096usize) }, Err(libc::EMSGSIZE));

        let mut c = block.cursor();
        assert_eq!(unsafe { c.alloc::<usize>(42usize) }.unwrap().len(), 42);

        let mut c = block.cursor();
        assert_eq!(&c.copy_slice(&[87, 2, 3]).unwrap(), &[87, 2, 3]);
```

Vertical spacing in order to group logical units.

sallyport/src/lib.rs
```diff
@@ -171,6 +171,49 @@ impl Block {
...
+
+    /// Copies data from a value into a slice using self.alloc().
+    ///
+    /// # Safety
+    /// This function calls self.alloc(), which is unsafe.
+    pub fn copy_slice<T: Copy>(&'a mut self, value: &'a [T]) -> Result<&[T], ()> {
+        let slice: &mut [T] = unsafe { self.alloc(value.len())? };
+        slice.copy_from_slice(value);
+        return Ok(slice);
+    }
```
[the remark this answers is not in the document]
lkatalin: That makes sense. Thanks for the explanation.

sallyport/src/lib.rs
```diff
@@ -206,9 +206,6 @@ impl<'a> Cursor<'a> {
     }
 
     /// Copies data from a value into a slice using self.alloc().
-    ///
-    /// # Safety
-    /// This function calls self.alloc(), which is unsafe.
```
npmccallum: This needs to go into the previous commit.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -286,19 +286,19 @@ impl<'a> Handler<'a> {
...
-
-        if input.len() > self.block.buf.len() {
-            return -libc::EMSGSIZE as u64;
-        }
+        let input: &[u8] = unsafe { self.aex.gpr.rsi.as_slice(self.aex.gpr.rdx.into()) };
 
         // Copy the encrypted input into unencrypted memory.
-        self.block.buf[..input.len()].copy_from_slice(input);
+        let mut c = self.block.cursor();
+        let untrusted = match c.copy_slice(input).or(Err(libc::EMSGSIZE)) {
```
npmccallum, line 293:
```suggestion
        let mut cursor = self.block.cursor();
        let untrusted = match cursor.copy_slice(input).or(Err(libc::EMSGSIZE)) {
```

sallyport/src/lib.rs
```diff
@@ -171,6 +171,46 @@ impl Block {
...
+
+        let (data, next) = self.0.split_at_mut(mid);
+
+        self.0 = next;
+
+        Ok(data.align_to_mut::<T>().1)
+    }
+
+    /// Copies data from a value into a slice using self.alloc().
+    pub fn copy_slice<T: Copy>(&'a mut self, value: &'a [T]) -> Result<&[T], ()> {
```
npmccallum, line 209:
```suggestion
    pub fn copy_slice<T: Copy>(&'a mut self, value: &[T]) -> Result<&'a [T], ()> {
```

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -286,19 +286,19 @@ impl<'a> Handler<'a> {
...
+        let req = request!(libc::SYS_write => self.aex.gpr.rdi, untrusted, input.len());
 
         // Do the syscall; replace encrypted memory with unencrypted memory.
-        let res = unsafe {
-            self.proxy(request!(libc::SYS_write => self.aex.gpr.rdi, &self.block.buf, input.len()))
-        };
+        let res = unsafe { self.proxy(req) };
 
         match res {
             Ok(res) => {
```
npmccallum, line 304: I just realized this commit doesn't **actually** make `buf` private.

lkatalin: This explains why I didn't need to edit the `sev` crate when I pushed that time.

enarx-keep-sev-shim/src/hostcall.rs
```diff
@@ -136,14 +136,16 @@ impl<'a> HostCall<'a> {
...
-        let buf_address = Address::from(self.0.buf.as_mut_ptr());
-        let shim_virt_address = ShimVirtAddr::try_from(buf_address).map_err(|_| libc::EFAULT)?;
+        let mut cursor = self.0.cursor();
 
-        let phys = ShimPhysAddr::from(shim_virt_address);
-        let request = request!(libc::SYS_write => fd, phys, bytes.len());
+        let untrusted = match cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE)) {
+            Ok(slice) => slice,
+            Err(e) => return Err(e),
+        };
```
npmccallum, line 144: I suspect we can use `?` here instead of match. @connorkuehl ?

connorkuehl: @npmccallum @lkatalin correct

lkatalin: I'll make that change. If it looks reasonable otherwise, then that's great.

PR #707, opened 2020-07-13
Integration Tests

npmccallum:
This PR supersedes #633.


integration-tests/tests/syscall_exit.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains integration tests, including the exit syscall payload test.
```
connorkuehl, line 3: I think the module name is self-describing. The only tests shown here are ones for `exit()` calls, so instead of saying "including ..." we might as well just shorten it to something like "sys_exit tests"

integration-tests/tests/syscall_exit.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the exit syscall payload in the SGX keep using the SGX shim.
```
connorkuehl, line 11: Let's remove this comment. The TEE is chosen at compile-time.

npmccallum:
> What do you think about a convenience `cfg` like this:
> 
> ```rust
> #[cfg_attr(any(has_sgx, has_sev), has_tee)]
> ```
> 
> This way it's easy to conditionally enable code so long as we have _a_ TEE without caring which one and we don't have to copy and paste the list of `has_sgx, has_sev...` Code that needs to be more specific can still query `has_sgx` or whichever TEE it cares about.

As we merge into `enarx-keep` all this code will get removed. So let's leave it.

integration-tests/tests/syscall_exit.rs
npmccallum: Fixed.

integration-tests/tests/syscall_exit.rs
npmccallum: Fixed.

PR #709, opened 2020-07-14
Port sev-show to enarx-keep

### LINKED ISSUE #687: Move sev-show to enarx-keep
We are unifying into `enarx-keep` and need to relocate the `sev-show` code.

connorkuehl:
Closes #687


enarx-keep/src/backend/probe/x86_64.rs
```diff
@@ -0,0 +1,32 @@
...
+use std::convert::Into;
+
+pub struct CpuId {
+    pub name: &'static str,
+    pub leaf: u32,
+    pub subl: u32,
+    pub func: fn(CpuidResult) -> (bool, Option<String>),
+}
+
+impl Into<Datum> for &CpuId {
```
npmccallum, line 37:
```suggestion
impl From<&CpuId> for Datum {
```
Always implement these conversions using the `From` trait. There is a `core` blanket implementation that implements `Into` in the opposite direction for all types that implement `From`. So if you implement `From`, you get `Into` for free.

enarx-keep/src/backend/probe/x86_64.rs
```diff
@@ -4,19 +4,46 @@ use crate::backend::Datum;
 
 use std::arch::x86_64::{CpuidResult, __cpuid_count};
 use std::convert::Into;
+use std::mem::transmute;
+
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub enum Vendor {
+    Amd,
+    Intel,
+    Unknown,
```
npmccallum, line 13: Is `Option<Vendor>` better than `Vendor::Unknown`?

connorkuehl:
> I don't quite understand what you're trying to do with the `Fail vendor-specific...` patch. I like having a global list of vendors. But I'm not sure why it gets added to every `CpuId` instance...

@npmccallum I was running into an issue when I was testing this code on both Intel and AMD. The initial code would fail the cpuid test case if it exceeded the standard max leaf size, so all of the SEV-* checks which use extended CPUID leaves were spuriously failing.

I experimented with making sure the cpuid fell within the standard or extended ranges as advertised by the initial cpuid call, but then I noticed that it would _still_ show data from the highly platform specific cpuid calls, which is meaningless data for every platform _except_ the one you're actually running on. For example, on AMD, the output would still query SGX-specific feature leaves and output _something_, which is technically incorrect because those cpuid calls have different meanings for AMD.

I think in the initial `-show` implementations we arranged the tests in a hierarchical tree-like fashion to avoid disclosing irrelevant information like this.

They're added to each CpuId struct so that no other logic needs to be added to the backend `data()` methods. There's probably room for improvement here but my brain is fried today and I'll see what else I can come up with tomorrow.

enarx-keep/src/backend/probe/x86_64.rs
connorkuehl: Yeah, and I think in hindsight I shouldn't have even had `Unknown`. It's probably better for the `Vendor::get()` constructor function to return `Result<Vendor>` and do away with the `Unknown` variant since the `Err` variant of that result is more descriptive.

enarx-keep/src/backend/probe/x86_64.rs
connorkuehl: I think the blanket trait implementation is only occurs if I also `impl From<Datum> for &CpuId` but since going from CpuId to Datum is so "lossy" I felt only having a 1 direction CpuId -> Datum conversion was sensible.

npmccallum: @connorkuehl It doesn't work like that.

enarx-keep/src/backend/probe/x86_64.rs
```diff
@@ -3,30 +3,76 @@
...
+            info: None,
+            mesg: None,
+        }
+    }
 }
 
-impl Into<Datum> for &CpuId {
-    fn into(self) -> Datum {
-        let max = unsafe { __cpuid_count(0x00000000, 0x00000000) }.eax;
+impl TryInto<Datum> for &CpuId {
```
npmccallum, line 54: You need `TryFrom` here instead. You basically never impl `Into` or `TryInto` in Rust.

npmccallum: On second thought, don't make this a fallible conversion. Make it an infallible conversion and merge `fail()` into the `cpu vendor mismatch` case.

enarx-keep/src/backend/probe/x86_64.rs
connorkuehl: @npmccallum https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d82fce13eb8ccf617c7c6f8b1dd43ddd

connorkuehl: I was wrong. I see what you mean now.

connorkuehl: Addressed this in the latest revision

enarx-keep/src/backend/probe/x86_64.rs
connorkuehl: Addressed in latest revision

enarx-keep/src/backend/probe/x86_64.rs
connorkuehl: Addressed in latest revision

PR #710, opened 2020-07-14
Add a test for Write system call

jyotsna-penumaka:
Related: #627

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::time::Duration;
 use wait_timeout::ChildExt;
 
@@ -67,7 +68,14 @@ impl IntegrationTest {
         }
     }
 
-    pub fn run(self, timeout: u64, exit_status: i32) {
+    pub fn run(
+        self,
+        timeout: u64,
+        exit_status: i32,
+        stdin: impl AsRef<[u8]>,
+        stdout: impl AsRef<[u8]>,
+        stderr: impl AsRef<[u8]>,
+    ) {
         let seconds = Duration::from_secs(timeout);
 
         let mut cmd = Command::new(self.keep)
@@ -76,9 +84,17 @@ impl IntegrationTest {
             .arg(self.code)
             .arg("--shim")
             .arg(self.shim)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
             .spawn()
             .expect("failed to run the test");
 
+        cmd.stdin
+            .as_mut()
+            .unwrap()
+            .write_all(stdin.as_ref())
+            .expect("Failed to write stdin");
         let ecode = match cmd.wait_timeout(seconds).unwrap() {
             Some(status) => status.code(),
             None => {
@@ -86,7 +102,12 @@ impl IntegrationTest {
                 panic!("killed by watchdog!");
             }
         };
+        let output = cmd
+            .wait_with_output()
+            .expect("Failed to read stdout/stderr");
 
+        assert_eq!(stdout.as_ref(), output.stdout.as_slice());
+        assert_eq!(stderr.as_ref(), output.stderr.as_slice());
         assert_eq!(exit_status, ecode.unwrap());
     }
 }
--- a/integration-tests/tests/syscall_write.rs
+++ b/integration-tests/tests/syscall_write.rs
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains integration tests, including the exit syscall payload test.
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the write syscall payload in the SGX keep using the SGX shim.
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write_stdout() {
+    IntegrationTest::new("write_stdout").run(5, 0, "", "hello world\n", "");
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write_with_wrong_output() {
+    IntegrationTest::new("write_stdout").run(5, 0, "", "hello", "");
+}
+
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write_stderr() {
+    IntegrationTest::new("write_stderr").run(5, 0, "", "", "hello world\n");
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write_with_wrong_error() {
+    IntegrationTest::new("write_stderr").run(5, 0, "", "", "hello");
+}
--- a/integration-tests/src/bin/write_stderr.rs
+++ b/integration-tests/src/bin/write_stderr.rs
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::write;
+
+fn main() {
+    let buf = "hello world\n";
+    unsafe {
+        write(libc::STDERR_FILENO, buf.as_ptr() as *const _, buf.len());
+    }
+}
--- a/integration-tests/src/bin/write_stdout.rs
+++ b/integration-tests/src/bin/write_stdout.rs
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::write;
+
+fn main() {
+    let buf = "hello world\n";
+    unsafe {
+        write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, buf.len());
+    }
+}
--- a/integration-tests/tests/syscall_exit.rs
+++ b/integration-tests/tests/syscall_exit.rs
@@ -8,18 +8,18 @@ use common::IntegrationTest;
 #[test]
 #[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
 fn exit_zero() {
-    IntegrationTest::new("exit_zero").run(5, 0);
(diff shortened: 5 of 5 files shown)


integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        let ecode = match cmd.wait_timeout(seconds).unwrap() {
+            Some(status) => status.code(),
+            None => {
+                cmd.kill().unwrap();
+                panic!("killed by watchdog!");
+            }
+        };
+        let output = cmd.wait_with_output().expect("Failed to read stdout");
+        if stdout != "" {
+            assert_eq!(String::from_utf8_lossy(&output.stdout), stdout);
```
jyotsna-penumaka, line 92: I made a few changes to integration-tests/tests/common/mod.rs to read the input from stdin and compare the output to the expected output
I also tried `String::from_utf8_lossy(&output.stdout[1..(output_length-3)])` to get rid of the escape characters which get appended to the output due the the `print!` statement in the test for read system call, but then the test fails for write.

integration-tests/src/bin/write.rs
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use core::ffi::c_void;
+use libc::write;
+
```
lkatalin, line 5:
```suggestion
use std::os::unix::io::AsRawFd;
use std::io::stdout;

```

integration-tests/tests/common/mod.rs
lkatalin: @jyotsna-penumaka Do you have an example of the escape characters? Are they still there if you use `trim()`?

integration-tests/tests/common/mod.rs
```diff
@@ -87,6 +89,10 @@ impl IntegrationTest {
             }
         };
 
+        let output = cmd.wait_with_output().expect("Failed to read stdout");
+        if stdout != "" {
```
lkatalin, line 93: Instead of checking if the `stdout` arg is an empty string, it seems more "Rusty" to make it an `Option`. Alternatively you could use `is_empty()` on `stdout`.

integration-tests/src/bin/write.rs
npmccallum: I'd prefer a slightly different approach. See below.

integration-tests/src/bin/write.rs
```diff
@@ -0,0 +1,11 @@
...
+fn main() {
+    let buf = "hello world\n".as_ptr() as *const c_void;
```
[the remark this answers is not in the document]
npmccallum: +1

integration-tests/src/bin/write.rs
```diff
@@ -0,0 +1,11 @@
...
+    unsafe {
+        write(1, buf, 12);
```
npmccallum, line 9:
```suggestion
        write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, buf.len().into());
```

integration-tests/tests/common/mod.rs
```diff
@@ -67,7 +67,7 @@ impl IntegrationTest {
         }
     }
 
-    pub fn run(self, timeout: u64, exit_status: i32) {
+    pub fn run(self, timeout: u64, exit_status: i32, stdout: &str) {
```
npmccallum, line 70:
```suggestion
    pub fn run(self, timeout: u64, exit_status: i32, stdout: impl AsRef<[u8]>) {
```
The output from the `Command` below returns a `Vec<u8>`. This can be converted to a slice (`[u8]`) trivially. Therefore, we want the input to the test to be any type that can also be trivially converted to a slice. The good news is that this **includes** `&str`! So you can still just pass a `&str` to this function. The end result is that you don't have to convert the output slice to a string. See below.

integration-tests/tests/common/mod.rs
npmccallum: Let's just skip this check altogether. We can do an unconditional:

```rust
             assert_eq!(stdout.as_ref(), &output.stdout);
```

integration-tests/tests/common/mod.rs
npmccallum: Since we don't have to convert to a string anymore, we don't have to worry about escape characters.

integration-tests/src/bin/write.rs
lkatalin: I was looking for something like `libc::STDOUT_FILENO`! Thank you @npmccallum .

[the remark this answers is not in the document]
jyotsna-penumaka: @lkatalin do we have to use .into(), I am not sure what the conversion is for?
It seems to work fine when I test it without .into()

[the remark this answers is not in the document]
jyotsna-penumaka: @lkatalin thanks! :)

integration-tests/tests/common/mod.rs
```diff
@@ -76,6 +82,8 @@ impl IntegrationTest {
             .arg(self.code)
             .arg("--shim")
             .arg(self.shim)
+            .stdin(Stdio::inherit())
+            .stdout(Stdio::piped())
```
lkatalin, line 86: Based on the comments in the [read PR](https://github.com/enarx/enarx/pull/711), I think these should be `piped()` and there should be a `.stderr`.

jyotsna-penumaka: Thanks for letting me know :)

integration-tests/tests/common/mod.rs
```diff
@@ -76,17 +84,29 @@ impl IntegrationTest {
             .arg(self.code)
             .arg("--shim")
             .arg(self.shim)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
             .spawn()
             .expect("failed to run the test");
 
+        let input = cmd.stdin.as_mut().unwrap();
```
npmccallum, line 93:
```suggestion
        let input = cmd.stdin.unwrap();
```

I don't understand the purpose of the `.as_mut()` here...

npmccallum: Also, I think we can get rid of the input temporary variable.

integration-tests/tests/common/mod.rs
```diff
@@ -76,17 +84,29 @@ impl IntegrationTest {
...
                 cmd.kill().unwrap();
                 panic!("killed by watchdog!");
             }
         };
+        let output = cmd
+            .wait_with_output()
+            .expect("Failed to read stdout/stderr");
 
+        assert_eq!(stdout.as_ref().to_vec(), output.stdout);
+        assert_eq!(stderr.as_ref().to_vec(), output.stderr);
```
npmccallum, line 109:
```suggestion
        assert_eq!(stdout.as_ref(), output.stdout.as_ref());
        assert_eq!(stderr.as_ref(), output.stderr.as_ref());
```

We don't want to use `to_vec()` here. `to_vec()` will allocate memory. Instead, we express the existing `Vec` as a slice using `as_ref()`, which doesn't allocate memory.

integration-tests/tests/syscall_write.rs
```diff
@@ -0,0 +1,23 @@
...
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the write syscall payload in the SGX keep using the SGX shim.
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write() {
+    IntegrationTest::new("write").run(5, 0, "hello world\n", "hello world\n", "");
```
npmccallum, line 15: The test doesn't read the input. So don't provide it.

integration-tests/tests/syscall_write.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn write_with_wrong_output() {
+    IntegrationTest::new("write").run(5, 0, "hello", "hello", "");
```
npmccallum, line 22: The test doesn't read the input. So don't provide it.

integration-tests/src/bin/write.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::write;
+
+fn main() {
+    let buf = "hello world\n";
+    unsafe {
+        write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, buf.len());
+    }
+}
```
npmccallum, line 10: Don't forget to add a second test which does exactly the same thing on `libc::STDERR_FILENO`.

integration-tests/tests/common/mod.rs
jyotsna-penumaka: You are right, we don't need the temporary variable, I got rid of it now.
The value stored in cmd is getting moved and is resulting in an error when I try getting the ecode without as_mut.
I am not entirely sure how creating a mutable reference with as_mut() is solving the problem

integration-tests/tests/common/mod.rs
jyotsna-penumaka: Yeah it makes sense to not allocate memory.
I am using as_slice to convert it from vec --> slice
 as_ref() is resulting in an error

integration-tests/tests/common/mod.rs
```diff
@@ -76,17 +84,30 @@ impl IntegrationTest {
...
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
             .spawn()
             .expect("failed to run the test");
 
+        cmd.stdin
+            .as_mut()
+            .unwrap()
+            .write_all(stdin.as_ref())
+            .expect("Failed to write stdin");
```
lkatalin, line 97: Is this being used for the `write()` test? If so, I'm not understanding the test correctly.

jyotsna-penumaka: No wouldn't need stdin for the write test, but we would need these changes for read.
Should I remove them?

lkatalin: I would probably opt for only running that if there is a `stdin` that isn't empty. @npmccallum were you suggesting that this should be run every time even without a `stdin` argument?

npmccallum: I'm okay to leave this. `write()` with a zero-length input is a noop.

PR #711, opened 2020-07-14
Add a test for Read system call

jyotsna-penumaka:
Related: #627

### DIFF
--- a/integration-tests/src/bin/read.rs
+++ b/integration-tests/src/bin/read.rs
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::{read, write, STDIN_FILENO as IN, STDOUT_FILENO as OUT};
+use std::cmp::min;
+use std::process::exit;
+
+fn main() {
+    let mut buf = [0u8; 16];
+    let mut len = 1;
+
+    loop {
+        match unsafe { read(IN, buf.as_mut_ptr() as *mut _, len) } {
+            sz if sz > 0 => unsafe {
+                write(OUT, buf.as_ptr() as *const _, sz as usize);
+            },
+            sz if sz < 0 => exit(1),
+            _ => break,
+        }
+
+        len = min(len * 2, buf.len());
+    }
+}
--- a/integration-tests/tests/syscall_read.rs
+++ b/integration-tests/tests/syscall_read.rs
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate contains integration tests, including the exit syscall payload test.
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the read syscall payload in the SGX keep using the SGX shim.
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn read() {
+    IntegrationTest::new("read").run(5, 0, "hello world\n", "hello world\n", "");
+}
+
+#[test]
+#[should_panic]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn read_with_wrong_output() {
+    IntegrationTest::new("read").run(5, 0, "hello", "h", "");
+}


integration-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        let ecode = match cmd.wait_timeout(seconds).unwrap() {
+            Some(status) => status.code(),
+            None => {
+                cmd.kill().unwrap();
+                panic!("killed by watchdog!");
+            }
+        };
+        let output = cmd.wait_with_output().expect("Failed to read stdout");
+        if stdout != "" {
+            assert_eq!(String::from_utf8_lossy(&output.stdout), stdout);
```
jyotsna-penumaka, line 92: I made a few changes to integration-tests/tests/common/mod.rs to read the input from stdin and compare the output to the expected output 
I also tried `String::from_utf8_lossy(&output.stdout[1..(output_length-3)])` to get rid of the escape characters which get appended to the output due the the `print!` statement in the test for read system call, but then the test fails for write.

integration-tests/src/bin/read.rs
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use core::ffi::c_void;
+use libc::read;
+
+fn main() {
+    let buf = [0u8; 12];
+    unsafe {
+        read(0, buf.as_ptr() as *mut c_void, 12);
```
lkatalin, line 9: Please use the comments on the write syscall test to update this to not use hardcoded values.

integration-tests/tests/common/mod.rs
```diff
@@ -87,6 +89,13 @@ impl IntegrationTest {
             }
         };
 
+        let output = cmd.wait_with_output().expect("Failed to read stdout");
+        if stdout != "" {
+            let output_length = output.stdout.len();
+            println!("length is : {:?}", output_length);
```
lkatalin, line 95: You can consolidate these two lines.

integration-tests/tests/syscall_read.rs
```diff
@@ -0,0 +1,23 @@
...
+
+//! This crate contains integration tests, including the exit syscall payload test.
+
+#![deny(clippy::all)]
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the write syscall payload in the SGX keep using the SGX shim.
```
lkatalin, line 11: I think you meant "read". : )

integration-tests/tests/common/mod.rs
```diff
@@ -76,6 +76,8 @@ impl IntegrationTest {
             .arg(self.code)
             .arg("--shim")
             .arg(self.shim)
+            .stdin(Stdio::inherit())
+            .stdout(Stdio::piped())
```
npmccallum, line 80:
```suggestion
            .stdin(Stdio::piped())
            .stdout(Stdio::piped())
            .stderr(Stdio::piped())
```

integration-tests/tests/common/mod.rs
```diff
@@ -87,6 +89,13 @@ impl IntegrationTest {
             }
         };
 
+        let output = cmd.wait_with_output().expect("Failed to read stdout");
```
npmccallum, line 92:
```suggestion
        cmd.stdin.unwrap().write_all(stdin.as_ref()).expect("Failed to write stdin");
        let output = cmd.wait_with_output().expect("Failed to read stdout/stderr");
```

integration-tests/src/bin/read.rs
```diff
@@ -0,0 +1,12 @@
...
+    }
+    print!("{:?}", String::from_utf8_lossy(&buf));
+}
```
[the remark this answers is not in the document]
lkatalin: @jyotsna-penumaka I think what @npmccallum was looking for here was to read the input byte by byte (and output the same byte) until there are no more bytes (`EOF`), but I see this as just reading one byte. You would want a loop to keep reading input byte by byte and writing out the bytes. @npmccallum Let me know if I'm misunderstanding.

integration-tests/tests/syscall_read.rs
```diff
@@ -0,0 +1,23 @@
...
+#![deny(missing_docs)]
+
+mod common;
+use common::IntegrationTest;
+
+/// This test runs the read syscall payload in the SGX keep using the SGX shim.
+#[test]
+#[cfg_attr(not(any(has_sgx, has_sev)), ignore)]
+fn read() {
+    IntegrationTest::new("read").run(5, 0, "hello world\n", "h", "");
```
lkatalin, line 15: If my comment above is accurate, the "h" here would turn into the full "hello world\n."

npmccallum: It is, indeed, accurate.

lkatalin:
Here is the approach we decided on:
We could allocate a buffer of 16, but pass 1 to read() for the length. Then we could double the length on each loop until we reach 16.

integration-tests/src/bin/read.rs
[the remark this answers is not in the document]
jyotsna-penumaka: @npmccallum I used break instead of exit system call, as the output would be 1 in that latter case, even though the input from stdin was read correctly.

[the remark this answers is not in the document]
npmccallum: @jyotsna-penumaka Returning from `main()` is the same as `exit(0)`.

integration-tests/src/bin/read.rs
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::{read, write};
+
+fn main() {
+    let mut buf = [0u8; 16];
+    for i in (0..).map(|x| 1 << x) {
```
npmccallum, line 7: If `i` is `i32`, this will panic on the 33rd iteration. We probably don't want that. We probably don't want a for loop here since `i` or `x` will eventually grow too large and panic for any large enough input.

integration-tests/src/bin/read.rs
```diff
@@ -0,0 +1,17 @@
...
+        let len = core::cmp::min(i, buf.len());
+        unsafe {
+            let sz = read(libc::STDIN_FILENO, buf.as_mut_ptr() as *mut _, len);
+            write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, sz as usize);
+            if sz < len as isize {
+                break;
+            }
+        }
```
npmccallum, line 15:
```suggestion
        let sz = unsafe {
            let sz = read(libc::STDIN_FILENO, buf.as_mut_ptr() as *mut _, len);
            write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, sz as usize);
            sz
        };
        
        if sz < len as isize {
            break;
        }
```

It is bad practice to use unsafe blocks larger than necessary.

integration-tests/src/bin/read.rs
```diff
@@ -0,0 +1,25 @@
...
+        unsafe {
+            write(libc::STDOUT_FILENO, buf.as_ptr() as *const _, sz as usize);
+        }
+        if sz < len as isize {
+            break;
+        }
+        i = 2 * i;
+    }
+}
```
npmccallum, line 25:
```suggestion
// SPDX-License-Identifier: Apache-2.0

use libc::{read, write, STDIN_FILENO as IN, STDOUT_FILENO as OUT};
use std::process::exit;
use std::cmp::min;

fn main() {
    let mut buf = [0u8; 16];
    let mut len = 1;

    loop {
        match unsafe { read(IN, buf.as_mut_ptr() as *mut _, len) } {
            sz if sz > 0 => unsafe { write(OUT, buf.as_ptr() as *const _, sz as usize); },
            sz if sz < 0 => exit(1),
            _ => break,
        }

        len = min(len * 2, buf.len());
    }
}
```

Take a few minutes to understand precisely why this works.

If `sz < len`, then we have EOF. But instead of immediately exiting, we run the loop one more time. On the next loop, `sz == 0` and we `break`.

jyotsna-penumaka: @npmccallum 
I took some time to understand how you used the match statement with place expressions. 
I get it now :)
Is there a reason why we are not libc::exit in case there is an error?

npmccallum: `std::process::exit()` is the Rust way. But it probably just calls the `exit()` syscall under the covers.

PR #712, opened 2020-07-14
Add KVM backend detection (kvm-show)

connorkuehl:
I intend to port the enarx-keep-sev "bare" VMM implementation onto this backend. The actual SEV-enabled VMM implementation will extend this backend with the SEV API, of course.

I've left this unlabeled since it's not _technically_ SEV but it is the groundwork for that.

### DIFF
--- a/enarx-keep/src/backend/kvm.rs
+++ b/enarx-keep/src/backend/kvm.rs
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::backend::{self, Datum, Keep};
+use crate::binary::Component;
+
+use kvm_ioctls::Kvm;
+
+use std::io::Result;
+use std::path::PathBuf;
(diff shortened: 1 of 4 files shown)


enarx-keep/src/backend/kvm.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use std::fs::File;
+use std::io::Result;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+
+fn dev_kvm() -> Datum {
+    let version = Kvm::new().and_then(|kvm| Ok(kvm.get_api_version()));
+    let (pass, info) = match version {
+        Ok(v) => (v == 12, Some("/dev/kvm".into())),
```
npmccallum, line 16: Can we split this datum into two data: one for `/dev/kvm` and one for the kvm version? This can help with debugging later.

PR #722, opened 2020-07-16
Port KVM backend from enarx-keep-sev to enarx-keep

### LINKED ISSUE #313: Implement enarx-keep-sev

### LINKED ISSUE #688: Move enarx-keep-sev into enarx-keep
We are unifying into `enarx-keep` and need to relocate the `enarx-keep-sev` code.

connorkuehl:
Closes #688 
Closes #313


enarx-keep/Cargo.toml
```diff
@@ -8,13 +8,17 @@ edition = "2018"
...
+kvm-bindings = "0.2.0"
 kvm-ioctls = "0.5.0"
 memory = { path = "../memory" }
 mmap = { path = "../mmap" }
 structopt = "0.3"
 colorful = "0.2"
 goblin = "0.2"
 libc = "0.2"
+units = { path = "../units" }
+x86_64 = { version = "0.11.0", default-features = false, features = ["stable"] }
```
npmccallum, line 24: I've generally tried to keep path vs version crates segregated. Could you follow this?

enarx-keep/src/backend/kvm.rs
```diff
@@ -42,10 +46,27 @@ impl backend::Backend for Backend {
...
+            ".",
+            "enarx-keep-sev-shim",
+            "target",
+            "x86_64-unknown-linux-musl",
+            "debug",
+            "enarx-keep-sev-shim",
+        ];
+
+        let path = components.iter().fold(PathBuf::new(), |p, c| p.join(c));
+        Ok(path)
```
npmccallum, line 59: We probably want this relative to `env!("CARGO_MANIFEST_DIR")` for now. Also, I'd prefer a long list of wrapped `.join()` calls because it is more obvious what is happening.

enarx-keep/src/backend/kvm/vm.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl Keep for VirtualMachine {
+    fn add_thread(self: Arc<Self>) -> Result<Box<dyn Thread>> {
+        // The startup CPU requires special setup that is cleanly
+        // done during the keep build phase. The VirtualMachine struct
+        // stashes that startup thread there and it is the first one
+        // we give out when a thread is requested. After this deferred
+        // thread is given out, future thread requests are allocated
+        // dynamically.
+        let thread = self.deferred_cpu.replace(None);
```
npmccallum, line 36:
```suggestion
        let thread = self.deferred_cpu.take();
```

enarx-keep/src/backend/kvm/vm.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use std::cell::RefCell;
+use std::io::Result;
+use std::sync::Arc;
+
+pub struct VirtualMachine {
+    _kvm: Kvm,
+    _fd: VmFd,
+    _address_space: Region,
+    deferred_cpu: RefCell<Option<Cpu>>,
```
npmccallum, line 25: Is `RefCell` the right thing here?

enarx-keep/src/backend/kvm/vm/cpu.rs
```diff
@@ -0,0 +1,90 @@
...
+use x86_64::registers::control::{Cr0Flags, Cr4Flags};
+use x86_64::registers::model_specific::EferFlags;
+use x86_64::PhysAddr;
+
+use std::convert::AsMut;
+use std::io::{Error, ErrorKind, Result};
+
+pub struct Cpu {
+    fd: VcpuFd,
+    sallyport: SallyPort,
```
npmccallum, line 19: I'm surprised that `Cpu` doesn't keep an `Arc<Mutex<VirtualMachine>>`.

npmccallum: Why not store the `sallyport::Block` directly here?

enarx-keep/Cargo.toml
connorkuehl: Of course. :joy: I just try to keep things alphabetized. I'll queue this up for the next revision.

enarx-keep/src/backend/kvm/vm.rs
connorkuehl: I think that method is nightly-only.

enarx-keep/src/backend/kvm/vm.rs
connorkuehl: `Keep::add_thread` doesn't provide a mutable reference to the keep (and by extension: this field). RefCell was the only trick I could think of for the interior mutability here.

enarx-keep/src/backend/kvm/vm/cpu.rs
connorkuehl: > I'm surprised that `Cpu` doesn't keep an `Arc<Mutex<VirtualMachine>>`.

I think it will need to grow to do so when we start hotplugging vCPUs.

> Why not store the `sallyport::Block` directly here?

I think this leaves us some flexibility in the event the shim decides to relocate the shared pages alongside itself but this is not a use case that any of us have expressed yet. Aside from that, there's no technical benefit from the way I did it.

enarx-keep/src/backend/kvm/vm.rs
connorkuehl: @npmccallum Do you want this PR to update that interface so that `Keep::add_thread` provides mutability? I chose not to initially because I didn't want to make any invasive changes for something that may be just a 1-off for my case. I'll be the first to admit that my foresight is probably more shortsighted here.

npmccallum: Yes, I think that's the better option. I think the best explanation for why it isn't mutable is an oversight on my part.

We probably need to `Arc<Mutex<dyn Keep>>` so that any thread can handle the new thread creation process via keeping a reference to the keep and locking it when necessary.

npmccallum: Mutex => RwLock

connorkuehl: Addressed in latest revision in a separate commit. The compiler didn't let me use `Arc<RwLock<dyn Keep>>` as a receiver like the trait function currently does with `Arc<dyn Keep>`. In my research I found that many solutions present an interface that looks single threaded and relies on Rust's compile-time checks to ensure the caller wraps their use in types that are thread safe if the data is sent across to another thread, so that's the approach that I took here as well.

enarx-keep/src/backend/kvm/vm/syscall.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// physical addresses to get the host virtual address it
+    /// corresponds to.
+    pub offset: u64,
+}
+
+impl AsMut<sallyport::Block> for SallyPort {
+    fn as_mut(&mut self) -> &mut sallyport::Block {
+        let blk = unsafe { self.addr.as_mut_ptr::<sallyport::Block>().as_mut().unwrap() };
+
+        if let Some(fixups) = get_fixup_table(unsafe { u64::from(blk.msg.req.num) }) {
```
haraldh, line 28: What do you think of letting the shim doing the fixup when constructing the request? Of course, it would need the translation information first.

connorkuehl: @haraldh I don't see any issues with this. The host virtual offset can be communicated via the bootinfo struct.

connorkuehl: The only possible use case that would complicate this slightly is if the host VMM has to `mremap` the address space. That would cause a coherency issue that the host VMM would have to preempt the guest to notify it of. In that case it's a whole bag of worms because the guest would have to revalidate all of its pages, so I don't know if that's a use case we are interested in.

enarx-keep/src/backend/kvm/vm/builder.rs
```diff
@@ -258,25 +261,36 @@ impl Builder<Code> {
...
+        let sallyport = SallyPort {
+            addr: VirtAddr::new(&host_setup.shared_page as *const _ as _),
+            offset: addr_space.start.as_u64(),
+        };
 
-        let cr3 = &guest_setup.pml4t as *const _ as _;
-        vm.add_vcpu(vcpu, self.data.shim_entry.unwrap(), cr3)?;
+        let entry = self.data.shim_entry.unwrap();
+        let cpu0 =
+            move |keep: Arc<RwLock<VirtualMachine>>| Cpu::new(vcpu, keep, sallyport, entry, cr3);
```
haraldh, line 286: interesting technique :+1:

enarx-keep/src/backend/kvm.rs
```diff
@@ -42,10 +47,33 @@ impl backend::Backend for Backend {
...
 
     fn shim(&self) -> Result<PathBuf> {
-        unimplemented!()
+        #[cfg(debug_assertions)]
+        const PROFILE: &str = "debug";
+
+        #[cfg(not(debug_assertions))]
+        const PROFILE: &str = "release";
+
+        let path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
haraldh, line 56: I would expect some debt issue to make this all configurable, instead of hardcoding the paths from the compilation layout.

PR #725, opened 2020-07-16
Use sallyport::Cursor for buf allocations

### LINKED ISSUE #695: Return EMSGSIZE when a write() call exceeds the sallyport block size
The current approach is incompatible with writing to a datagram socket and mending it by coalescing writes from the shim on the VMM side is currently not worth the trouble right now.

### LINKED ISSUE #724: Make sallyport::Block.buf private
`sallyport::Cursor` is used for writing to the `buf` as of a9366a5c74109800058dffff589be21ec00a0059

connorkuehl:
Closes #724 
Closes #695

### DIFF
--- a/enarx-keep-sev-shim/src/hostcall.rs
+++ b/enarx-keep-sev-shim/src/hostcall.rs
@@ -134,16 +134,16 @@ impl<'a> HostCall<'a> {
     ///
     /// The parameters returned can't be trusted.
     pub unsafe fn write(&mut self, fd: usize, bytes: &[u8]) -> Result<libc::c_int, libc::c_int> {
-        let bytes = &bytes[..core::cmp::min(bytes.len(), Block::buf_capacity())];
+        let mut cursor = self.0.cursor();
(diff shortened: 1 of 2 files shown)


enarx-keep-sev-shim/src/hostcall.rs
```diff
@@ -136,14 +136,16 @@ impl<'a> HostCall<'a> {
     pub unsafe fn write(&mut self, fd: usize, bytes: &[u8]) -> Result<libc::c_int, libc::c_int> {
         let bytes = &bytes[..core::cmp::min(bytes.len(), Block::buf_capacity())];
 
-        let buf_address = Address::from(self.0.buf.as_mut_ptr());
+        let mut cursor = self.0.cursor();
+        let buf = cursor.copy_slice(bytes).or(Err(libc::ENOMEM))?;
```
lkatalin, line 140: Is there a reason we prefer `ENOMEM` here but `EMSGSIZE` in SGX?

npmccallum: Good catch! This needs to be `EMSGSIZE`.

PR #735, opened 2020-07-20
Various cleanups

npmccallum:
This work will probably conflict with @lkatalin 's work. So we should merge her PRs first and then rebase.

### DIFF
--- a/enarx-keep-sgx-shim/src/handler.rs
+++ b/enarx-keep-sgx-shim/src/handler.rs
@@ -15,7 +15,7 @@ const TRACE: bool = false;
 // arch_prctl syscalls not available in the libc crate as of version 0.2.69
 enumerate::enumerate! {
     #[derive(Copy, Clone)]
-    enum ArchPrctlTask: u64 {
+    enum ArchPrctlTask: usize {
         ArchSetGs = 0x1001,
         ArchSetFs = 0x1002,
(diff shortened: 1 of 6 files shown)


enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -321,12 +321,13 @@ impl<'a> Handler<'a> {
...
         }
 
-        let utsname = unsafe { &mut *(self.aex.gpr.rdi.raw() as *mut libc::utsname) };
-        fill(&mut utsname.sysname, "Linux");
-        fill(&mut utsname.nodename, "localhost.localdomain");
-        fill(&mut utsname.release, "5.6.0");
-        fill(&mut utsname.version, "#1");
-        fill(&mut utsname.machine, "x86_64");
+        let u: *mut libc::utsname = self.aex.gpr.rdi.into();
+        let u = unsafe { &mut *u };
```
lkatalin, line 325: This is interesting.

npmccallum: What's interesting? We have `Into` conversions for pointers from registers. But registers can't tell us lifetimes. Therefore, it is unsafe to convert a pointer to a reference because this requires a pointer dereference.

lkatalin: Thanks!

PR #742, opened 2020-07-21
sev-shim: add dbg!() macro for faster debugging

haraldh:


enarx-keep-sev-shim/src/print.rs
```diff
@@ -113,3 +113,28 @@ macro_rules! eprintln {
...
+        match $val {
+            tmp => {
+                $crate::eprintln!("[{}:{}] {} = {:#?}",
+                    file!(), line!(), stringify!($val), &tmp);
+                tmp
+            }
+        }
+    };
+    // Trailing comma with single argument is ignored
+    ($val:expr,) => { $crate::dbg!($val) };
```
npmccallum, line 136:
```suggestion
    ($val:expr $(,)?) => {
        // Use of `match` here is intentional because it affects the lifetimes
        // of temporaries - https://stackoverflow.com/a/48732525/1063961
        match $val {
            tmp => {
                $crate::eprintln!("[{}:{}] {} = {:#?}",
                    file!(), line!(), stringify!($val), &tmp);
                tmp
            }
        }
    };
```

haraldh: I just copied the macro from rust-lang `std` ...

npmccallum: Which was probably first written before `$(,)?` was legal. :)

haraldh: > Which was probably first written before `$(,)?` was legal. :)

No, because the next match arm is `($($val:expr),+ $(,)?) => {`

haraldh: anyway, your suggestions seems to work

PR #746, opened 2020-07-21
Consolidate multiple sgx util crates

### LINKED ISSUE #519: Consolidate sgx crates
We should consolidate the following crates:

- [x] rename sgx-types => sgx
- [x] merge sgx-crypto => sgx
- [x] merge iocuddle-sgx => sgx
- [x] merge attestation-types => sgx
- [ ] ~~merge intel-types => upstream x86_64 as outlined in #442~~

SEV crates are undergoing similar change, detailed here: #518 

Edit: This would allow us to easily merge the `testing` crate into the remaining `sgx` crate as detailed in #592 .

EDIT: As discussed during planning, upstreaming `intel-types` into `x86_64` will be treated as a separate issue #442, while this one will pertain mainly consolidates the crates in the workspace.

### LINKED ISSUE #592: Merge `testing` crate into other SGX crate(s)
There isn't much in that crate, so a separate crate for this probably isn't needed. The macro from `testing` is used mainly (but not only) in `sgx-types`, which I believe is also going to get merged into another crate eventually. This will reduce unnecessary directories and also help to avoid confusion since we're creating the new `tests/` crate for general testing (of SGX payload, syscalls, etc.).

kubkon:
This PR consolidates the following crates:
* `sgx-types`
* `sgx-crypto`
* `iocuddle-sgx`
* `attestation-types`

into one crate `sgx`. However, since we don't want to the user to having to use all functionality, several inner modules are feature-gated. And so, `sgx-types` being renamed to `sgx::types` is available in `no_std` context. The rest requires `std`. Furthermore, `sgx-crypto` being renamed to `sgx::crypto` is behind a feature `crypto`. All features *are* enabled by default.

While here, I've gone ahead and migrated `testing::testaso!` macro into the `sgx` crate and put it behind a `test` feature. Also, since `intel-types` crate is going upstream, I've done the same for that crate and completely purged `testing` crate from existence.

EDIT:

Closes #592 
Closes #519


sgx/src/lib.rs
```diff
@@ -10,5 +10,7 @@
 #![deny(missing_docs)]
 
 pub mod types;
-#[cfg(feature = "std")]
+#[cfg(feature = "iocuddle")]
+pub mod iocuddle;
```
npmccallum:
```suggestion
pub mod ioctls;
```

sgx/Cargo.toml
```diff
@@ -18,8 +20,10 @@ enumerate  = { path = "../enumerate" }
 intel-types = { path = "../intel-types" }
 memory = { path = "../memory" }
 openssl = { version = "=0.10.29", features = ["vendored"], optional = true }
+iocuddle_lib = { path = "../iocuddle", optional = true, package = "iocuddle" }
```
npmccallum: I suggested renaming the module instead below, which obviates the need for the awkward renaming here.

sgx/src/lib.rs
```diff
@@ -9,6 +9,72 @@
 #![allow(clippy::identity_op)]
 #![deny(missing_docs)]
 
+#[cfg(test)]
+#[macro_use]
+mod testing {
```
npmccallum, line 14: I don't see any reason to have the macro in its own module at this point.

sgx/Cargo.toml
kubkon: Oh, sorry, I've already got rid of this. If you refresh the changes you'll see the rename is no longer needed.

sgx/src/lib.rs
kubkon: So you're saying we want to put `iocuddle` module behind a feature flag, correct?

kubkon:
> In addition to the changes requested, please open three new follow-up issues that mention this PR.

Will do!

> 1. `sgx::crypto::Signer` should be converted to a conditionally compiled method instead of an extension trait. The use of the trait was only due to needing to extend a type cross-crate. This also means the only thing left in the crypto module is the `Hasher`, so the module should probably be renamed.
> 2. Rethink openssl. The `openssl` crate is actively ignoring our PRs. Our crypto needs are _very_ small in this crate. So we probably don't need the whole of `openssl`. We only need RSA and SHA256. So maybe we should use the `rsa` and `sha2` crates?

Yeah, I remember us having this conversation back in February (!) and here I am and still see the changes didn't make it into `openssl` upstream. If the maintainer is ignoring the changes, I concur with finding an alternative route.

> 3. Expose a high-level interface. We already have an imperfect one in `enarx-keep-sgx::builder`. We could move that from `enarx-keep-sgx` to `sgx`. Then `sgx::ioctls` could be made private.

That sounds like a great idea actually!

sgx/src/lib.rs
npmccallum: No. I was suggesting naming the module `ioctls` to avoid accidental collision with the `iocuddle` crate. IMHO there should be two features: `std` and `crypto`.

sgx/Cargo.toml
npmccallum: One style thing: we do merges of logical commits. We don't do stream-of-conciousness style PRs. So if you make this change here and then undo it in a later patch set, we're going to expect that these get squashed (as appropriate).

sgx/src/attestation_types/mod.rs
```diff
@@ -4,9 +4,7 @@
 //! Section references in further documentation refer to this document.
 //! https://www.intel.com/content/dam/www/public/emea/xe/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3d-part-4-manual.pdf
 
-#![deny(clippy::all)]
-#![allow(clippy::identity_op)]
-#![deny(missing_docs)]
+#![cfg(feature = "std")]
```
lkatalin, line 7: Not sure if this changes in one of the later commits, but since the `attestation-types` module will be used inside the enclave, it should not be `std`.

kubkon: Hmm, that's interesting because currently `attestation-types` requires `std`. Anyhow, I'm fine in adapting it to `core`, but perhaps we should make it in a subsequent PR?

kubkon: Unless, the changes are trivial but I'll need to take a closer look :-)

sgx/src/lib.rs
kubkon: Ah cool. Thanks for the clarification!

sgx/Cargo.toml
kubkon: Ah, ok. I wasn't sure cause every project is governed by its own policy. Would it be OK if I rebased selectively as well then? FWIW I'm personally a big fan of rebases and squashes but sometimes selective/retrospective rebases especially after the PR was submitted may lead to weirdness on Github. Let me know what you think @npmccallum!

sgx/src/attestation_types/mod.rs
lkatalin: @kubkon That *is* interesting! In that case, feel free to leave as-is and we can file a separate issue for it when it comes up.

sgx/Cargo.toml
npmccallum: Yes, we are a fan of rebases and squashes. FYI, GitHub recently fixed the biggest issue I've seen with rebase and squash. https://github.blog/changelog/2020-07-14-pull-request-commits-now-ordered-chronologically/

sgx/src/attestation_types/mod.rs
kubkon: @lkatalin OK, I've had a closer look, and it seems we're using `std::vec::Vec` which means, for the module to be `no_std` compatible we'd either need to replace any structure requiring a global allocator with a statically-sized equivalent such as a slice, or write our own alloc, and then we are OK to use `alloc::vec::Vec`. Note however that in the latter case we're forced to rely on unstable nightly features since not everything has been stabilised in Rust wrt to custom global allocators (in fact, I'm thinking here mainly of [this issue](https://github.com/rust-lang/rust/issues/51540)). FWIW, I believe that we could easily write our own "fixed memory allocator" which would essentially be a wrapper around a fixed chunk of statically-sized memory we could operate on. Either way, I think this warrants a separate issue and a broader discussion, would you agree?

sgx/src/lib.rs
kubkon: Done!

sgx/Cargo.toml
kubkon: Rebased and squashed!

sgx/src/lib.rs
kubkon: Excellent point. Module removed!

kubkon:
@npmccallum @lkatalin comments addressed, have a look and lemme know what you think!

kubkon:
@npmccallum I've now created the three issues you asked for: #752, #753 and #754.

sgx/src/attestation_types/mod.rs
kubkon: FYI @lkatalin, I've moved the discussion to #755.

kubkon:
> Approving. However, this probably went too far in the other direction of squashing. Generally we like small, logically discrete changes that represent the end state. So the problem in the previous patch set was not that there were multiple commits, but that earlier commits did things that were undone by later commits. The reason we're so picky here is because, as a security project, we want to have high standards for auditability. This goes towards building trust. Nevertheless, this PR is still fine, just a little long.

Thanks, and noted! I'll try to break things up in my subsequent work!

kubkon:
Closes #592.

sgx/src/attestation_types/mod.rs
lkatalin: That's great, thanks for filing the issue @kubkon .

PR #751, opened 2020-07-22
sev-shim: use lazy_static for the GDT

haraldh:


enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -18,8 +18,89 @@ use x86_64::{PrivilegeLevel, VirtAddr};
...
+            debug_assert_eq!(STACKS.len(), tss.interrupt_stack_table.len());
+        }
+
+        tss
+    };
+}
+
+/// The Selectors used in the GDT setup
+pub struct Selectors {
+    /// shim code selectore
```
connorkuehl, line 55: ~~selectore~~ selector

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -18,8 +18,89 @@ use x86_64::{PrivilegeLevel, VirtAddr};
...
+    pub code_sel: SegmentSelector,
+    /// shim data selectore
```
connorkuehl, line 57: ~~selectore~~ selector

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -18,8 +18,89 @@ use x86_64::{PrivilegeLevel, VirtAddr};
...
+
+            debug_assert_eq!(STACKS.len(), tss.interrupt_stack_table.len());
+        }
+
+        tss
+    };
+}
+
+/// The Selectors used in the GDT setup
+pub struct Selectors {
```
connorkuehl, line 54: I think the `_sel` suffix on these fields should be removed, since I think their intent is already well-described by accessing a field of a struct called `Selectors`.

enarx-keep-sev-shim/src/gdt.rs
```diff
@@ -18,8 +18,89 @@ use x86_64::{PrivilegeLevel, VirtAddr};
...
+        (
+            gdt,
+            Selectors {
+                code_sel,
+                data_sel,
+                user_data_sel,
+                user_code_sel,
+                tss_sel,
+            },
+        )
```
connorkuehl, line 101: Style nitpick: constructing the `Selectors` struct before the expression would let us avoid the `(` and `)` from being given their own lines. I think it's easier to tell what's going on that way.

```rust
let selectors = Selectors {
    code_sel,
    data_sel,
    user_data_sel,
    user_code_sel,
    tss_sel,
};

(gdt, selectors)

File: enarx-keep-sev-shim/src/gdt.rs [hunk 5]
```diff
@@ -18,8 +18,89 @@ use x86_64::{PrivilegeLevel, VirtAddr};
 const STACK_NUM_PAGES: usize = 5;
 static mut STACKS: [[Page; STACK_NUM_PAGES]; 7] = [[Page::zeroed(); STACK_NUM_PAGES]; 7];
 
-pub(crate) static mut TSS: TaskStateSegment = TaskStateSegment::new();
-pub(crate) static mut GDT: GlobalDescriptorTable = GlobalDescriptorTable::new();
+/// FIXME: will be replaced by a dynamically allocated stack with safe guards
+pub static mut LEVEL_0_STACK: [Page; 5] = [Page::zeroed(); 5];
```
connorkuehl, line 22: Do we need a lock here if we have multiple CPUs?

enarx-keep-sev-shim/src/gdt.rs
haraldh: done

enarx-keep-sev-shim/src/gdt.rs
haraldh: done

enarx-keep-sev-shim/src/gdt.rs
haraldh: done

enarx-keep-sev-shim/src/gdt.rs
haraldh: hah, cause, it is in a macro, rustfmt doesn't do a format anyway.

enarx-keep-sev-shim/src/gdt.rs
haraldh: Not for this (will be removed anyway in the next patch).

Each CPU will get it's own GDT and TSS. So, yes, it will have a lock, then.

PR #756, opened 2020-07-22
make sallyport::Cursor safe again

haraldh:


enarx-keep-sev-shim/src/hostcall.rs
```diff
@@ -85,7 +85,7 @@ impl<'a> HostCall<'a> {
     /// The parameters returned can't be trusted.
     pub unsafe fn write(&mut self, fd: usize, bytes: &[u8]) -> Result<usize, libc::c_int> {
         let cursor = self.0.cursor();
-        let buf = cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE))?;
+        let (_c, buf) = cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE))?;
```
npmccallum, line 88:
```suggestion
        let (_, buf) = cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE))?;
```

Do that everywhere please.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -181,7 +181,8 @@ impl<'a> Handler<'a> {
 
         let cursor = self.block.cursor();
         let trusted: &mut [u8] = unsafe { self.aex.gpr.rsi.into_slice_mut(self.aex.gpr.rdx) };
-        let untrusted: &mut [u8] = unsafe { cursor.alloc(trusted.len()).or(Err(libc::EMSGSIZE))? };
+        let (_cursor, untrusted) =
+            unsafe { cursor.alloc::<u8>(trusted.len()).or(Err(libc::EMSGSIZE))? };
```
npmccallum, line 185:
```suggestion
        let (_, untrusted) = unsafe { cursor.alloc::<u8>(trusted.len()).or(Err(libc::EMSGSIZE))? };
```

Plain underscore should resolve this unnecessary wrap.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -263,10 +267,16 @@ impl<'a> Handler<'a> {
         }
 
         let cursor = self.block.cursor();
-        let _: &mut [libc::iovec] = unsafe { cursor.alloc(trusted.len()).or(Err(libc::EMSGSIZE))? };
+        let (mut cursor, _) = unsafe {
```
npmccallum, line 270:
```suggestion
        let (mut c, _) = unsafe {
```
Let's just name cursor `c` in this function. This will resolve multiple instances of jankyness.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -282,11 +292,14 @@ impl<'a> Handler<'a> {
         let mut size = 0usize;
         let cursor = self.block.cursor();
         let trusted = unsafe { self.aex.gpr.rsi.into_slice_mut(self.aex.gpr.rdx) };
-        let untrusted: &mut [libc::iovec] = cursor.copy_slice(trusted).or(Err(libc::EMSGSIZE))?;
+        let (mut cursor, untrusted) = cursor
```
npmccallum, line 295:
```suggestion
        let (mut c, untrusted) = cursor
```

Same here. Rename `cursor` to `c` everywhere.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -457,7 +470,8 @@ impl<'a> Handler<'a> {
         let trusted = unsafe { self.aex.gpr.rsi.into_slice_mut(1usize) };
 
         let cursor = self.block.cursor();
-        let untrusted = unsafe { cursor.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))? };
+        let (_cursor, untrusted) =
```
npmccallum, line 473:
```suggestion
        let (_, untrusted) =
```

Discard with naked underscore.

enarx-keep-sgx-shim/src/handler.rs
```diff
@@ -466,7 +480,8 @@ impl<'a> Handler<'a> {
         }
 
         let cursor = self.block.cursor();
-        let untrusted = unsafe { cursor.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))? };
+        let (_cursor, untrusted) =
```
npmccallum, line 483:
```suggestion
        let (_, untrusted) =
```

Discard with naked underscore.

sallyport/src/lib.rs
```diff
@@ -179,24 +179,26 @@ impl Block {
...
+        Cursor(&mut self.buf)
     }
 }
 
 /// Helper for allocation of untrusted memory in a Block.
-pub struct Cursor<'a>(UnsafeCell<&'a mut [u8]>);
+pub struct Cursor<'a>(&'a mut [u8]);
 
-impl Cursor<'_> {
+impl<'a> Cursor<'a> {
```
npmccallum, line 190:
```suggestion
impl Cursor<'_> {
```

sallyport/src/lib.rs
```diff
@@ -179,24 +179,26 @@ impl Block {
...
     ///
     /// # Safety
     ///
     /// This function is unsafe because it returns an uninitialized array.
-    pub unsafe fn alloc<T>(&self, count: usize) -> core::result::Result<&mut [T], ()> {
-        let inner_ptr = self.0.get();
+    pub unsafe fn alloc<T>(
+        self,
+        count: usize,
+    ) -> core::result::Result<(Cursor<'a>, &'a mut [T]), ()> {
```
npmccallum, line 199:
```suggestion
    ) -> core::result::Result<(Cursor<'_>, &mut [T]), ()> {
```

I think explicit elision should work here. If not, ignore this comment and the previous one.

sallyport/src/lib.rs
```diff
@@ -205,18 +207,19 @@ impl Cursor<'_> {
...
+        Ok((Cursor(next), data.align_to_mut::<T>().1))
     }
 
     /// Copies data from a value into a slice using self.alloc().
-    pub fn copy_slice<T: Copy>(&self, value: &[T]) -> core::result::Result<&mut [T], ()> {
-        let slice: &mut [T] = unsafe { self.alloc(value.len())? };
+    pub fn copy_slice<T: Copy>(
+        self,
+        value: &[T],
+    ) -> core::result::Result<(Cursor<'a>, &'a mut [T]), ()> {
```
npmccallum, line 219: Explicit elision?

kubkon: Hmm, I'm not sure you can apply explicit elision here since we do have two lifetimes going on here: 1) is the lifetime bound to cursor *and* returned slice, and 2) is the lifetime bound to the input slice. I think we are fine here saying the two do not necessarily have to be equal, but rather, one can be a subset of the other, right? In this case, explicit elision will collapse both lifetimes into one and the same.

npmccallum: Which goes to my overall concern posted at the top of the review. I'm concerned this patch might allow the allocated slice to live longer than the cursor.

sallyport/src/lib.rs
kubkon: It is legal in Rust to return a lifetime-annotated struct with an anonymous lifetime: [play.rust-lang.org](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ae185c6940a47a2bc6681a9fc7863de3).

sallyport/src/lib.rs
kubkon: Because of the comment below, an explicit lifetime is required here.

sallyport/src/lib.rs
kubkon: Oh, the output slice is still bound by the same lifetime as the original block stored in the first `Cursor`. Effectively what @haraldh has done here was adapting the [`std::slice::split_at_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_at_mut) pattern in `Cursor`. What this implies is that we cannot hold multiple mutable borrows to overlapping slices (i.e., alias the memory), but we can to non-overlapping chunks.

haraldh:
> One really important question I have about this change is whether the returned slice is still bound to the lifetime of the cursor. That is, what happens if you create a second cursor? If it allows simultaneous access to slices from the old and new cursors, we have a definite problem.

```console
error[E0499]: cannot borrow `block` as mutable more than once at a time
   --> src/lib.rs:331:18
    |
329 |         let c = block.cursor();
    |                 ----- first mutable borrow occurs here
330 |         let (c, slice) = c.copy_slice(&[87, 2, 3]).unwrap();
331 |         let c2 = block.cursor();
    |                  ^^^^^ second mutable borrow occurs here
332 |         assert_eq!(unsafe { c.alloc::<usize>(42usize) }.unwrap().1.len(), 42);
    |                             - first borrow later used here
```

enarx-keep-sev-shim/src/hostcall.rs
haraldh: I hate doing this just to satisfy rustfmt... ah well..

haraldh:
renaming `cursor` to `c` does not help in all cases.

Funny sidenote, adding `{}`:
```diff
-        let (c, untrusted) = c
-            .copy_slice::<libc::iovec>(trusted)
-            .or(Err(libc::EMSGSIZE))?;
+        let (c, untrusted) = { c.copy_slice::<libc::iovec>(trusted) }.or(Err(libc::EMSGSIZE))?;
 ```

**Comment by kubkon** [comment]
> renaming `cursor` to `c` does not help in all cases.
> 
> Funny sidenote, adding `{}`:
> 
> ```diff
> -        let (c, untrusted) = c
> -            .copy_slice::<libc::iovec>(trusted)
> -            .or(Err(libc::EMSGSIZE))?;
> +        let (c, untrusted) = { c.copy_slice::<libc::iovec>(trusted) }.or(Err(libc::EMSGSIZE))?;
> ```

Hmm, I'm actually partial to multi-line wrapping in this case, but that's apples and oranges ;-)

**Comment by haraldh** [comment]
@npmccallum you might want/like

```console
$ cargo fmt -- --config 'use_small_heuristics=Max'
```

haraldh:
meh... sorry, @kubkon, you have to approve again. @mbestavros removed your review approval

Issue #747, opened 2020-07-21
Dynamic output from integration tests


npmccallum:
Currently, our [integration tests allow us to specify](https://github.com/enarx/enarx/blob/master/integration-tests/tests/common/mod.rs#L71-L78) the input to `stdin` and the expected output on `stdout` and `stderr`. However, this presumes that the outputs are deterministic. https://github.com/enarx/enarx/pull/681 breaks that assumption.

When testing `clock_gettime` we should have the payload print the current time to `stdout` and then have the test harness validate that the output is within an acceptable delta from its own version of `clock_gettime`. Therefore, `run()` function for the tests should be made to return the outputs as `Vec<u8>` and the `clock_gettime` test should be rewritten to handle this.

- [x] Make `run()` return the outputs.
- [x] Fix up existing tests.
- [x] Rewrite `clock_gettime` to output the time, evaluated by the harness.

npmccallum:
FYI @jyotsna-penumaka

jyotsna-penumaka:
@npmccallum should I take it up?

PR #757, opened 2020-07-22
Add better failure reporting in integration tests

kubkon:
This commit tweaks the standard `assert_eq`s in `integration-tests`
so that instead of only dumping the raw bytes when asserting equality
between expected and given stdio outputs, it now also prints the
mismatch converted to a UTF8 string for easier debugging. The conversion
is lossy so that invalid UTF8 don't trigger false positives in our
testing harness.

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -106,8 +106,20 @@ impl IntegrationTest {
             .wait_with_output()
             .expect("Failed to read stdout/stderr");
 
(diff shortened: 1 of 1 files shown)


kubkon:
> Approved. However, we still have the problem outlined in #747 .

Thanks for the link @npmccallum! Unless @jyotsna-penumaka wants to take/already took it, I can look into this issue as well :+1:

jyotsna-penumaka:
@kubkon you can take this issue :)

PR #761, opened 2020-07-23
Use sallyport for SGX syscall proxying

lkatalin:
This change removes the `handle()` function passed to the VDSO, instead
handling the process flow in the Keep's `main()`. It also transitions to
using the `sallyport::Block` to proxy system calls.

Related to #614

### DIFF
--- a/enarx-keep-sgx/src/main.rs
+++ b/enarx-keep-sgx/src/main.rs
@@ -131,22 +131,38 @@ fn main() {
     let enclave = load();
 
     let mut block = Block::default();
+    let mut leaf = Leaf::Enter;
+    const SYS_ERESUME: usize = !0;
 
-    // The main loop event handing is divided into two halves.
+    // The main loop event handles different types of enclave exits and
+    // re-enters the enclave with specific parameters.
     //
-    //   1. EEXIT events (including syscall proxying and ERESUMEs [CSSA--])
-    //      are handled by the handler callback to the vDSO function. See
-    //      enclave.rs and enclave.S. This allows us to pass registers
-    //      directly to the syscall instruction.
+    //   1. Asynchronous exits (AEX) with an invalid opcode indicate
+    //      that a syscall should be performed. Execution continues in
+    //      the enclave with EENTER[CSSSA = 1]. The syscall
+    //      is proxied and potentially passed back out to the host.
     //
-    //   2. Asynchronous exits (AEX) are handled here to minimize the amount
-    //      of assembly code used.
+    //   2. OK with a syscall number other than SYS_ERESUME indicates the syscall
+    //      to be performed. The syscall is performed here and enclave
+    //      execution resumes with EENTER[CSSA = 1].
+    //
+    //   3. OK with a syscall number of SYS_ERESUME indicates that a syscall has
+    //      been performed as well as handled internally in the enclave
+    //      and normal enclave execution should resume
+    //      with ERESUME[CSSA = 0].
+    //
+    //   4. Asynchronous exits other than invalid opcode will panic.
     loop {
-        match enclave.enter(&mut block as *const _ as _, 0, 0, Leaf::Enter, 0, 0) {
-            // On InvalidOpcode: re-enter the enclave with EENTER (CSSA++).
-            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => (),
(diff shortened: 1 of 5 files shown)


enarx-keep-sgx/src/main.rs
```diff
@@ -131,6 +131,9 @@ fn main() {
     let enclave = load();
 
     let mut block = Block::default();
+    let mut leaf = Leaf::Enter;
+    let mut cssa = 0;
+    const ERESUME: usize = !0;
```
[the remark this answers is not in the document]
npmccallum: Yes, eventually. We need a libshim. But I *think* that needs to wait until we make the shim static-pie.

enarx-keep-sgx/src/main.rs
```diff
@@ -142,9 +145,27 @@ fn main() {
...
     //      of assembly code used.
     loop {
-        match enclave.enter(&mut block as *const _ as _, 0, 0, Leaf::Enter, 0, 0) {
+        match enclave.enter(&mut block as *const _ as _, 0, 0, leaf, 0, 0) {
             // On InvalidOpcode: re-enter the enclave with EENTER (CSSA++).
-            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => (),
+            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => {
+                leaf = Leaf::Enter;
+                cssa += 1;
+            }
```
npmccallum, line 153:
```suggestion
            Ok(_) if unsafe { block.msg.req.num } == ERESUME.into() => Leaf::Resume,
            Ok(_) => { block.msg.rep = unsafe { block.msg.req.syscall() }; Leaf::Enter },
            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => Leaf::Enter,
```

Something like that to show the clear options. I don't care if you reintroduce `cssa` if it is used in debugging output. But it is currently unused.

enarx-keep-sgx/src/main.rs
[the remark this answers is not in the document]
lkatalin: I made a debt issue: https://github.com/enarx/enarx/issues/764.

connorkuehl:
@lkatalin Could you rebase now that #765 is merged?

enarx-keep-sgx/src/main.rs
```diff
@@ -131,22 +131,38 @@ fn main() {
...
+    //   2. OK with a syscall number other than ~0 indicates the syscall
+    //      to be performed. The syscall is performed here and enclave
+    //      execution resumes with EENTER[CSSA = 1].
+    //
+    //   3. OK with a syscall number of ~0 indicates that a syscall has
+    //      been performed as well as handled internally in the enclave
+    //      and normal enclave execution should resume
+    //      with ERESUME[CSSA = 0].
+    //
+    //   4. Asynchronous exits other than invalid opcode will panic.
```
npmccallum, line 154: Update the documentation to not reference `~0` now that we have a constant for it.

enarx-keep-sgx/src/main.rs
```diff
@@ -131,22 +131,38 @@ fn main() {
...
-            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => (),
+        leaf = match enclave.enter(&mut block as *const _ as _, 0, 0, leaf, 0, 0) {
+            Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => Leaf::Enter,
+
+            Ok(_) if unsafe { block.msg.req.num } != ERESUME.into() => {
+                block.msg.rep = unsafe { block.msg.req.syscall() };
+                Leaf::Enter
+            }
+
+            Ok(_) => Leaf::Resume,
```
npmccallum, line 164: This check should be inverted. We're basically creating a syscall filter here. So "no match" should be "pass the syscall to the kernel."

npmccallum: One other change: let's rename `ERESUME` to `SYS_ERESUME` since it is a virtual syscall.

lkatalin: I flipped the match arms so that it would match the documentation and be easier to follow. But it is probably fine either way. I can change it back.

lkatalin:
@connorkuehl @npmccallum I didn't change anything since your last approval; we're just trying to figure out why the bot didn't merge this, so I did a rebase.

PR #767, opened 2020-07-26
keep-runtime: Implement resource bundling in Wasm file

ueno:
This implements the [proposal](https://github.com/enarx/rfcs/pull/29) to embed application resources in the Wasm file through a custom section.


ueno:
To actually test bundling, use [this tool](https://github.com/ueno/wasm-bundle).

keep-runtime/src/virtfs.rs
```diff
@@ -0,0 +1,82 @@
...
+        let read_count = std::cmp::min(buf.len(), data_remaining);
+
+        (&mut buf[..read_count]).copy_from_slice(&self.content[offset..][..read_count]);
+        Ok(read_count)
+    }
+
+    fn pwrite(&mut self, _buf: &[u8], _offset: types::Filesize) -> Result<usize> {
+        Err(types::Errno::Inval)
+    }
+}
```
kubkon, line 60: I assume all of this will go away once #2070 lands?

ueno: Unfortunately this needs to be kept, because `VecFileContents` is not exported from wasi-common and we need to implement `FileContents` backed by the bytes extracted from the tar archive.

npmccallum:
This PR is outdated.

PR #777, opened 2020-07-29
Let the vDSO preserve %rbx

### LINKED ISSUE #614: Use sallyport with SGX to communicate with host

lkatalin:
Closes #614

### DIFF
--- a/enarx-keep-sgx/src/enclave.S
+++ b/enarx-keep-sgx/src/enclave.S
@@ -19,24 +19,23 @@
     .global eenter
     .type eenter, @function
 eenter:
+    # Preserve registers not preserved by vDSO
     .cfi_startproc
     rpush   %r15
     rpush   %r14
     rpush   %r13
     rpush   %r12
-    rpush   %rbx
 
-    dpush   $0
-    dpush   0x48(%rsp)
-    dpush   0x48(%rsp)
-    dpush   0x48(%rsp)
+    # Prepare stack for vDSO
+    dpush   0x38(%rsp)          # exit handler
+    dpush   0x38(%rsp)          # exception info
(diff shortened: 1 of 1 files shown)


lkatalin:
> Honestly I don't know what's going on here.. there is not even a comment what is supposed to be in `0x48(%rsp)` or `0x38(%rsp)` in the patched version.
> 
> Leaving out 2 pushes, needs correction by `0x10` .. I get that.
> 
> Leaving out the `dpush $0` (maybe it was there for alignment) , which was skipped with the `add` leads to a `0x8` offset correction.
> 
> LGTM, I just don't know what the parameters are.

@haraldh You're right, I think this should have more descriptive comments. I will do a new version of this PR or maybe a separate PR. `dpush $0` was in fact there for alignment.

lkatalin:
@npmccallum @haraldh Since the bot requested re-approval anyway, I pushed a new version with some comments.

PR #787, opened 2020-08-05
shim sev/sgx: remove the indirection of the inital function rust call

haraldh:
This was introduced with the static-pie rcrt1 dynamic symbol correction,
copying the behaviour of the musl lib rcrt1.o. Apparently it is not
necessary to use the indirection, as the code base is not big enough
yet, that it exceeds the maximum relative jmp/call distance.

If that is ever the case, we should revert this patch.


lkatalin:
@haraldh What do we gain by removing this? Easier to read code?

haraldh:
yes

PR #790, opened 2020-08-05
Add dist-clean script to clean Cargo.lock files

ambaxter:
Signed-off-by: Adam Baxter <ambaxter@users.noreply.github.com>

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -169,3 +169,10 @@ dependencies = [
 	"misc-lints-clippy-all",
 	"misc-lints-missing-docs",
 ]
+
+# Remove Cargo.lock files in cases where compilation fails
+# due to incorrect library update propogation
+[tasks.dist-clean]
+dependencies = ["clean"]
+command = "rm"
+args = ["-f", "Cargo.lock"]


Makefile.toml
```diff
@@ -169,3 +169,17 @@ dependencies = [
...
+workspace = false
+script_runner = "@duckscript"
+script = [
+'''
+workspace_members = split ${CARGO_MAKE_CRATE_WORKSPACE_MEMBERS} ";"
+for workspace in ${workspace_members}
+    rm "${workspace}/Cargo.lock"
+end
+'''
+]
```
mbestavros, line 185: Needs a newline at the end.

ambaxter: Done

npmccallum, line 185: AFAICS, running `cargo make dist-clean` won't actually (implicitly) run `cargo make clean`, which is probably the expected behavior.

ambaxter: Done

ambaxter:
> What do you think about something like this instead:
> 
> ```toml
> [tasks.dist-clean]
> dependencies = ["clean"]
> command = "rm"
> args = ["-f", "Cargo.lock"]
> ```

Even better. I'll update the makefile. Would there ever be any situation where 'rm' doesn't exist?

connorkuehl:
> Even better. I'll update the makefile. Would there ever be any situation where 'rm' doesn't exist?

Right now our only real "supported" development environment is Fedora, so I think the risk of `rm` being missing is extremely, extremely unlikely. Off the top of my head, I can't think of a Linux distro that doesn't have `rm`/coreutils.

My vote is that if a portability issue with `rm` comes up, we'll cross that bridge when the ticket is filed. :-)

PR #796, opened 2020-08-07
Make rustfmt and clippy non-nightly

haraldh:
Using nightly rustfmt will cause CI errors all the time, when rustfmt
changes.

Also all components should build with stable by default and only
selected components should switch to nightly.

Additionally the `--channel=nightly` argument for `rustup.sh` did not
work for me.

```console
error: error: Found argument '--channel' which wasn't expected, or isn't valid in this context
```


haraldh:
A nightly crate would have an additional `Makefile.toml` with

```toml
[tasks.build]
toolchain = "nightly"

[tasks.build-release]
toolchain = "nightly"

[tasks.test]
toolchain = "nightly"
```

PR #798, opened 2020-08-10
enarx-keep: add `--shim-log-level <u8>` argument

haraldh:
This patch adds `--log-level <u8>` to enarx-keep.

It also adds a log_level element to the SEV `BootInfo` struct, where the value will get measured.

### DIFF
--- a/enarx-keep-sev-shim/src/lib.rs
+++ b/enarx-keep-sev-shim/src/lib.rs
@@ -87,8 +87,11 @@ pub struct BootInfo {
     pub shim: Line<usize>,
     /// Memory size
     pub mem_size: usize,
+    /// FIXME: remove volatile `virt_offset`
     /// Loader virtual memory offset to shim physical memory
     pub virt_offset: usize,
+    /// Log level
+    pub shim_log_level: u8,
 }
 
 /// Error returned, if the virtual machine memory is to small for the shim to operate.
@@ -129,6 +132,7 @@ impl BootInfo {
             shim,
             mem_size,
             virt_offset,
+            shim_log_level: 0,
         })
     }
 }
--- a/enarx-keep-sev-shim/src/print.rs
+++ b/enarx-keep-sev-shim/src/print.rs
@@ -6,6 +6,7 @@ use crate::hostcall::{self, HostFd};
 
 struct HostWrite(HostFd);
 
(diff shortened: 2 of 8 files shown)


enarx-keep-sev-shim/Cargo.toml
```diff
@@ -33,3 +33,7 @@ panic = "abort"
 codegen-units = 1
 incremental = false
 lto = true
+
+[features]
+default = []
+debug = []
```
connorkuehl, line 39: Why a feature? Why not use `#[cfg(debug_assertions)]` or `#[cfg(not(debug_assertions))]`?

haraldh: Because for the integration tests the actual output is tested, so every additional output will fail the integration tests.

connorkuehl:
This definitely works to avoid polluting stdout/stderr, but I don't think it fully solves the problem that we have two separate entities (loader/VMM + shim) sharing those host-side file descriptors. It would also require a new build to turn these back on if we ever need to access this information for testing purposes.

I think a better short term solution would be to remove the shim print statements entirely; at least while we work toward having the shim store its log messages in something like a ring buffer. Then we wouldn't have two entities sharing these host-side file descriptors and the payload can use them transparently without any fear of competing with the shim. What do you think?

enarx-keep-sev-shim/Cargo.toml
npmccallum: @haraldh I think that just raises the question: should our test suite run in release mode? I rather like the idea that our CI tests the bits we would ship.

haraldh: @npmccallum it should run in `--debug` and `--release` to catch all the `debug_assert!()`

haraldh:
Ok, I changed it to a `--log-level <u8>` argument for `enarx-keep`.

It also adds a log_level element to the SEV `BootInfo` struct, where the value will get measured.

haraldh:
renamed `log_level` to `shim_log_level` and `--log-level <u8>` to `--shim-log-level <u8>`

haraldh:
opened https://github.com/enarx/enarx/issues/831

haraldh:
moving it back to a draft state until #831 is fleshed out

PR #799, opened 2020-08-10
sev-shim: implement read syscall

haraldh:


enarx-keep-sev-shim/src/syscall.rs
```diff
@@ -85,6 +90,35 @@ impl Handler {
...
+        let (_, buf) = unsafe { c.alloc::<u8>(trusted.len()).or(Err(libc::EMSGSIZE))? };
+
+        let buf_address = Address::from(buf.as_ptr());
+        let shim_virt_address = ShimVirtAddr::try_from(buf_address).map_err(|_| libc::EFAULT)?;
+        let host_virt: HostVirtAddr<_> = ShimPhysAddr::from(shim_virt_address).into();
+
+        block.msg.req = request!(libc::SYS_read => fd, host_virt, len);
+        let result = unsafe { host_call.hostcall() };
+        let result = result.map(|r| r[0].into())?;
+
```
lkatalin, line 113: Before copying from the untrusted sallyport buffer to trusted, should it check that the number of bytes read is the expected number and does not exceed the trusted buffer size? In SGX we have something like
```
if trusted.len() < ret[0].into() {
            self.attacked();
}
```

haraldh: right... fixed :+1:

PR #803, opened 2020-08-10
sev: split enarx-keep-sev-shim into binary and lib crate

haraldh:
Because Cargo.toml does not let you specify different dependencies for
lib and bin, a move to nightly for the bin, will make everything needing
lib also nightly.


npmccallum:
@haraldh Only `enarx-keep` and `enarx-keep-sev-shim` will depend on `enarx-keep-sev-lib`, right? Both of those will require nightly anyway (`enarx-keep` will definitely require nightly for `sgx` support). So I don't see a lot of reason to divide this.

haraldh:
ok, cool

PR #804, opened 2020-08-10
containers/fedora: install nightly clippy

haraldh:
`cargo ci-flow` runs clippy and clippy errors out on `#[feature()]`

```
error[E0554]: `#![feature]` may not be used on the stable release channel
```


haraldh:
although `cargo make` installs nightly clippy, it can happen, that it is not available for this nightly version... therefore the PR

haraldh:
> What about scenarios where clippy isn't available? Sometimes updates will fail if it's not available.

that's why allow downgrade is specified. rustup will search for the last version all components are available

PR #806, opened 2020-08-12
keep-runtime: Port to using wasmtime's Linker interface

ueno:
The [wasmtime::Linker](https://docs.rs/wasmtime/0.19.0/wasmtime/struct.Linker.html) interface transparently performs name resolution which we currently do by our own.


MikeCamel:
You beat me to it!  I was going to work on that this morning.

PR #830, opened 2020-08-13
cargo clippy fixes

haraldh:
```
$ cargo +nightly make clippy
```


mmap/src/lib.rs
```diff
@@ -28,6 +34,11 @@ impl Drop for Unmap {
     }
 }
 
+/// mmap a memory region
+///
+/// # Safety
+///
+/// FIXME: yeah, why is this unsafe? The usage is unsafe, but IMHO not this operation
```
haraldh, line 41: @connorkuehl ?

haraldh: fixed

sgx/src/lib.rs
```diff
@@ -78,7 +78,10 @@ pub mod crypto;
 
 #[cfg(feature = "std")]
 pub mod attestation_types;
+
 #[cfg(feature = "std")]
+// FIXME:
+#[allow(clippy::module_inception)]
```
haraldh, line 84: @npmccallum clippy does not like it:
https://rust-lang.github.io/rust-clippy/master/index.html#module_inception

sgx/src/lib.rs
```diff
@@ -78,7 +78,10 @@ pub mod crypto;
...
 pub mod enclave;
```
haraldh, line 85: @npmccallum clippy does not like it:
https://rust-lang.github.io/rust-clippy/master/index.html#module_inception

haraldh: There is `src/enclave/enclave.rs`.

sevctl/src/main.rs
```diff
@@ -22,18 +22,18 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
...
     }
 
     let mut buf = Vec::new();
     rsp.copy_to(&mut buf)
-        .expect(&format!("unable to complete {} download", usage));
+        .unwrap_or_else(|_| panic!("unable to complete {} download", usage));
 
     sev::Certificate::decode(&mut &buf[..], ())
-        .expect(&format!("unable to parse downloaded {}", usage))
+        .unwrap_or_else(|_| panic!("unable to parse downloaded {}", usage))
```
connorkuehl, line 36: Does clippy still complain if we don't borrow the original `format!()` statement? I feel like this is precisely the use-case for `expect()`

enarx-keep-sev-shim/src/payload.rs
```diff
@@ -54,7 +55,7 @@ lazy_static! {
     pub static ref NEXT_MMAP_RWLOCK: RwLock<VirtAddr> = {
         let boot_info = BOOT_INFO.read().unwrap();
 
-        let mmap_start = (*PAYLOAD_VIRT_ADDR.read().deref() + (boot_info.code.end - boot_info.code.start)).align_up(4096u64);
+        let mmap_start = (*PAYLOAD_VIRT_ADDR.read().deref() + (boot_info.code.end.checked_sub(boot_info.code.start).unwrap())).align_up(4096u64);
```
connorkuehl, line 58: Nit: replace `4096` with `memory::Page::size()`.

sevctl/src/main.rs
haraldh: > Does clippy still complain if we don't borrow the original `format!()` statement? I feel like this is precisely the use-case for `expect()`

```
error[E0308]: mismatched types
  --> src/main.rs:25:30
   |
25 |     let mut rsp = rsp.expect(format!("unable to parse downloaded {}", usage));
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found struct `std::string::String`
   |
```

haraldh: but thinking more about it, with `unwrap_or_else` we should also output the error. Will fix.

enarx-keep-sev-shim/src/payload.rs
haraldh: right

haraldh:
@connorkuehl fixed

haraldh:
> Will you cherry pick this or just copy-paste the contents so that our CI lints everything automatically:
> 
> [connorkuehl@a9c73d7](https://github.com/connorkuehl/enarx/commit/a9c73d7be8777420877a6fcede13cc7021733d60)

oh, right forgot this... added

haraldh:
Filed a cargo-make issue: https://github.com/sagiegurari/cargo-make/issues/447

PR #839, opened 2020-08-16
Merge shim crates into enarx-keep

npmccallum:
This experiment merges `enarx-keep-sgx-shim` and `enarx-keep-sev-shim` into
`enarx-keep`. Then `enarx-keep` grows a build script which builds and installs
the shim binaries into `OUT_DIR`.  Finally, the shim binaries are embedded
into the `enarx-keep` binary via `include_bytes!()`.

The major advantage of this approach is that we can `cargo install enarx-keep`
and the resulting binary is usable. There is no need to fiddle with paths or
other complications. This is a great user experience. It further means we
could merge in `integration-tests` into `enarx-keep` (as a further pull
request) and `enarx-keep` becomes entirely self-contained. It could
potentially be forked off into an independenty repository. Likewise,
`cargo run` can be made to work in a straightforward way.

There are some downsides. First, we don't get any build output from the build
script unless builds fail. Because the shims have dependencies, this could
mean we're waiting on silence for a while. Second, utilities like `cargo fmt`
or `cargo check` wouldn't directly propagate to the shims. So we'd have to
work around this a bit in the CI workflows. The biggest example of this is
`cargo test`. Although integration tests would run, unit tests within the
shims themselves would not run unless you did a manual `cargo test` in the
right directories. Third, this kind of setup might confused some IDEs.
However, I suspect it is no more confusing than our existing `cargo make`
setup; in fact, it might be potentially less confusing.


npmccallum:
Unresolved:

1. We should probably pass more information through the build script. I'm thinking of things like using cargo from `CARGO`.

2. The build script appears to run during `cargo test`. So it might be possible to propagate and run unit tests.

3. I'm sure I've missed other stuff...

npmccallum:
> 2. The build script appears to run during `cargo test`. So it might be possible to propagate and run unit tests.

It is probably just easier to write a test that propagates for each shim.

connorkuehl:
> > 1. The build script appears to run during `cargo test`. So it might be possible to propagate and run unit tests.
> 
> It is probably just easier to write a test that propagates for each shim.

@npmccallum I think something like this can maintain cargo-make coverage with the interior crates (have not tested):

```diff
diff --git a/Makefile.toml b/Makefile.toml
index 53a58c7..b058afa 100644
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -6,8 +6,8 @@ CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = [
        "bounds",
        "crt0stack",
        "enarx-keep",
-       "enarx-keep-sev-shim",
-       "enarx-keep-sgx-shim",
+       "enarx-keep/shims/sev",
+       "enarx-keep/shims/sgx",
        "enumerate",
        "integration-tests",
        "intel-types",
```

npmccallum:
@connorkuehl Yes, but part of the goal is to isolate the inside of `enarx-keep` from `cargo make`.

connorkuehl:
> @connorkuehl Yes, but part of the goal is to isolate the inside of `enarx-keep` from `cargo make`.

@npmccallum Understood. I thought it'd be helpful for resolving some of the downsides described in your opening comment regarding applying CI tasks like `fmt` or our other lints.

mbestavros:
Superseded by https://github.com/enarx/enarx/pull/847

PR #846, opened 2020-08-18
Clean up crt0stack in preparation for a split-off

npmccallum:


crt0stack/src/lib.rs
```diff
@@ -37,11 +37,35 @@
...
+const AT_PLATFORM: usize = 15;
+const AT_HWCAP: usize = 16;
+const AT_SECURE: usize = 23;
+const AT_BASE_PLATFORM: usize = 24;
+const AT_RANDOM: usize = 25;
+const AT_HWCAP2: usize = 26;
+const AT_EXECFN: usize = 31;
+const AT_SYSINFO: usize = 32;
+const AT_SYSINFO_EHDR: usize = 33;
+
```
haraldh: Hmmm... I think I had it that way in my first versions :stuck_out_tongue:

crt0stack/src/builder.rs
```diff
@@ -51,8 +51,14 @@ impl<'a> Handle<'a> {
...
-    pub fn start_ptr(&self) -> &'a () {
-        unsafe { &*(&self.0[self.1] as *const u8 as *const ()) }
+    pub fn start_ptr(&self) -> &'a Stack {
+        #[repr(C, align(16))]
+        struct Aligned(u128);
+
+        let (pre, body, _) = unsafe { self.0[self.1..].align_to::<Aligned>() };
+        assert!(pre.is_empty());
+
+        unsafe { &*(body.as_ptr() as *const _ as *const _) }
```
haraldh, line 62: `as *const _ as *const _` ?

crt0stack/src/builder.rs
```diff
@@ -51,8 +51,14 @@ impl<'a> Handle<'a> {
     /// This reference can be used as the initial stack pointer
     /// to execute a Linux ELF binary.
     #[inline]
-    pub fn start_ptr(&self) -> &'a () {
-        unsafe { &*(&self.0[self.1] as *const u8 as *const ()) }
+    pub fn start_ptr(&self) -> &'a Stack {
+        #[repr(C, align(16))]
+        struct Aligned(u128);
```
haraldh, line 57: Why not take the already defined `Stack` which is already defined? Is it because it is unsized?

crt0stack/src/builder.rs
```diff
@@ -477,7 +478,7 @@ mod tests {
             builder.push(&Entry::Random([0u8; 16])).unwrap();
 
             let handle = builder.done().unwrap();
-            let alignment = (handle.start_ptr() as *const _ as usize) % align_of::<Stack>();
+            let alignment = (&*handle as *const _ as usize) % align_of::<Stack>();
```
haraldh, line 481: I like to call `.deref()` directly to show, that it has a special meaning.

crt0stack/src/builder.rs
haraldh: This has passed `cargo clippy` ? wow...

crt0stack/src/builder.rs
npmccallum: Yes

crt0stack/src/builder.rs
npmccallum: While a nice idea, I think that goes against Rust idiom.

npmccallum:
@haraldh Rebased. No other changes.

haraldh (approved):

crt0stack/src/entry.rs
```diff
@@ -0,0 +1,74 @@
...
+    HwCap(usize),
+
+    /// frequency at which times() increments
+    ClockTick(usize),
+
+    /// secure mode boolean
+    Secure(bool),
+
+    /// string identifying real platform, may differ from Platform.
+    BasePlatform(&'a str),
```
lkatalin, line 58: What is a "real platform"?

haraldh: So, some 64bit platforms offer a 32bit mode, which then is the `Platform`. The `BasePlatform` would be the 64bit one.

lkatalin: Interesting... thanks, @haraldh !

PR #847, opened 2020-08-18
Merge shim crates into enarx-keep

haraldh:
Extends and replaces https://github.com/enarx/enarx/pull/839

@npmccallum wrote:

This experiment merges `enarx-keep-sgx-shim` and `enarx-keep-sev-shim` into
`enarx-keep`. Then `enarx-keep` grows a build script which builds and installs
the shim binaries into `OUT_DIR`.  Finally, the shim binaries are embedded
into the `enarx-keep` binary via `include_bytes!()`.

The major advantage of this approach is that we can `cargo install enarx-keep`
and the resulting binary is usable. There is no need to fiddle with paths or
other complications. This is a great user experience. It further means we
could merge in `integration-tests` into `enarx-keep` (as a further pull
request) and `enarx-keep` becomes entirely self-contained. It could
potentially be forked off into an independenty repository. Likewise,
`cargo run` can be made to work in a straightforward way.

There are some downsides. First, we don't get any build output from the build
script unless builds fail. Because the shims have dependencies, this could
mean we're waiting on silence for a while. Second, utilities like `cargo fmt`
or `cargo check` wouldn't directly propagate to the shims. So we'd have to
work around this a bit in the CI workflows. The biggest example of this is
`cargo test`. Although integration tests would run, unit tests within the
shims themselves would not run unless you did a manual `cargo test` in the
right directories. Third, this kind of setup might confused some IDEs.
However, I suspect it is no more confusing than our existing `cargo make`
setup; in fact, it might be potentially less confusing.


enarx-keep/build.rs
```diff
@@ -32,34 +34,94 @@ fn main() {
...
+            .map(Stdio::from)
+            .unwrap_or_else(|_| Stdio::inherit());
+
+        let stderr: Stdio = OpenOptions::new()
+            .write(true)
+            .open("/dev/tty")
+            .map(Stdio::from)
+            .unwrap_or_else(|_| Stdio::inherit());
+
+        let status = Command::new("cargo")
```
npmccallum, line 96: We probably need to use `std::env::var("CARGO")` here.

enarx-keep/shims/shim-sgx/src/hostlib.rs
```diff
@@ -1,34 +1,39 @@
...
-//! The common Layout for `enarx-keep-sgx` and `enarx-keep-sgx-shim`
-
-#![no_std]
-#![deny(clippy::all)]
-#![deny(missing_docs)]
+// The common Layout for `enarx-keep-sgx` and `enarx-keep-sgx-shim`
 
 use bounds::{Contains as _, Line, Span};
 
+#[allow(dead_code)]
```
npmccallum, line 7:
```suggestion
#![allow(dead_code)]
```

This would avoid a lot of redundancy.

npmccallum: Actually, I've changed my mind. We shouldn't do this at all. There should be no dead code in the shared interfaces at all. That means one side isn't using it and it shouldn't be in the shared file.

npmccallum: In fact, I thought I moved this code into the loader backend...

enarx-keep/shims/shim-sev/.cargo/config
```diff
@@ -2,5 +2,6 @@
 target = "x86_64-unknown-linux-musl"
 rustflags = [
     "-C", "relocation-model=pic",
+    "-C", "link-args=-Wl,--sort-section=alignment",
```
npmccallum, line 5: This should be a separate PR. We can merge this immediately.

enarx-keep/shims/shim-sgx/src/hostlib.rs
haraldh: fixed

enarx-keep/shims/shim-sev/.cargo/config
haraldh: will be, after this one

enarx-keep/build.rs
haraldh: can be a follow-up PR

PR #854, opened 2020-08-19
Allocate the exception stacks with page guards

haraldh:
- factor out `init_stack_with_guard` to be used by the shim and payload
- dynamically allocate the exception stacks with page guards

### DIFF
--- a/enarx-keep/shims/shim-sev/src/gdt.rs
+++ b/enarx-keep/shims/shim-sev/src/gdt.rs
@@ -2,10 +2,10 @@
 
 //! Global Descriptor Table init
 
-use crate::shim_stack::init_shim_stack;
+use crate::shim_stack::{init_stack_with_guard, GuardedStack};
 use crate::syscall::_syscall_enter;
 use core::ops::Deref;
-use memory::Page;
+use nbytes::bytes;
 use x86_64::instructions::segmentation::{load_ds, load_es, load_fs, load_gs, load_ss, set_cs};
 use x86_64::instructions::tables::load_tss;
 use x86_64::registers::model_specific::{KernelGsBase, LStar, SFMask, Star};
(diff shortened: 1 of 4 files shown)


enarx-keep/shims/shim-sev/src/gdt.rs
```diff
@@ -2,46 +2,61 @@
...
+use x86_64::{align_up, PrivilegeLevel, VirtAddr};
 
-const STACK_NUM_PAGES: usize = 5;
-static mut STACKS: [[Page; STACK_NUM_PAGES]; 7] = [[Page::zeroed(); STACK_NUM_PAGES]; 7];
+/// The virtual address of the main kernel stack
+pub const SHIM_STACK_START: u64 = 0xFFFF_FF48_4800_0000;
+
+/// The size of the main kernel stack
+#[allow(clippy::integer_arithmetic)]
+pub const SHIM_STACK_SIZE: u64 = bytes![4; MiB];
```
npmccallum, line 25: Should we integrate this clippy allow into the bytes macro directly?

haraldh: @npmccallum That would be perfect!

npmccallum: https://github.com/enarx/nbytes Patches welcome!

haraldh: I just don't know how

enarx-keep/shims/shim-sev/src/shim_stack.rs
```diff
@@ -5,26 +5,29 @@
...
+    stack_size: u64,
+    extra_flags: PageTableFlags,
+) -> GuardedStack {
     // guard page
     FRAME_ALLOCATOR
         .write()
         .allocate_and_map_memory(
             SHIM_PAGETABLE.write().deref_mut(),
-            VirtAddr::new(SHIM_STACK_START - Page::<Size4KiB>::SIZE),
+            start - Page::<Size4KiB>::SIZE,
```
connorkuehl, line 30: @haraldh Should this be `start + Page::<Size4KiB>::SIZE`? It looks like this guard page overlaps with the first page of the stack. The stack pointer below will be set to this guard page, rather than the usable stack area.

haraldh: yeah.. it's one page before and one page after the usable stack

PR #802, opened 2020-08-10
Keep manager

MikeCamel:
Initial pull request, fixes #728.  Fixes #808.  Fixes #807. Fixes #809. Fixes #849. Major files include:

- keep-runtime/main.rs - a runtime which creates a self-signed certificate, and then listens on a provided port for a wasm file. 
- host-components/main.rs - a keep-manager which creates and communicates with keeps over HTTPS.  A self-signed certificate is provided.  Keep-loaders are created using systemctl
- host-components/keep-loader - a keep-loader which creates run-times (only wasm supported currently).  It is created by keep-manager and listens on a Unix socket
- host-components/keep-manager-tester - a simple client for the keep-manager.  A self-signed certificate is provided
- host-components/app-loader-tester - a simple client for the runtime
- host-components/lib.rs - shared structs and consts for the host-component files.  NOTE - hard-coded path!
- host-components/external/enarx-keep@.service - a systemctl unit file: NOTE: hard-coded path!

All communications are in JSON, and those between keep-manager and its client, and the runtime and its client are encrypted. The communications between the runtime and its client require client and server certificates to be provided.
Logging and error messages from keep-loaders are to hard-coded files.

PR #858, opened 2020-08-20
Add a socket for preattestation

npmccallum:
Some technologies, like SEV, SEV-ES and PEF, do attestation during the creation of the keep rather than afterwards. These technologies need a way to talk to a client during the keep build process. The general expectation is that if no socket is specified and the technology does preattestation, it should bring up the backend either without attestation or with simulated attestation.

### DIFF
--- a/enarx-keep/src/backend/mod.rs
+++ b/enarx-keep/src/backend/mod.rs
@@ -8,6 +8,7 @@ mod probe;
 
 use crate::binary::Component;
 
+use std::path::Path;
 use std::sync::Arc;
 
 use anyhow::Result;
@@ -26,7 +27,7 @@ pub trait Backend {
     fn data(&self) -> Vec<Datum>;
 
     /// Create a keep instance on this backend
-    fn build(&self, code: Component) -> Result<Arc<dyn Keep>>;
+    fn build(&self, code: Component, sock: Option<&Path>) -> Result<Arc<dyn Keep>>;
 }
 
 pub struct Datum {
--- a/enarx-keep/src/main.rs
+++ b/enarx-keep/src/main.rs
@@ -14,6 +14,8 @@ use binary::Component;
 use anyhow::Result;
 use structopt::StructOpt;
 
+use std::path::PathBuf;
+
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const AUTHORS: &str = env!("CARGO_PKG_AUTHORS");
 
@@ -28,8 +30,12 @@ struct Exec {
     #[structopt(short, long)]
     keep: Option<String>,
 
+    /// The socket to use for preattestation
+    #[structopt(short, long)]
+    sock: Option<PathBuf>,
+
     /// The payload to run inside the keep
-    code: String,
+    code: PathBuf,
 }
 
 #[derive(StructOpt)]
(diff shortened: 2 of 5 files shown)


npmccallum:
@MikeCamel This implements what we discussed yesterday. When preattestation is required, `enarx-keep` will connect to the specified socket and proxy backend-specific requests/responses.

The `enarx-keepmgr` can spawn an `enarx-keep` instance for each keep, with support for preattestation. As soon as the keep is instantiated (i.e. preattestation is complete) the socket connection is dropped.

enarx-keep/src/backend/mod.rs
```diff
@@ -43,6 +44,16 @@ pub struct Datum {
     pub mesg: Option<String>,
 }
 
+#[allow(dead_code)]
+pub enum State {
+    Building(Box<dyn Builder>),
+    Complete(Arc<dyn Keep>),
```
MikeCamel, line 50: I think that "Complete" is likely to be a little coarse, and that we'll probably need more states after that, but this looks good for now.

enarx-keep/src/binary/component.rs
```diff
@@ -47,7 +47,7 @@ impl Component {
 
         // Validate header assumptions.
         assert_eq!(elf.header.e_machine, EM_X86_64);
-        assert_eq!(elf.header.e_version, EV_CURRENT as _);
+        assert_eq!(elf.header.e_version, u32::from(EV_CURRENT));
```
MikeCamel, line 50: I'm going to be pernickety and ask for some documentation of the different constants here.  Did some quick grepping at the /enarx level, and didn't easily find a single source of truth.  I'm assuming that it comes from the goblin crate, but I think it's fair to assume that readers may not easily map these constants.

MikeCamel (changes requested):

npmccallum:
@MikeCamel I decided to take a simpler approach and just pass the socket into the `build()` method. This means that backends can simulate attestation if no socket is provided, which can be useful for some cases.

enarx-keep/src/main.rs
```diff
@@ -14,6 +14,8 @@ use binary::Component;
 use anyhow::Result;
 use structopt::StructOpt;
 
+use std::path::PathBuf;
```
MikeCamel, line 17: If we're just going to be using a UnixListener (which I suspect we are), there's no need for an explicit path.   You can create a new one with a cast String (see https://github.com/enarx/enarx/pull/802/files#diff-4ceb5185f3d05238cb8d7553500d56f5)
https://doc.rust-lang.org/beta/std/os/unix/net/struct.UnixListener.html#method.bind

MikeCamel (approved):
To your previous, @npmccallum, yes, I noticed passing the socket via the build() method: makes sense to me.  None of my comments should be a blocker to approval.

enarx-keep/src/main.rs
npmccallum: @MikeCamel We aren't using a `UnixListener`. We're doing `UnixStream::connect()` which takes a path. `enarx-keepmgr` should create the `UnixListener` and `enarx-keep` should receive the address to the socket it should connect to. All details of the protocol over that socket are backend specific.

enarx-keep/src/main.rs
```diff
@@ -93,7 +99,7 @@ fn exec(backends: &[Box<dyn Backend>], opts: Exec) -> Result<()> {
         .find(|b| b.have())
         .expect("No supported backend found!");
 
-    let keep = backend.build(code)?;
+    let keep = backend.build(code, opts.sock.as_deref())?;
```
haraldh, line 102: TIL: `.as_deref()` .. since 1.40 .. yeah!

haraldh (approved):

PR #874, opened 2020-08-25
sgx: don't use deprecated methods

haraldh:


npmccallum:
@haraldh You need to do a `cargo make clean`. Your build is still using the patched version of openssl.

haraldh:
doh!

PR #876, opened 2020-08-25
Two improvements to the sgx crate

npmccallum:


sgx/src/enclave/enclave.rs
```diff
@@ -10,16 +10,14 @@ use mmap::Unmap;
...
 #[repr(u32)]
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum Leaf {
-    /// EAX = 2, `EENTER`
+pub enum Entry {
+    /// Enter an enclave normally
     Enter = 2,
 
-    /// EAX = 3, `ERESUME`
+    /// Resume an enclave after an asynchronous exit
```
lkatalin, line 20:
```suggestion
    /// Resume an enclave after an asynchronous exit.
    /// In a Keep, this is used after a proxied syscall is complete.
```

sgx/src/enclave/enclave.rs
```diff
@@ -141,41 +115,63 @@ impl Thread {
...
         const EEXIT: i32 = 0;
 
         let mut exc = MaybeUninit::<ExceptionInfo>::uninit();
+        let mut rdx: usize = arg.into().into();
+        let rax: i32;
+
+        unsafe {
+            asm!(
+                "push rbp",
+                "mov  rbp, rsp",
```
lkatalin, line 145:
```suggestion
                "mov  rsp, rbp",
```
Is the function prologue not that the base pointer should take the current stack pointer's value?

sgx/src/enclave/enclave.rs
```diff
@@ -141,41 +115,63 @@ impl Thread {
...
+                "and  rsp, ~0xf",
```
lkatalin, line 146: The bitwise arithmetic cuts off the least significant digit of the hex value of `rsp`, but why?

sgx/src/enclave/enclave.rs
```diff
@@ -141,41 +115,63 @@ impl Thread {
...
+
+                "push 0",  // align
+                "push 0",  // exit handler
+                "push {}", // exception info
+                "push {}", // tcs
+                "call {}", // vdso
+
+                "mov  rsp, rbp",
```
lkatalin, line 154: Same question as above.

sgx/src/enclave/enclave.rs
npmccallum: This library doesn't know anything about keeps. So we can't mention them in documentation.

sgx/src/enclave/enclave.rs
npmccallum: You're thinking AT&T syntax. This is Intel syntax. So: `mov DST, SRC`.

lkatalin: Never mind, I see the inline asm operand syntax is backwards compared to AT&T.

sgx/src/enclave/enclave.rs
npmccallum: Same answer.

sgx/src/enclave/enclave.rs
lkatalin: Is this just creating space on the stack?

lkatalin (dismissed):
Changing to "approve," as my last question is more from curiosity.

sgx/src/enclave/enclave.rs
lkatalin: @npmccallum Yep, just saw that.

sgx/src/enclave/enclave.rs
lkatalin: @npmccallum Fair!

sgx/src/enclave/enclave.rs
npmccallum: To ensure stack alignment to 16 bytes.

With inline assembly, the value of `rsp` is guaranteed to be valid but everything else about it is undefined. So this code works like this:

1. Save `rbp` on the stack and save `rsp` in `rbp`.
2. Align the stack to 16 bytes.
3. Pad the stack.
4. Push function arguments to the stack.
5. Call the function (stack **MUST** be 16-bytes aligned here).
6. Restore `rsp` and `rbp` to their original values.

lkatalin: Thanks @npmccallum

npmccallum:
@lkatalin I added some additional comments to make the assembly easier to understand.

sgx/src/enclave/enclave.rs
```diff
@@ -141,41 +118,71 @@ impl Thread {
         Some(Self { enc, tcs, fnc })
     }
 
-    /// Issues `EENTER` instruction to the enclave.
+    /// Enter an enclave.
     ///
-    /// TODO add more comprehensive docs
+    /// This function enters an enclave using `Leaf` and provides the
```
connorkuehl, line 123: I think the reference to `Leaf` should be dropped here.

npmccallum: Fixed.

Issue #818, opened 2020-08-12
Add high level SEV launch API to sev


haraldh:
Turn on SEV encryption and inject a secret in the keep.

- [x] [sev] Add LaunchStart ioctl
- [x] [sev] LaunchUpdateData ioctl
~~[sev] Add LaunchUpdateVmsa ioctl~~ (SEV-ES+)
- [x] [sev] Add LaunchSecret ioctl
- [x] [sev] Add LaunchMeasure ioctl
- [x] [sev] Add LaunchFinish ioctl
- [x] [sev] Create high level interface for launching guest

"Real" attestation may be out of scope for this issue, but is detailed here just in case:

Incorporate these elements into keep construction:

- Verify certificate chain
- Craft an SEV policy
- Measure the VM's address space
- Inject code

PR #879, opened 2020-08-25
Add SEV launch API to sev

connorkuehl:
Related: #818


connorkuehl:
> 2020-08-25T22:51:11.7488882Z thread 'sev' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 5, kind: Other, message: "Input/output error" }', tests/launch.rs:98:24

Well that's new :-) I'll dig into this tomorrow.

connorkuehl:
Now that I've finally chased down (I think) the weird errors that have surfaced when I pushed this, I want to start a discussion specifically regarding how I've integrated the SEV launch into the unit tests and the consequences that I am aware of.

I've made the SEV launch test mutually exclusive with the tests gated behind "dangerous_tests". I learned the hard way that `cargo test`s multiple threads running different tests will allow things like a platform reset mid-launch, which explains the initial errors I started seeing when the PR was first opened. This seemed like a straightforward way to avoid having some hairy lock in our test modules. (Something like this would technically also work, but is not expressible in a cargo config to my knowledge: `cargo test --all-features -- --test-threads=1`)

An alternative to this would be using the [`serial_test`](https://crates.io/crates/serial_test) crate. I did perform some testing with this, and it seemed to work. However, if we go this route we'll want to do so after we split 'sev' off and drop the dependency on `cargo-make` for building. I'll explain why in a moment.

There's another big issue which is also slightly ameliorated by splitting 'sev' off and stuffing it in its own repo. AMD's web service for getting a signed CEK will rate limit us. This is a problem for single CI runs because our cargo-make script would actually end up invoking it twice (this is because our `deep-ci` task wraps the `ci`task) which would cause a spurious failure due to AMD rejecting the request for rate-limiting reasons. So, dropping `cargo-make` during the split will at least remove 1 guaranteed spurious failure due to rate-limiting.

However, it's entirely possible that even after the split we receive enough pull requests fast enough in a row such that we get rate-limited and some of them spuriously fail and must be retriggered. Not the end of the world, but certainly not a good look, and still brittle.


I think my preferred way forward would be to unconditionally ignore the tests which modify the platform state. This would mean we can export the CI chain for the CI machine (perhaps in the kickstart(?)) and avoid making an HTTP request to AMD on every push to a PR. The test could simply use the cached certificate chain. These tests are testing stable code, and there's nothing stopping us from running them manually if development ever calls for it. This strikes me as substantially less brittle than everything I just described above.

connorkuehl:
Some of these commits are certainly orthogonal. I'm going to break this PR down into multiple to avoid code review fatigue. Sorry for the noise.

connorkuehl:
`sev` is splitting off the main repo, I will file more up-to-date patches there.

PR #883, opened 2020-08-26
Move integration tests into enarx-keep

lkatalin:
To do: test `read()`

### DIFF
--- a/enarx-keep/build.rs
+++ b/enarx-keep/build.rs
@@ -5,6 +5,9 @@ use std::fs::OpenOptions;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
+const CRATE: &str = env!("CARGO_MANIFEST_DIR");
+const TEST_BINS_IN: &str = "tests/bin";
+
 fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
     for entry in std::fs::read_dir(path)? {
         let path = entry?.path();
@@ -30,6 +33,36 @@ fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
     Ok(())
 }
 
+fn build_tests(in_path: &Path, out_path: &Path) {
+    let test_in_dir = Path::new(CRATE).join(TEST_BINS_IN);
+
+    for entry in in_path.read_dir().expect("failed to read tests/bin dir") {
+        let file = entry.expect("failed to read file in tests/bin dir");
+        let filename = file.file_name();
+        let output = file.path().file_stem().unwrap().to_os_string();
+
+        let mut cmd = cc::Build::new()
+            .no_default_flags(true)
+            .get_compiler()
+            .to_command();
+
+        let status = cmd
+            .current_dir(&out_path)
+            .arg("-nostdlib")
+            .arg("-static-pie")
+            .arg("-fPIC")
+            .arg("-o")
+            .arg(output)
+            .arg(test_in_dir.join(&filename))
+            .status()
+            .expect("failed to compile binary");
+
+        if !status.success() {
+            panic!("Failed to compile {:?}", &filename);
+        }
+    }
+}
+
 fn main() {
     println!("cargo:rerun-if-env-changed=OUT_DIR");
     println!("cargo:rerun-if-env-changed=PROFILE");
@@ -45,6 +78,8 @@ fn main() {
         }
         Ok(_) => {}
     }
+
+    build_tests(&Path::new(CRATE).join(TEST_BINS_IN), &out_dir_bin);
 
     let profile: &[&str] = match std::env::var("PROFILE").unwrap().as_str() {
         "release" => &["--release"],
--- a/enarx-keep/tests/bin/clock_gettime.c
+++ b/enarx-keep/tests/bin/clock_gettime.c
@@ -1,0 +1,20 @@
+#include <sys/syscall.h>
+#include <time.h>
+
+
+void _start(void) {
+    struct timespec t;
+    asm(
+        "syscall"
+        : // no ouputs
+        : "a" (SYS_clock_gettime), "D" (CLOCK_MONOTONIC), "S" (&t)
+        : // no clobbers
+    );
+
+    asm(
+        "syscall; ud2"
+        : // no outputs
+        : "a" (SYS_exit), "D" (0)
+        : // no clobbers
+    );
+}
--- a/enarx-keep/tests/bin/exit_one.c
+++ b/enarx-keep/tests/bin/exit_one.c
@@ -1,0 +1,12 @@
+#include <sys/syscall.h>
+
+#define STATUS 1
+
+void _start(void) {
+    asm(
+        "syscall; ud2"
+        : // no ouputs
+        : "a" (SYS_exit), "D" (STATUS)
+        : // no clobbers
+    );
+}
--- a/enarx-keep/tests/bin/exit_zero.c
+++ b/enarx-keep/tests/bin/exit_zero.c
@@ -1,0 +1,12 @@
+#include <sys/syscall.h>
+
+#define STATUS 0
+
+void _start(void) {
+    asm(
+        "syscall; ud2"
+        : // no ouputs
+        : "a" (SYS_exit), "D" (STATUS)
+        : // no clobbers
+    );
+}
--- a/enarx-keep/tests/bin/write_stderr.c
+++ b/enarx-keep/tests/bin/write_stderr.c
@@ -1,0 +1,21 @@
+#include <sys/syscall.h>
+#include <unistd.h>
+
+void _start(void) {
+    char s[3] = "hi\n";
+    int len = 3;
+
+    asm(
+        "syscall"
+        : // no ouputs
+        : "a" (SYS_write), "D" (STDERR_FILENO), "S" (&s), "d" (len)
+        : // no clobbers
+    );
+
+    asm(
+        "syscall; ud2"
+        : // no outputs
+        : "a" (SYS_exit), "D" (0)
+        : // no clobbers
+    );
+}
--- a/enarx-keep/tests/bin/write_stdout.c
+++ b/enarx-keep/tests/bin/write_stdout.c
@@ -1,0 +1,21 @@
+#include <sys/syscall.h>
+#include <unistd.h>
+
+void _start(void) {
+    char s[3] = "hi\n";
+    int len = 3;
+
+    asm(
+        "syscall"
+        : // no ouputs
+        : "a" (SYS_write), "D" (STDOUT_FILENO), "S" (&s), "d" (len)
+        : // no clobbers
+    );
+
+    asm(
+        "syscall; ud2"
+        : // no outputs
+        : "a" (SYS_exit), "D" (0)
+        : // no clobbers
+    );
+}
--- a/enarx-keep/tests/integration_tests.rs
+++ b/enarx-keep/tests/integration_tests.rs
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io::Read;
+use std::path::Path;
+use std::process::{Command, Stdio};
+use std::time::Duration;
+use wait_timeout::ChildExt;
+
+const CRATE: &str = env!("CARGO_MANIFEST_DIR");
+const KEEP_BIN: &str = env!("CARGO_BIN_EXE_enarx-keep");
+const OUT_DIR: &str = env!("OUT_DIR");
+const TEST_BINS_OUT: &str = "bin";
+const TIMEOUT_SECS: u64 = 5;
+
+/// Returns a handle to a child process through which output (stdout, stderr) can
+/// be accessed.
+fn run_test(bin: &str, status: i32) -> std::process::Child {
+    let bin = Path::new(CRATE).join(OUT_DIR).join(TEST_BINS_OUT).join(bin);
+
+    let mut child = Command::new(&String::from(KEEP_BIN))
+        .current_dir(CRATE)
+        .arg("exec")
+        .arg(bin)
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("failed to run test bin");
+
+    let code = match child
+        .wait_timeout(Duration::from_secs(TIMEOUT_SECS))
+        .unwrap()
+    {
+        Some(status) => status.code().unwrap(),
+        None => {
+            child.kill().unwrap();
+            panic!("error: test timeout");
+        }
+    };
+
+    assert_eq!(code, status);
+    child
+}
+
+#[test]
+fn exit_zero() {
+    run_test("exit_zero", 0);
+}
+
+#[test]
+fn exit_one() {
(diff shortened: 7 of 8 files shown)


enarx-keep/build.rs
```diff
@@ -30,21 +33,78 @@ fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
...
+    create_dir(&asm_out_dir);
+
+    for entry in path.read_dir().expect("failed to read test-bin dir") {
+        let file = entry.expect("failed to read file in test-bin dir");
+        let f = file.file_name();
+        let filename = f.to_str().unwrap();
+        let output: &str = filename.split('.').collect::<Vec<&str>>()[0];
+
+        std::process::Command::new(&compiler)
+            .current_dir(&asm_out_dir)
```
haraldh, line 85:
```suggestion
        let mut cmd = cc::Build::new()
            .no_default_flags(true)
            .get_compiler()
            .to_command();

        cmd.current_dir(&asm_out_dir)
```

haraldh (changes requested):

enarx-keep/src/backend/kvm/vm/x86_64.rs
```diff
@@ -15,6 +15,7 @@ pub struct VMSetup<'a> {
 }
 
 impl VMSetup<'_> {
+    #[allow(dead_code)]
```
haraldh, line 18: please don't ... let @connorkuehl decide

haraldh:
Needs https://github.com/enarx/enarx/pull/885 before the CI succeeds on KVM/SEV

enarx-keep/src/backend/kvm/vm/x86_64.rs
lkatalin: Wouldn't dream of editing the `kvm` crate normally, but it was emitting a warning that was showing up in my `stderr` test. Maybe there is another way to keep from picking that up.

haraldh: How so? Is it recompiling everything?

connorkuehl: My opinion here is that we should just remove the dead code rather than `#[allow(dead_code)]` -- this function was superseded by one of Harald's PRs anyway so it's no longer required.

I'm also surprised to hear that the compilation process of the crate is polluting the integration tests stderr. I think that should be addressed directly.

lkatalin: I was having a lot of trouble with extraneous cargo output in the `stderr` test generally, and this happened pretty late in the day, so between those two things it's likely that the error is on my end. I'll remove this change and see if it still crops up.

enarx-keep/build.rs
lkatalin: Thanks @haraldh , that seemed to work.

enarx-keep/build.rs
```diff
@@ -30,21 +33,56 @@ fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
...
+fn create_dir(path: &PathBuf) {
+    match std::fs::create_dir(&path) {
+        Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {}
+        Err(e) => {
+            eprintln!("Can't create {:#?} : {:#?}", path, e);
+            std::process::exit(1);
+        }
+        Ok(_) => {}
+    }
+}
```
haraldh, line 45: might be sufficient to call `std::fs::create_dir_all()` instead of `create_dir()`

enarx-keep/build.rs
```diff
@@ -30,21 +33,56 @@ fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
...
+
+        cmd.current_dir(&asm_out_dir)
+            .arg("-nostdlib")
+            .arg("-static-pie")
+            .arg("-fPIC")
+            .arg("-o")
+            .arg(output)
+            .arg(test_dir.join(filename))
+            .status()
+            .expect("failed to compile binary");
```
haraldh, line 71: you need to check for the status also, as it can be a return value != 0

haraldh:
```rust
    if !status.success() {
        panic!("Failed to compile {}", filename);
    }
```

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,173 @@
...
+        .current_dir(CRATE)
+        .arg("+nightly")
+        .env_clear()
+        .envs(filtered_env)
+        .arg("run")
+        .arg("-q")
+        .arg("exec")
+        .arg(bin)
+        .spawn()
+        .expect("failed to run test bin");
```
npmccallum, line 77: This is why you're having so much problem with extra output and having to work around it with a bunch of hacks. When `cargo test` is run, the binary is already built before the tests are run. You just need to execute it directly rather than using `cargo +nightly run`.

I created a [working example of this strategy](https://github.com/npmccallum/demo/). We execute the binary [like this](https://github.com/npmccallum/demo/blob/master/tests/test.rs#L7-L8). This gives no extra output we have to work around.

In short, if you run the binary directly and don't use `cargo run` you can probably remove half the code in this commit which works around the problems of using `cargo run`.

npmccallum:
One other thing I forgot: your `.s` files have mixed tabs and spaces.

enarx-keep/tests/test-bin/clock_gettime.s
```diff
@@ -0,0 +1,17 @@
...
+ 	lea ts(%rip),   %rsi /* timespec */
+        syscall
+
+        mov $60,   	%rax /* SYS_exit */
+        mov $0,    	%rdi /* exit status */
+        syscall
+
+.data
+ts:
+	.space 128
```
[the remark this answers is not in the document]
lkatalin: @npmccallum Why is it preferable to put this on the stack?

enarx-keep/build.rs
```diff
@@ -30,21 +33,56 @@ fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
     Ok(())
 }
 
+fn create_dir(path: &PathBuf) {
+    match std::fs::create_dir(&path) {
+        Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {}
+        Err(e) => {
+            eprintln!("Can't create {:#?} : {:#?}", path, e);
+            std::process::exit(1);
+        }
```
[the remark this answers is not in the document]
lkatalin: @npmccallum I changed the code to use `std::fs::create_dir_all()` instead of this separate function based on @haraldh 's suggestion, since the former will not error if a directory already exists. Is there a reason we should *not* do it that way?

enarx-keep/tests/integration_tests.rs
lkatalin: @npmccallum Thank you!!!

lkatalin:
This new version ended up not needing @haraldh 's code since @npmccallum pointed out the better way to run the test is not using `cargo`.

enarx-keep/build.rs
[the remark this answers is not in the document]
lkatalin: Just kidding, I ended up leaving the original code as-is.

enarx-keep/tests/bin/clock_gettime.c
```diff
@@ -0,0 +1,21 @@
+#include <sys/syscall.h>
+#include <time.h>
+
+struct timespec t;
+
+void _start(void) {
```
haraldh, line 6:
```suggestion
__attribute__((naked)) void _start(void) {
```

gives better generated code with `ud2` after the last syscall:

```asm
0000000000001000 <_start>:
    1000:	ba e4 00 00 00       	mov    $0xe4,%edx
    1005:	b9 01 00 00 00       	mov    $0x1,%ecx
    100a:	48 8d 05 ef 2f 00 00 	lea    0x2fef(%rip),%rax        # 4000 <t>
    1011:	48 89 c6             	mov    %rax,%rsi
    1014:	89 d0                	mov    %edx,%eax
    1016:	89 cf                	mov    %ecx,%edi
    1018:	0f 05                	syscall 
    101a:	b8 3c 00 00 00       	mov    $0x3c,%eax
    101f:	ba 00 00 00 00       	mov    $0x0,%edx
    1024:	89 d7                	mov    %edx,%edi
    1026:	0f 05                	syscall 
    1028:	90                   	nop
    1029:	0f 0b                	ud2    
```

enarx-keep/tests/bin/exit_one.c
```diff
@@ -0,0 +1,12 @@
+#include <sys/syscall.h>
+
+#define STATUS 1
+
+void _start(void) {
```
haraldh, line 5:
```suggestion
__attribute__((naked)) void _start(void) {
```

enarx-keep/tests/bin/exit_zero.c
```diff
@@ -0,0 +1,12 @@
+#include <sys/syscall.h>
+
+#define STATUS 0
+
+void _start(void) {
```
haraldh, line 5:
```suggestion
__attribute__((naked)) void _start(void) {
```

enarx-keep/tests/bin/write_stderr.c
```diff
@@ -0,0 +1,22 @@
+#include <sys/syscall.h>
+#include <unistd.h>
+
+void _start(void) {
```
haraldh, line 4:
```suggestion
__attribute__((naked)) void _start(void) {
```

enarx-keep/tests/bin/write_stdout.c
```diff
@@ -0,0 +1,22 @@
+#include <sys/syscall.h>
+#include <unistd.h>
+
+void _start(void) {
```
haraldh, line 4:
```suggestion
__attribute__((naked)) void _start(void) {
```

haraldh (changes requested):

enarx-keep/tests/bin/clock_gettime.c
lkatalin: Thanks @haraldh , just updated this. What does the `ud2` get us? Why do we want an invalid opcode after the last syscall?

lkatalin: @haraldh Looks like the naked function does not want to work with the strings:  `cannot allocate stack for variable s, naked function`.

haraldh: > @haraldh Looks like the naked function does not want to work with the strings: `cannot allocate stack for variable s, naked function`.

@lkatalin boooh.. bad compiler.. bad suggestion of me :-1:

haraldh: just add `ud2` manually

npmccallum: @lkatalin `ud2` always causes a CPU fault. Basically, it is the manual way of guaranteeing that the `exit` syscall never returns without a CPU fault.

lkatalin: Makes sense now!

haraldh (approved):

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+        return status.success();
+    } else {
+        cmd.kill().unwrap();
+        panic!("error: test timeout");
+    }
+}
+
+/// Returns a handle to a child process through which output (stdout, stderr) can
+/// be accessed.
+fn run_test(bin: PathBuf) -> std::process::Child {
```
npmccallum, line 42:
```suggestion
fn run_test(bin: &str, code: Option<i32>) -> std::process::Child {
```

See below...

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("failed to run test bin");
+
+    if let Ok(None) = cmd.wait_timeout(Duration::from_secs(TIMEOUT_SECS)) {
+        cmd.kill().unwrap();
+        panic!("error: test timeout");
+    } else {
+        cmd
+    }
```
npmccallum, line 58:
```suggestion
    }
    
    assert_eq!(cmd.code(), code);
    cmd
```

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+        .spawn()
+        .expect("failed to run test bin");
+
+    if let Some(status) = cmd.wait_timeout(Duration::from_secs(TIMEOUT_SECS)).unwrap() {
+        return status.success();
+    } else {
+        cmd.kill().unwrap();
+        panic!("error: test timeout");
+    }
+}
```
npmccallum, line 38: Remove this whole function.

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+const OUT_DIR: &str = env!("OUT_DIR");
+const TEST_BINS_OUT: &str = "bin";
+const TIMEOUT_SECS: u64 = 5;
+
+fn path(input: &str) -> PathBuf {
+    Path::new(CRATE)
+        .join(OUT_DIR)
+        .join(TEST_BINS_OUT)
+        .join(input)
+}
```
npmccallum, line 20: Merge this function into `run_test()` below.

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+        panic!("error: test timeout");
+    } else {
+        cmd
+    }
+}
+
+#[test]
+fn exit_zero() {
+    let code = path("exit_zero");
+    assert!(run_succeeds(code));
```
npmccallum, line 64:
```suggestion
    run_test("exit_zero", Some(0));
```

npmccallum: ... and so forth for all tests.

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io::Read;
+use std::path::Path;
+use std::process::{Command, Stdio};
+use std::time::Duration;
+use wait_timeout::ChildExt;
+
+const CRATE: &str = env!("CARGO_MANIFEST_DIR");
+const KEEP_BIN: &str = "/target/debug/enarx-keep";
```
npmccallum, line 10: This will break the moment we do `cargo test --profile=release`.

enarx-keep/tests/integration_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+const OUT_DIR: &str = env!("OUT_DIR");
+const TEST_BINS_OUT: &str = "bin";
+const TIMEOUT_SECS: u64 = 5;
+
+/// Returns a handle to a child process through which output (stdout, stderr) can
+/// be accessed.
+fn run_test(bin: &str, status: Option<i32>) -> std::process::Child {
```
npmccallum, line 17: Since we always specify an exit status...

```suggestion
fn run_test(bin: &str, status: i32) -> std::process::Child {
```

haraldh (approved):

PR #884, opened 2020-08-27
A new mmap implementation

npmccallum:
This is a new take on the mmap crate. It doesn't currently support all mmap
features. But it does support a lot of them. Most operations are safe.
Ownership of a map is implicitly tracked and `munmap()` is automatically
called. Additionally, we get compile-time validation of memory permissions.

### DIFF
--- a/mmap/src/lib.rs
+++ b/mmap/src/lib.rs
@@ -1,58 +1,493 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use lset::Span;
+//! The `mmap` crate wraps the underlying system `mmap()` call in safe semantics.
+//!
+//! For example:
+//!
+//! ```rust
+//! use std::num::NonZeroUsize;
+//! use mmap::{Kind, Builder, perms};
+//!
+//! let mut zero = std::fs::File::open("/dev/zero").unwrap();
+//!
+//! let map = Builder::map(4096)
+//!     .near(128 * 1024 * 1024)
+//!     .from(&mut zero, 0)
+//!     .known::<perms::Read>(Kind::Private)
+//!     .unwrap();
+//!
+//! assert_eq!(&*map, &[0; 4096]);
+//! ```
+//!
+//! You can also remap an existing mapping:
+//!
+//! ```rust
+//! use std::num::NonZeroUsize;
+//! use mmap::{Kind, Builder, perms};
+//!
+//! let mut zero = std::fs::File::open("/dev/zero").unwrap();
+//!
+//! let mut map = Builder::map(4096)
+//!     .anywhere()
+//!     .from(&mut zero, 0)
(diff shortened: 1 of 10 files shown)


mmap/src/lib.rs
```diff
@@ -1,58 +1,478 @@
...
     }
 }
 
-impl Drop for Unmap {
-    fn drop(&mut self) {
-        unsafe { libc::munmap(self.0.start as *mut _, self.0.count) };
+impl<'a> Builder<Source<'a>> {
+    /// Creates a mapping with unknown (i.e. run-time) permissions
+    ///
+    /// You should use a moderate amount of effort to avoid using thie method.
```
haraldh, line 298: s/thie/this

mmap/src/lib.rs
```diff
@@ -1,58 +1,478 @@
...
+    ///
+    /// This construction may seem strange at first. However, it allows us
+    /// retain the validity of the old mapping if the creation of the new
+    /// mapping fails without the need to return the old mapping in the
+    /// error condition - which breaks the use of the `?` operator.
+    ///
+    /// In short, don't use the old mapping if the remapping is successful.
+    /// If you do, you're still safe, but you'll probably have logic bugs.
+    #[inline]
+    pub fn remap<'a>(&'a mut self) -> Builder<Destination<'a>> {
```
haraldh, line 452:
```suggestion
    pub fn remap(&mut self) -> Builder<Destination> {
```

haraldh (approved):

mmap/src/lib.rs
npmccallum: Fixed.

mmap/src/lib.rs
npmccallum: Fixed.

npmccallum:
I have added support for specifying huge table sizes. I think that's all the additional features we care about.

There is one design choice I'm still unhappy about, but I don't see a way around it. In `remap()` and `reprotect()`, we "steal" the inner mapping for the new mapping with the new settings. This feels very C++ish and not very Rust-ish. But the only other option that I see is to move the map into the function and then return it along with the error code. But this also isn't great because it breaks the try (`?`) operator. I'm inclined to leave it, but I'd love some other opinions.

haraldh:
> There is one design choice I'm still unhappy about, but I don't see a way around it. In `remap()` and `reprotect()`, we "steal" the inner mapping for the new mapping with the new settings. This feels very C++ish and not very Rust-ish. But the only other option that I see is to move the map into the function and then return it along with the error code. But this also isn't great because it breaks the try (`?`) operator. I'm inclined to leave it, but I'd love some other opinions.

I would move the map into the function and then return it along with the error code, as this error path requires special attention anyway.

haraldh (approved):

PR #890, opened 2020-08-27
shim-sev: switch to a guarded stack as soon as possible

haraldh:


enarx-keep/shims/shim-sev/src/main.rs
```diff
@@ -55,9 +55,21 @@ static BOOT_INFO: RwLock<Option<BootInfo>> =
...
+/// Switch the stack and jump to a function
+pub unsafe fn switch_shim_stack(ip: extern "C" fn() -> !, sp: u64) -> ! {
+    asm!("
+        mov      rsp, {0}
+        jmp      {1}
+          ",
+    in(reg) sp,
+    in(reg) ip,
+    options(noreturn, nomem)
+    );
```
npmccallum, line 67:
```suggestion
    asm!(
        "mov rsp, {0}",
        "jmp {1}",
        in(reg) sp,
        in(reg) ip,
        options(noreturn, nomem)
    );
```

haraldh: fixed

PR #893, opened 2020-08-28
Improve the new mmap crate interface

npmccallum:
In particular, we define a new error type that wraps `std::io::Error`
and conveys ownership of an object. This allows us to return the
old mapping along with an error condition without breaking the try
operator.

### DIFF
--- a/mmap/src/lib.rs
+++ b/mmap/src/lib.rs
@@ -85,7 +85,7 @@
 #![deny(missing_docs)]
 
 use std::convert::TryInto;
-use std::io::{Error, ErrorKind, Result};
+use std::io::ErrorKind;
 use std::marker::PhantomData;
 use std::mem::forget;
 use std::os::unix::io::{AsRawFd, RawFd};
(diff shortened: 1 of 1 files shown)


mmap/src/lib.rs
```diff
@@ -155,6 +155,62 @@ enum Address {
...
+        self.err.fmt(f)
+    }
+}
+
+impl<T: std::fmt::Debug> std::error::Error for Error<T> {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.err)
+    }
+
+    fn cause(&self) -> Option<&dyn std::error::Error> {
```
haraldh, line 185: you don't have to implement this... it's already implemented in the trait using `source`

haraldh (changes requested):

mmap/src/lib.rs
npmccallum: fixed

haraldh (approved):

Issue #856, opened 2020-08-19
Merge sallyport into enarx-keep


lkatalin:
The sallyport crate supports enarx-keep. Merging them into the crate will allow for packaging it all in one.

This is the next step after #847 and #855 .

npmccallum:
Fixed in https://github.com/enarx/enarx/commit/92ccf937decf9ce8ad56fa6f21750aca19c88fa4

PR #896, opened 2020-08-31
move sallyport and rcrt1 into enarx-keep

haraldh:
- also move the shims in the `internal` directory
- extend the `build.rs` to cope with `Cargo.tml` -> `Cargo.toml`
  renaming


mbestavros (dismissed):

enarx-keep/build.rs
```diff
@@ -59,22 +59,42 @@ fn main() {
...
+    for entry in std::fs::read_dir("internal").unwrap() {
+        let shim_path = entry.unwrap().path();
+
+        let cargo_toml = shim_path.join("Cargo.toml");
+        let cargo_tml = shim_path.join("Cargo.tml");
+
+        if cargo_tml.exists() {
+            std::fs::copy(cargo_tml, cargo_toml).unwrap();
+        }
+    }
```
npmccallum, line 74: It is my understanding that this modifies files in the source directory. I think that's frowned upon. But... they can continue frowning until `cargo` provides us a better solution.

enarx-keep/build.rs
```diff
@@ -59,22 +59,42 @@ fn main() {
         })
         .collect();
 
-    for entry in std::fs::read_dir("shims").unwrap() {
+    // internal crates are not included, if there is a `Cargo.toml` file
+    // trick cargo by renaming the `Cargo.toml` to `Cargo.tml` before
+    // publishing and rename it back here.
+    for entry in std::fs::read_dir("internal").unwrap() {
+        let shim_path = entry.unwrap().path();
```
npmccallum, line 66:
```suggestion
        let path = entry.unwrap().path();
```

enarx-keep/src/backend/kvm/vm/cpu.rs
```diff
@@ -7,11 +7,11 @@ use crate::backend::kvm::shim::{
     MemInfo, SYSCALL_TRIGGER_PORT, SYS_ENARX_BALLOON_MEMORY, SYS_ENARX_MEM_INFO,
 };
 use crate::backend::{Command, Thread};
+use crate::sallyport;
```
npmccallum, line 10: I'd prefer to stick with `use crate::sallyport::Reply`... Is there some reason for this change?

enarx-keep/src/backend/mod.rs
```diff
@@ -7,12 +7,12 @@ pub mod sgx;
 mod probe;
 
 use crate::binary::Component;
+use crate::sallyport;
```
npmccallum, line 10: `sallyport::Block`

enarx-keep/src/main.rs
```diff
@@ -4,9 +4,14 @@
...
 #![deny(clippy::all)]
 #![deny(missing_docs)]
+#![feature(asm)]
 
 mod backend;
 mod binary;
+mod sallyport;
+
+// workaround for sallyport tests, until we have internal crates
+pub use sallyport::Request;
```
npmccallum, line 14: Why not `pub mod sallyport`?

enarx-keep/build.rs
haraldh: yep

enarx-keep/build.rs
haraldh: what's your problem here?

enarx-keep/src/main.rs
haraldh: doesn't make a difference. The macro in the tests wants `$crate::Request`

enarx-keep/src/backend/kvm/vm/cpu.rs
haraldh: no reason, changed `sallyport::Block` instead then

enarx-keep/src/backend/mod.rs
haraldh: ok

enarx-keep/build.rs
haraldh: ah, oh, right

lkatalin:
Looks like this resolves #856

PR #901, opened 2020-09-01
shim-sev: don't use `impl` in arguments

haraldh:
Using `impl` in arguments prevents using the turbofish at the caller.

### DIFF
--- a/enarx-keep/internal/shim-sev/src/frame_allocator.rs
+++ b/enarx-keep/internal/shim-sev/src/frame_allocator.rs
@@ -353,9 +353,9 @@ impl FrameAllocator {
     }
 
     /// Map physical memory to the given virtual address
-    pub fn map_memory(
+    pub fn map_memory<T: Mapper<Size4KiB> + Mapper<Size2MiB>>(
         &mut self,
-        mapper: &mut (impl Mapper<Size4KiB> + Mapper<Size2MiB>),
+        mapper: &mut T,
         map_from: PhysAddr,
         map_to: VirtAddr,
         size: usize,


haraldh:
> I'm not sure I agree with the commit message. There are times when it _is_ idiomatic. For example `fn read_func(mut reader: impl Read) -> std::io::Result<()>` is completely idiomatic.
> 
> I would agree that this particular usage is not ergonomic, however.

fixed the commit message

enarx-keep/internal/shim-sev/src/frame_allocator.rs
```diff
@@ -353,15 +353,18 @@ impl FrameAllocator {
     }
 
     /// Map physical memory to the given virtual address
-    pub fn map_memory(
+    pub fn map_memory<T>(
```
npmccallum, line 356:
```suggestion
    pub fn map_memory<T: Mapper<Size4KiB> + Mapper<Size2MiB>>(
```

npmccallum: For simple declarations, this is my preference. For more complex situations, us `where`.

haraldh: as you wish

npmccallum (approved):

connorkuehl (approved):
Interesting. I feel like I thought about this way too much. Does this end up being just a stylistic change or will the Rust compiler now be generating code for each concrete `T: Mapper<Size4KiB> + Mapper<Size2MiB>` as opposed to before which is probably (oversimplified) a pointer and dynamic dispatch?

npmccallum:
> Interesting. I feel like I thought about this way too much. Does this end up being just a stylistic change or will the Rust compiler now be generating code for each concrete `T: Mapper<Size4KiB> + Mapper<Size2MiB>` as opposed to before which is probably (oversimplified) a pointer and dynamic dispatch?

@connorkuehl This is just a style change. There is no dynamic dispatch unless you use the `dyn` keyword. Both approaches will monomorphize the function with the concrete type. But with `impl Trait` you don't get to specify the concrete type with turbofish.

PR #903, opened 2020-09-01
various integration test improvements

### LINKED ISSUE #731: Write syscall test should fail on very large write
We should add a test for a very large write (larger than the size of the buffer) that fails and returns `EMSGSIZE`.

### LINKED ISSUE #747: Dynamic output from integration tests

npmccallum:
Resolves #731
Resolves #747

### DIFF
--- a/enarx-keep/tests/bin/write_emsgsize.c
+++ b/enarx-keep/tests/bin/write_emsgsize.c
@@ -1,0 +1,11 @@
+#include "libc.h"
+
+int main(void) {
+    char msg[128 * 1024];
+
+    if (!is_enarx()) {
+        return 0;
+    }
+
+    return write(STDOUT_FILENO, msg, sizeof(msg)) != -EMSGSIZE;
+}
--- a/enarx-keep/tests/bin/clock_gettime.c
+++ b/enarx-keep/tests/bin/clock_gettime.c
@@ -1,20 +1,12 @@
-#include <sys/syscall.h>
-#include <time.h>
+#include "libc.h"
 
-
-void _start(void) {
+int main(void) {
     struct timespec t;
-    asm(
-        "syscall"
-        : // no ouputs
-        : "a" (SYS_clock_gettime), "D" (CLOCK_MONOTONIC), "S" (&t)
-        : // no clobbers
-    );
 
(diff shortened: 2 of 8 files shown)


haraldh (approved):

npmccallum:
@haraldh The `EMSGSIZE` test appears to be failing on your kernel...

haraldh:
> @haraldh The `EMSGSIZE` test appears to be failing on your kernel...

@npmccallum no `sev` passes... the github native container does not have any, so it is using the binary directly, which does not emit `-EMSGSIZE` but happily writes the stuff.

enarx-keep/tests/bin/write_emsgsize.c
```diff
@@ -0,0 +1,22 @@
+#include <sys/syscall.h>
+#include <unistd.h>
+#include <errno.h>
+
+void _start(void) {
+    char msg[128 * 1024];
```
lkatalin, line 6: How is this size chosen for "too large"?

npmccallum: The largest single write that I would consider "reasonable" is the ~64k maximum size of a UDP packet.

lkatalin:
Also looks like it resolves #747 , unless there was more we wanted to do with that @npmccallum ?

npmccallum:
This PR requires https://github.com/enarx/enarx/pull/907

PR #904, opened 2020-09-01
Add test for read syscall

### LINKED ISSUE #855: Merge integration-tests into enarx-keep
The integration tests are testing enarx-keep. Merging them into the crate will allow for packaging it all in one.
The tests should run with `cargo test` and should **not** use `cargo make`.

This is the next step after #847 .

lkatalin:
Resolves #855 

I'll rebase this on top of #903 and adopt its good ideas as soon as that's merged.


npmccallum:
@lkatalin Also, don't forget to follow the practices in the new `libc.h`.

enarx-keep/tests/bin/read.c
```diff
@@ -0,0 +1,22 @@
...
+        if (bytes > 0) {
+           write(STDOUT_FILENO, buf, bytes);
+        } else
+            break;
+
+        len = min(len * 2, 12);
+    }
+
+    return bytes;
+}
```
npmccallum, line 22:
```suggestion
int min(int a, int b) {
    return a < b ? a : b;
}

int main() {
    char buf[16] = {};

    for (size_t in = 1;; in = min(in * 2, sizeof(buf)))  {
        ssize_t out = read(STDIN_FILENO, buf, in);
        if (out <= 0)
            break;

        write(STDOUT_FILENO, buf, out);
    }
}
```

enarx-keep/tests/bin/read.c
```diff
@@ -0,0 +1,22 @@
+#include "libc.h"
+
+int main(void) {
+    int min(int a, int b) { return a < b ? a : b; }
```
npmccallum, line 4: Nested functions are a GCC extension and don't work on all compilers.

enarx-keep/tests/bin/read.c
```diff
@@ -0,0 +1,22 @@
...
+
+    char buf[12] = {'\0'};
```
npmccallum, line 6: This initializer is a bit strange because of what is actually happening. Your initializer is roughly "zero the first byte." Fortunately, C specifies what should happen when some bytes are left unspecified: they are zeroed. Since you're relying on that behavior anyway, it is simpler and cleaner to just do:

```suggestion
    char buf[12] = {};
```

Also, I'm not sure why `12` is used as the size here since it isn't a power of two. But maybe there is a reason.

enarx-keep/tests/bin/read.c
```diff
@@ -0,0 +1,22 @@
...
+    int len = 1;
+    int bytes = 0;
+
+    while (1) {
```
npmccallum, line 10: This can be simplified using a for loop.

enarx-keep/tests/bin/read.c
```diff
@@ -0,0 +1,22 @@
...
+        bytes = read(STDIN_FILENO, buf, len);
+
+        if (bytes > 0) {
+           write(STDOUT_FILENO, buf, bytes);
+        } else
+            break;
```
npmccallum, line 16: As a matter of good C practice, never unbalance your braces. It can lead to weird bugs. Do either this:

```suggestion
        if (bytes > 0)
            write(STDOUT_FILENO, buf, bytes);
        else
            break;
```

Or this (preferred by some):

```suggestion
        if (bytes > 0) {
            write(STDOUT_FILENO, buf, bytes);
        } else {
            break;
        }
```

enarx-keep/tests/integration_tests.rs
```diff
@@ -17,7 +17,7 @@ const TIMEOUT_SECS: u64 = 5;
 
 /// Returns a handle to a child process through which output (stdout, stderr) can
 /// be accessed.
-fn run_test(bin: &str, status: i32) -> std::process::Child {
+fn run_test(bin: &str, status: i32, input: Option<&str>) -> std::process::Child {
```
npmccallum, line 20:
```suggestion
fn run_test(bin: &str, status: i32, input: impl Into<Option<&str>>) -> std::process::Child {
```

There's a neat Rust trick here. See below...

enarx-keep/tests/integration_tests.rs
```diff
@@ -30,6 +30,17 @@ fn run_test(bin: &str, status: i32) -> std::process::Child {
         .spawn()
         .expect("failed to run test bin");
 
+    if let Some(input) = input {
+        child
+            .stdin
+            .as_mut()
+            .unwrap()
+            .write_all(input.as_ref())
```
npmccallum, line 38:
```suggestion
            .write_all(input.into().as_ref())
```

enarx-keep/tests/integration_tests.rs
```diff
@@ -111,5 +122,15 @@ fn write_stderr() {
 
 #[test]
 fn write_emsgsize() {
-    run_test("write_emsgsize", 0);
+    run_test("write_emsgsize", 0, None);
+}
+
+#[test]
+fn read() {
+    let input = "hello world\n";
```
npmccallum, line 130:
```suggestion
    const INPUT: &str = "hello world\n";
```

enarx-keep/tests/integration_tests.rs
```diff
@@ -111,5 +122,15 @@ fn write_stderr() {
...
+    let child = run_test("read", 0, Some(input));
+    let stdout = child.stdout.unwrap();
+
+    let buf: [u8; 12] = read_item(stdout).unwrap();
```
npmccallum, line 134:
```suggestion
    let buf: [u8; INPUT.len()] = read_item(stdout).unwrap();
```

I wish this was possible... :(

npmccallum (changes requested):

enarx-keep/tests/bin/read.c
lkatalin: It was the size of the current string. So, a reason, but not a good reason. : )
Didn't know that about initializing, either.

enarx-keep/tests/bin/read.c
lkatalin: Wow - I didn't realize you could leave the middle test expression out.

enarx-keep/tests/integration_tests.rs
lkatalin: I think it worked.

enarx-keep/tests/integration_tests.rs
lkatalin: I had to use `to_string()` so that it didn't resolve to `T`.

enarx-keep/tests/integration_tests.rs
```diff
@@ -111,5 +122,15 @@ fn write_stderr() {
...
+    run_test("write_emsgsize", 0, None);
+}
+
+#[test]
+fn read() {
+    const INPUT: &str = "hello world\n";
+    let child = run_test("read", 0, INPUT);
+    let stdout = child.stdout.unwrap();
+
+    let buf: [u8; INPUT.len()] = read_item(stdout).unwrap();
```
npmccallum, line 134: There is still one problem with this. Although we know that the string above is 12 bytes, `INPUT.len()` here returns 12 **characters**, which might be more than 12 bytes (but isn't in actuality). We probably actually need something like `INPUT.bytes().len()` but I'm not sure if that will work.

npmccallum (changes requested):

npmccallum (approved):

PR #916, opened 2020-09-02
Update the description of Enarx goals in the README

### LINKED ISSUE #172: Update README to better describe the properties Enarx offers
The README currently mentions this:
> Create a way to create and run "private, fungible, serverless" applications using Trusted Execution Environments (TEEs). In other words, to provide a platform abstraction for TEEs.
> 
> They should:
> 
>  * be confidential from the host and other workloads;
>  * be able to run on
>    * multiple silicon architectures; 
>    * multiple server types;
>    * public cloud or private cloud;
>  * employ serverless deployment and execution mechanisms.

@steveeJ was suggesting the following:
> Rather than the buzzword I'd describe the properties this implies in your understanding.

This issue is to track and discuss this.

axelsimon:
A quick update of the README, closes #172.

Signed-off-by: axelsimon <github@axelsimon.net>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,37 @@
 # Enarx
 
 ## Introduction
-Enarx is an application deployment system enabling applications to run within Trusted Execution Environments (TEEs) without rewriting for particular platforms or SDKs. It handles attestation and delivery into a run-time “Keep” based on WebAssembly, offering developers a wide range of language choices for implementation. Enarx is CPU-architecture independent, enabling the same application code to be deployed across multiple targets, abstracting issues such as cross-compilation and differing attestation mechanisms between hardware vendors. Work is currently underway on AMD SEV and Intel SGX.
+Enarx is an application deployment system enabling applications
+to run within Trusted Execution Environments (TEEs) without rewriting for
+particular platforms or SDKs. It handles attestation and delivery into a
+run-time “Keep” based on WebAssembly, offering developers a wide range of
+language choices for implementation. Enarx is CPU-architecture independent,
+enabling the same application code to be deployed across multiple targets,
+abstracting issues such as cross-compilation and differing attestation
+mechanisms between hardware vendors. Work is currently underway on AMD SEV and
+Intel SGX.
 
-We've known for a long time that we need encryption for data at rest and in transit: Enarx helps you do encryption for data as well as algorithms in use.
+We've known for a long time that we need encryption for data at rest and in
+transit: Enarx helps you do encryption for data as well as algorithms in use.
 
 [Read more](https://github.com/enarx/enarx/wiki/Enarx-Introduction)
 
 ## Project aim
-Create a way to create and run "private, fungible, serverless" applications using Trusted Execution Environments (TEEs). In other words, to provide a platform abstraction for TEEs.
+The aim of the Enarx project is to create a way to allow easy deployment of
+applications using Trusted Execution Environments (TEEs) across multiple
+platforms, with strong security properties.
 
-They should:
 
- * be confidential from the host and other workloads;
- * be able to run on
-   * multiple silicon architectures;
-   * multiple server types;
-   * public cloud or private cloud;
- * employ serverless deployment and execution mechanisms.
+Deployed workloads should be:
+* Confidentiality protected from the host and from other workloads
+* Integrity protected from the host and from other workloads
+* Cryptographically attested at every instance deployment
+* Platform-agnostic at deployment time
+  * able to run on multiple silicon architectures without recompilation
+  * able to run on any cloud, Edge or other environment which provides
+     appropriate hardware 
+* Language-agnostic at development time
+  * able to be written in multiple languages and compiled to a single package
 
 The mechanism itself should be capable of being FIPS-certified.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,23 +1,36 @@
 # Enarx
 
-## Introduction
-Enarx is an application deployment system enabling applications to run within Trusted Execution Environments (TEEs) without rewriting for particular platforms or SDKs. It handles attestation and delivery into a run-time “Keep” based on WebAssembly, offering developers a wide range of language choices for implementation. Enarx is CPU-architecture independent, enabling the same application code to be deployed across multiple targets, abstracting issues such as cross-compilation and differing attestation mechanisms between hardware vendors. Work is currently underway on AMD SEV and Intel SGX.
-
-We've known for a long time that we need encryption for data at rest and in transit: Enarx helps you do encryption for data as well as algorithms in use.
+## Introduction Enarx is an application deployment system enabling applications
+to run within Trusted Execution Environments (TEEs) without rewriting for
```
connorkuehl, line 4: Was it your intention to have the first part of this paragraph as a header here?

If so, I think the word "Introduction" should be removed. 

If not, I think the words following "Introduction" should be moved into the paragraph section.

README.md
```diff
@@ -1,23 +1,36 @@
...
+particular platforms or SDKs. It handles attestation and delivery into a
```
connorkuehl, line 5: I think something along these lines would add more precision to this paragraph:

"It handles attestation and delivery of WebAssembly workloads into a run-time "Keep", [..]"

README.md
```diff
@@ -1,23 +1,36 @@
...
+run-time “Keep” based on WebAssembly, offering developers a wide range of
+language choices for implementation. Enarx is CPU-architecture independent,
+enabling the same application code to be deployed across multiple targets,
+abstracting issues such as cross-compilation and differing attestation
+mechanisms between hardware vendors. Work is currently underway on AMD SEV and
+Intel SGX.
+
+We've known for a long time that we need encryption for data at rest and in
+transit: Enarx helps you do encryption for data as well as algorithms in use.
```
connorkuehl, line 14: "data ~~as well as~~ and algorithms in use."

README.md
```diff
@@ -1,23 +1,36 @@
...
-   * multiple silicon architectures;
-   * multiple server types;
-   * public cloud or private cloud;
- * employ serverless deployment and execution mechanisms.
+The aim of the Enarx project is to create a way to allow easy deployment of
+applications using Trusted Execution Environments (TEEs) across multiple
+platforms, with strong security properties.
+
+
+Deployed workloads should be:
```
connorkuehl, line 24: ~~Deployed workloads should be~~ -> "Enarx-deployed workloads will have these properties:"

README.md
```diff
@@ -1,23 +1,36 @@
...
+* Confidentiality protected from the host and other workloads
```
connorkuehl, line 25: Suggestion:

"Confidential: nothing (except for the workload itself) has the visibility required to disclose the workload's sensitive information. All unauthorized access, even from the host computer the workload is deployed on, is blocked."

README.md
```diff
@@ -1,23 +1,36 @@
...
+* Integrity protected from the host and other workloads
```
connorkuehl, line 26: Suggestion:

"Integrity-protected: all attempts to manipulate the workload's memory (whether malicious or by accident) are blocked. The workload's data will _never_ change without its blessing."

README.md
```diff
@@ -1,23 +1,36 @@
...
+* Cryptographically attested at every instance deployment
+* Platform-agnostic at deployment time
```
connorkuehl, line 28: I think it's more effective to remove this bullet point and just pull its sub-bullet points out to the top level

README.md
```diff
@@ -1,23 +1,36 @@
...
-
-They should:
-
- * be confidential from the host and other workloads;
- * be able to run on
-   * multiple silicon architectures;
-   * multiple server types;
-   * public cloud or private cloud;
- * employ serverless deployment and execution mechanisms.
+The aim of the Enarx project is to create a way to allow easy deployment of
```
mbestavros, line 19:
```suggestion
Create a way to allow easy deployment of
```

We already specify that this is the project's aim; no need to say it twice.

README.md
mbestavros: A suggestion to your suggestion: "Confidentiality" in your first sentence should become "Confidential".

Otherwise, it reads like "Deployed workloads should be confidentiality", which isn't right.

mbestavros (changes requested):
I think I agree with most of @connorkuehl's suggested changes, and I have a few of my own to add.

This is a good improvement regardless, though,

README.md
mbestavros: A suggestion to your suggestion: "Confidentiality" in your first sentence should become "Confidential".

Otherwise, it reads like "Deployed workloads should be confidentiality", which isn't right.

README.md
haraldh, line 14: I don't like "We" and "you".

```suggestion
Enarx ensures to keep data at rest and in transit encrypted, as well as the algorithms in use.
```

haraldh (changes requested):

README.md
axelsimon: Good catch. Will correct, thanks.

README.md
axelsimon: Thanks, we want to keep the wording simple, so i think we'll keep the current one.

README.md
axelsimon: Don't entirely agree, "Project aim" is the name on the section. I'd argue using fully formed sentences after the title of the section to repeat the same idea is fine, the opposite is a bit too terse.

README.md
axelsimon: I think that's already too specific, we want to keep this both factually correct and readable by various audiences, including some for which "WebAssembly workload" will be less clear than "based on WebAssembly", which is arguably more hand-wavy, but on purpose.

My feeling:
"WebAssembly workloads" --> Wait, do I use that? Am I doing that? Will I need to change things?
"based on WebAssembly" --> Right, I've heard of that technology, let's see how they use it.

README.md
axelsimon: This is good but should probably go in another document, where we can explain this in more detail. We are trying to keep the readme short and simple. @MikeCamel, what do you think?

README.md
axelsimon: I think we've been using this wording (platform-agnostic, deployment time) and should keep it for now.

README.md
axelsimon: Same as previous comment. Good explanation, but probably better suited to a longer intro.

axelsimon (commented):
Thanks for your comments. To be honest, the idea was to do a quick update of that central part of the README, not to rework the whole thing / discuss style: @MikeCamel and i reworked this on a collaborative (g)doc and were hoping to avoid the whole PR process as much as possible :wink:  

There are multiple useful comments, but some of the improvements suggested would probably better suited to a longer intro. I'll work them into the wiki's intro, which i think should be longer than the README (really, the README should be the bare minimum, and take people to the wiki for more).

axelsimon:
New version pushed, i'll re-request review.

mbestavros (approved):

PR #924, opened 2020-09-03
Use env `ENARX_BACKEND` instead of arg `--keep`

npmccallum:
This allows us to pass this configuration parameter across intermediary
commands (such as `cargo test`).

### DIFF
--- a/enarx-keep/src/main.rs
+++ b/enarx-keep/src/main.rs
@@ -36,10 +36,6 @@ struct Info {}
 /// Executes a keep
 #[derive(StructOpt)]
 struct Exec {
-    /// The specific keep backend to use
-    #[structopt(short, long)]
-    keep: Option<String>,
-
     /// The socket to use for preattestation
     #[structopt(short, long)]
     sock: Option<PathBuf>,
@@ -101,9 +97,11 @@ fn info(backends: &[Box<dyn Backend>]) -> Result<()> {
 
 #[allow(unreachable_code)]
 fn exec(backends: &[Box<dyn Backend>], opts: Exec) -> Result<()> {
+    let keep = std::env::var_os("ENARX_BACKEND").map(|x| x.into_string().unwrap());
+
     let backend = backends
         .iter()
-        .filter(|b| opts.keep.is_none() || opts.keep == Some(b.name().into()))
+        .filter(|b| keep.is_none() || keep == Some(b.name().into()))
         .find(|b| b.have());
 
     if let Some(backend) = backend {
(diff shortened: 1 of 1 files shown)


npmccallum:
Tests will pass once #904 is merged.

connorkuehl (approved):

lkatalin (approved):

enarx-keep/src/main.rs
```diff
@@ -120,7 +118,7 @@ fn exec(backends: &[Box<dyn Backend>], opts: Exec) -> Result<()> {
             }
         }
     } else {
-        match opts.keep {
+        match keep {
```
MikeCamel, line 121: Sorry if I'm late to the party - here are my comments having come to this code "fresh".
I don't understand what's being checked here. It looks like if we have a name, then we panic if it _isn't_ "nil". 
1. where does "nil" come from - is this default from an env variable if it's not assigned?
1a. why are we matching against a hard-coded string here, and not a const?
2. I _strongly_ suggest that any time we have any unsafe blocks that they are heavily commented.  I know this isn't new code (I wasn't on the review before for when it came in), but line 127 is basically completely opaque, and it's the core line of the entire file, afaics.

MikeCamel (commented):
Just some lack of clarity on my part on the match statement.  Should be easily explained.

enarx-keep/src/main.rs
npmccallum: > Sorry if I'm late to the party - here are my comments having come to this code "fresh".
> I don't understand what's being checked here. It looks like if we have a name, then we panic if it _isn't_ "nil".

Correct. The name `nil` refers to a pseudo-backend that uses the `execl()` function from `libc` to execute another binary **in place** of the current binary.

Basically, we:
1. search for the named backend
2. if we find the backend, we check to see if it is supported
3. if it is supported, we run the backend
4. if the named backend isn't supported, we have two cases:
  1. the backend is `nil`: then we call `execl()`
  2. otherwise, we alert the user to the fact the backend they specified isn't supported

In practice this means `enarx-keep` can launch in three modes:
1. `nil`: exactly the same as if you had run the binary directly
2. `kvm`: the binary is launched in an unencrypted VM
3. `sgx` or `sev`: the binary is launched inside a TEE

All three modes have *exactly* the same launch sequence.

> 1. where does "nil" come from - is this default from an env variable if it's not assigned?

It is specified by the user. Yes, we need to document this.

>    1a. why are we matching against a hard-coded string here, and not a const?

Because nobody asked for that during review.

> 2. I _strongly_ suggest that any time we have any unsafe blocks that they are heavily commented.  I know this isn't new code (I wasn't on the review before for when it came in), but line 127 is basically completely opaque, and it's the core line of the entire file, afaics.

That would be a heavy handed policy.

An `unsafe` block doesn't mean that the code inside is dangerous. It merely means "I accept that the Rust compiler cannot verify that everything in this block is correct."

In this case, `execl()` is a very well known function. It is unsafe because it is a C function and all FFI functions are unsafe by default (regardless of whether they are dangerous or not). There's nothing really dangerous about this invocation and, if it succeeds the `enarx-keepldr` process will be replaced by the contents of the executable at the path `cstr`.

PR #928, opened 2020-09-08
Explicitly accept ISC licenses for cargo deny

### LINKED ISSUE #909: Investigate acceptability of ISC licence for project.
The ISC licence has come up as a blocker for acceptance of some of our code.  Need to decide whether to add it to our list of allowed licences.

mbestavros:
This was brought up in #909, but the discussion should probably be continued here. The updated config for `deny` will now accept `ISC` explicitly if merged. Please discuss further.

Resolves #909.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -15,6 +15,7 @@ allow = [
     "CC0-1.0",
     "MIT",
     "Zlib",
+    "ISC",
 ]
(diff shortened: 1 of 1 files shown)


npmccallum:
@mbestavros This file is now removed. So I don't think this PR applies anymore.

PR #935, opened 2020-10-12
Repo architecture

MikeCamel:
Resolves https://github.com/enarx/enarx.github.io/issues/36
Unclear if I'm supposed to check in the generated file.


lkatalin (approved):
LGTM; we can have a separate diagram that covers use of the various repos. Also, the shading is a nice addition.

npmccallum (approved):

PR #987, opened 2021-10-05
Rename `enarx-keepldr` to `enarx`

npmccallum:


wgwoods (approved):
Looks good to me - you caught everything I could think of and then some.

haraldh (approved):

PR #1003, opened 2021-10-11
Add Nix development shell support

rvolosatovs:
This adds support for Nix development shells, which means that users with `nix` package manager installed (see https://nixos.org) should be able to just do:
```sh
$ nix-shell
```
(on legacy, stable `nix` installs)

or:
```sh
$ nix develop
```
(on more recent installs with (still experimental) flake support)

And immediately start developing as usual (i.e. `cargo test` and `cargo build` just work).

Note, that LD/CC environment variable thing is quite hacky and fragile and should ideally not be necessary. I will later file an issue upstream regarding this, or perhaps ask around for some input, but for now it seems that `-nostdlib` is (perhaps slightly) broken in Nixpkgs or at least very difficult/tricky to use right. Nevertheless, I believe this should work on most, if not all, systems with `nix`. @haraldh would be nice if you could test this as well.

Removed `rustup`-specific `+nightly`, since `rustup` is not used in this setup and it should not be necessary, especially since we have https://github.com/enarx/enarx/blob/6b54cdcebdec406f26c0a52109fb36aeb3063302/rust-toolchain.toml available.


flake.nix
```diff
@@ -0,0 +1,51 @@
...
+          musl.dev
+        ]) ++ [
+          rust
+        ];
+
+        nativeBuildInputs = with pkgs; [
+          pkg-config
+        ];
+
+        unsetEnv = pkgs.lib.concatMapStringsSep "\n" (var: "unset ${var}") [
```
rvolosatovs, line 34: We might be setting the wrong input fields here, by the way https://nixos.org/manual/nixpkgs/stable/#variables-specifying-dependencies

haraldh: I have no idea

haraldh:
ok, with `git` already installed, this works for me

haraldh (approved):

PR #1007, opened 2021-10-12
Rework CLI

wgwoods:
This PR refactors/improves the CLI in preparation for a first release!

 Visible changes:

- [x] Add `enarx run` for running WebAssembly modules 
- [x] Change `enarx exec` back to executing a provided static-PIE binary
- [x] Hide/remove `enarx exec` from release builds  
- [x] Add `-v` / `--verbose` to increase log verbosity
- [x] Add `--log-filter` for complex log filtering (like RUST_LOG)
- [x] Also accept log filters with `ENARX_LOG` environment variable
- [x] Add `--log-target={stdout,stderr}` for choosing log target
- [x] Add `--backend=` argument equivalent to `ENARX_BACKEND`
- [x] Update README.md

Under-the-hood changes:

- [x] Add `cli` module
- [x] Move existing subcommands to their own `cli` submodules
- [x] Put common CLI structs etc (e.g. logging options, backend selection) in `cli`
- [x] Fix integration tests for the updated CLI
 
Features that I'm deferring to a followup PR:

- Make `--log-filter`/`ENARX_LOG`/`-v` turn on logging inside workldr too

### DIFF
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -10,7 +10,6 @@ pub struct Options {}
 
 impl Options {
     /// Display nicely-formatted info about each backend
-    #[allow(clippy::unnecessary_wraps)]
     pub fn display(self) -> Result<()> {
         use colorful::*;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -78,7 +78,6 @@ struct Options {
     cmd: cli::Command,
 }
 
-#[allow(clippy::unnecessary_wraps)]
 fn main() -> Result<()> {
     let opts = Options::from_args();
     opts.log.init_logger();


wgwoods:
For bikeshedding, here's the new & improved CLI help output:
```
$ enarx help
enarx 0.1.0
`enarx` subcommands and their options/arguments

USAGE:
    enarx [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
    -v, --verbose    Increase log verbosity. Pass multiple times for more log output

OPTIONS:
        --log-filter <log-filter>    Set fancier logging filters [env: ENARX_LOG=]
        --log-target <log-target>    Set log output target ("stderr", "stdout") [default: stderr]

SUBCOMMANDS:
    info    Show details about backend support on this system
    exec    Execute a static-PIE binary inside an Enarx Keep
    run     Run a WebAssembly module inside an Enarx Keep
    help    Prints this message or the help of the given subcommand(s)
```

```
$ enarx help info
enarx-info 0.1.0
Show details about backend support on this system

USAGE:
    enarx info

FLAGS:
```

```
$ enarx help run
enarx-run 0.1.0
Run a WebAssembly module inside an Enarx Keep

USAGE:
    enarx run [OPTIONS] <MODULE> [-- <ARGS>...]

FLAGS:

OPTIONS:
        --backend <backend>    Set which backend to use [env: ENARX_BACKEND=]
    -e, --env <NAME=VAL>...    Set a WASI environment variable

ARGS:
    <MODULE>     Path of the WebAssembly module to run
    <ARGS>...    Arguments to pass to the WebAssembly module
```

```
$ enarx help exec
enarx-exec 0.1.0
Execute a static-PIE binary inside an Enarx Keep

USAGE:
    enarx exec [OPTIONS] <code>

FLAGS:

OPTIONS:
        --backend <backend>    Set which backend to use [env: ENARX_BACKEND=]

ARGS:
    <code>    External static PIE binary to run inside the keep
```

Cargo.toml
```diff
@@ -51,6 +51,9 @@ goblin = "0.4"
 libc = "0.2"
 lset = "0.2"
 vdso = "0.1"
+log = "0.4"
+once_cell = "1.8"
```
haraldh, line 55: I would prefer the `spinning` crate. See https://github.com/wgwoods/enarx/pull/2

src/cli/info.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::backend::builtin_backends;
+use crate::cli::{Result, StructOpt};
+
+/// Show details about backend support on this system
+#[derive(StructOpt, Debug)]
+pub struct Options {}
+
+impl Options {
+    /// Display nicely-formatted info about each backend
+    #[allow(clippy::unnecessary_wraps)]
```
haraldh, line 12: Can this be removed?

Cargo.toml
wgwoods: Hm. not sure which of these to take, though - using pure static vars seems better, but.. if we have external/tenant-supplied shims then the set of *shims* is not static. Does that mean `BACKENDS` would also be mutable? If I built `enarx` without `backend-sgx` but handed it `shim-sgx`, could we build a keep with it? If so, that variable should probably be called `BUILTIN_BACKENDS` or something, right?

On the other hand, if `BACKENDS` defines which backends it's _possible_ for this `enarx` build to load shims for, then I'm cool with it as-is.

src/cli/log.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Represents targets for debug logging.
+/// This will probably grow over time.
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+enum LogTarget {
+    Stdout,
+    Stderr,
+    // FUTURE: file path, syslog/journal, ...
```
rvolosatovs, line 37: Just a question for my own understanding, how is `FUTURE` different from `TODO`?

src/main.rs
```diff
@@ -57,124 +49,64 @@
...
-        for datum in &data {
-            if let Some(mesg) = datum.mesg.as_ref() {
-                println!("\n{}\n", mesg);
-            }
+        cli::Command::Run(run) => {
+            let modfile = File::open(run.module)?;
+            let open_fd = modfile.as_raw_fd();
+            // TODO: we should pass the fd to workldr, but we don't actually
+            // have a way to pass arguments to workldr yet, so for now..
+            assert!(open_fd == 3, "module got unexpected fd {}", open_fd);
```
rvolosatovs, line 101: Shouldn't this result in an error, rather than a panic? (the outer function returns a `Result`)

src/cli/log.rs
wgwoods: Heh, yeah, it is kind of a fuzzy difference. Here's how I use `FUTURE`, `TODO`, and `FIXME`:

I use `FUTURE` for possible future features - ideas that might be good or interesting, but wouldn't be trivial to add, aren't required by any specification, aren't on the roadmap, and aren't a blocker for release.

I use `TODO` when I'm pointing out known-missing/incomplete parts of _working_ code - it's not _broken_, but there's a feature/enhancement we know we'll need later. It's for things that are on the roadmap, or part of some spec, or for tech debt we know we'll need to repay later. Still not a release blocker, but shouldn't be thrown out lightly.

And then `FIXME` is for when I'm pointing out something that I know is just plain wrong, or broken, or silly. Those _should_ block release, except in special circumstances where we have a clear explanation of why we're leaving known-broken code in place, and there's a plan for how and when we're going to fix it.

...we should maybe put something in `STYLE.md` about this.

PR #1010, opened 2021-10-14
Clarify `super::` vs `crate::` style import usage

rvolosatovs:
Clarified when to use `super::` vs `crate::` style imports.


npmccallum:
@wgwoods Can you please provide your thoughts?

haraldh:
> Use `super` when importing from direct ancestors up to 3 (?) levels up inclusive

2 level `super::super::` is my personal super-limit :smile:

haraldh:
> Okay, so it looks like the rules should say:
> 
>     1. Prefer `super::` to `crate::`.

Although not totally convinced of this point, I can arrange with it, as long as it does not mandate this.

PR #1011, opened 2021-10-15
Turn on logging & document hardcoded keep config

wgwoods:
We don't yet have a well-defined way to pass runtime configuration from the frontend/CLI into the keep. For v0.1.0 we've decided to just hardcode a default configuration, as follows:

* the .wasm module is open on fd3 and gets no arguments or env vars
* stdin, stdout, and stderr are enabled and should go to fd 0,1,2
* logging should be turned on at "debug" level

So these two patches turn on logging (by adding the RUST_LOG environment variable to the keep environment in both shims), and also document how/where we're applying our hardcoded default configuration.

All the code with a `KEEP-CONFIG HACK` comment above it is intended to be replaced/removed immediately after our first release. This is all temporary, but at least now it's explicit - and we know what we need to fix.

### DIFF
--- a/internal/shim-sev/src/payload.rs
+++ b/internal/shim-sev/src/payload.rs
@@ -123,6 +123,18 @@ fn crt0setup(
     builder.push("/init").unwrap();
     let mut builder = builder.done().unwrap();
     builder.push("LANG=C").unwrap();
+    // FIXME - v0.1.0 KEEP-CONFIG HACK
+    // We don't yet have a well-defined way to pass runtime configuration from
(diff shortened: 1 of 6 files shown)


haraldh:
We have to disable the 'write_stderr' test, because stderr is clobbered with all the debug output.

wgwoods:
Rebased to pick up the bugfix in commit 69c746e8e62a39b69a48cad4b30f43594d9feea7 and dropped the `DEBUG=true` from `shim-sgx`

Issue #1020, opened 2021-10-18
Tool to get SEV-SNP certificate chain


haraldh:
Get vcek key with:
- https://kdsintf.amd.com/vcek/v1/Milan/8ba826b2dd6ab65e401e0c4d4128ef4b434ed0ccb213f66c5f577b518730ef5892f78a78be259976973125a3b9b3d19f286c912cf5776fdfcee5260fa4576c4b?blSPL=00&teeSPL=00&snpSPL=03&ucodeSPL=29

blSPL: `tcb_version.bootloader`
teeSPL: `tcb_version.tee`
ucodeSPL: `tcb_version.microcode`
snpSPL: `tcb_version.snp`

- find out what the hex-string is

- convert to `pem` programmatically like openssl with `openssl x509 -inform der -in vcek.crt -out vcek.pem`

Get ARK and ASK (`Milan.pem`) with:
- https://kdsintf.amd.com/vcek/v1/Milan/cert_chain 

Output to stdout: `vcek.pem` and `Milan.pem` concatenated.

All without storing anything to disk.

How to verify:

```console
$ snp-get-certs > test.pem
$ firefox ./test.pem
```

Select "open with viewer (standard)" in firefox... should display all certs nicely.

PR #1026, opened 2021-10-19
fix: remove sev crate dependency

haraldh:
And put the needed parts in `backend::sev::snp`.

Co-authored-by: Roman Volosatovs <roman@profian.com>

### DIFF
--- a/src/backend/sev/builder.rs
+++ b/src/backend/sev/builder.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::cpuid_page::CpuidPage;
+use super::snp::firmware::Firmware;
+use super::snp::launch::*;
+
 use super::SnpKeepPersonality;
 use crate::backend::kvm::builder::kvm_try_from_builder;
 use crate::backend::kvm::mem::Region;
@@ -14,8 +17,6 @@ use kvm_ioctls::{Kvm, VmFd};
 use mmarinus::{perms, Map};
 use primordial::Page;
 use sallyport::elf::pf::snp::{CPUID, SECRETS};
-use sev::firmware::Firmware;
-use sev::launch::snp::*;
 use x86_64::VirtAddr;
 
 pub struct Builder {
@@ -33,11 +34,12 @@ impl TryFrom<super::super::kvm::config::Config> for Builder {
         let vm_fd = kvm_fd.create_vm()?;
 
         let sev = Firmware::open()?;
+
         let launcher = Launcher::new(vm_fd, sev)?;
 
-        let start = SnpStart {
-            policy: SnpPolicy {
-                flags: SnpPolicyFlags::SMT,
+        let start = Start {
+            policy: Policy {
+                flags: PolicyFlags::SMT,
                 ..Default::default()
             },
             gosvw: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16],
@@ -91,11 +93,11 @@ impl super::super::Mapper for Builder {
                 guest_cpuid_page.write(cpuid_page);
             }
 
-            let update = SnpUpdate::new(
+            let update = Update::new(
                 to as u64 >> 12,
                 &pages,
                 false,
-                SnpPageType::Cpuid,
+                PageType::Cpuid,
                 (dp, dp, dp),
             );
 
@@ -108,11 +110,11 @@ impl super::super::Mapper for Builder {
         } else if with & SECRETS != 0 {
             assert_eq!(pages.len(), Page::SIZE);
 
-            let update = SnpUpdate::new(
+            let update = Update::new(
                 to as u64 >> 12,
                 &pages,
                 false,
-                SnpPageType::Secrets,
+                PageType::Secrets,
                 (dp, dp, dp),
             );
 
@@ -120,11 +122,11 @@ impl super::super::Mapper for Builder {
                 .update_data(update)
                 .context("SNP Launcher update_data")?;
         } else {
-            let update = SnpUpdate::new(
+            let update = Update::new(
                 to as u64 >> 12,
                 &pages,
                 false,
-                SnpPageType::Normal,
+                PageType::Normal,
                 (dp, dp, dp),
             );
 
@@ -149,7 +151,7 @@ impl TryFrom<Builder> for Arc<dyn super::super::Keep> {
             builder.launcher.as_mut(),
         )?;
 
-        let finish = SnpFinish::new(None, None, [0u8; 32]);
+        let finish = Finish::new(None, None, [0u8; 32]);
 
         let (vm_fd, sev_fd) = builder.launcher.finish(finish)?;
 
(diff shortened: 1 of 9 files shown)


src/backend/sev/snp/launch/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+}
+
+impl<U: AsRawFd, V: AsRawFd> Launcher<Started, U, V> {
+    /// Encrypt guest SNP data.
+    pub fn update_data(&mut self, update: Update) -> Result<()> {
+        let launch_update_data = LaunchUpdate::from(update);
+        let mut cmd = Command::from(&mut self.sev, &launch_update_data);
+
+        KvmEncRegion::new(update.uaddr).register(&mut self.vm_fd)?;
```
rvolosatovs, line 93: I thought we wanted to depend on https://github.com/rust-vmm/kvm-ioctls/pull/178 for this functionality, such that we would not need to implement/maintain this ourselves, is that not the case anymore? If still yes, should we add a TODO?

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+            write!(f, "{:02X}", b)?;
+        }
+
+        Ok(())
+    }
+}
+
+/// Information regarding the SEV-SNP platform's TCB version.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SnpTcbStatus {
```
rvolosatovs, line 299: Is `Snp` prefix still necessary here (and others), given that we're within `...snp::firmware`?

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+
+impl<T: Debug + Display> error::Error for Indeterminate<T> {}
+
+/// Error conditions returned by the SEV platform or by layers above it
+/// (i.e., the Linux kernel).
+///
+/// These error conditions are documented in the AMD SEV API spec, but
+/// their documentation has been copied here for completeness.
+#[derive(Debug)]
+pub enum Error {
```
rvolosatovs, line 52: Should we move this and related into a separate module (e.g. `error.rs`)? Also I wonder if all of these are still relevant, given that some functionality is removed :thinking:. But probably good to have everything for completeness.

src/backend/sev/snp/launch/mod.rs
haraldh: yes, we should add a todo issue

src/backend/sev/snp/firmware/mod.rs
haraldh: let me fix this

src/backend/sev/builder.rs
```diff
@@ -32,12 +33,16 @@ impl TryFrom<super::super::kvm::config::Config> for Builder {
         let kvm_fd = Kvm::new()?;
         let vm_fd = kvm_fd.create_vm()?;
 
-        let sev = Firmware::open()?;
+        let mut sev = Firmware::open()?;
+
+        // FIXME: use platform_status to get tcb_version and VCEK?
+        let _platform_status = sev.platform_status()?;
```
rvolosatovs, line 39: If we discard the result value of this, why do we need to call it in the first place?
If there's something (SEV) implementation-specific going on, maybe worth adding a comment about it?
Otherwise, should we file an issue about the
> use platform_status to get tcb_version and VCEK

?

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Lowest current patch level of all the cores.
+    pub microcode: u8,
+}
+
+/// Query the SEV-SNP platform status.
+///
+/// (Chapter 8.3; Table 38)
+#[derive(Default)]
+#[repr(C)]
+pub struct SnpPlatformStatus {
```
rvolosatovs, line 149: Do we really need two status structs here?
If yes, should we keep `repr(C)` structs in a separate submodule? (could (but does not need to) be even defined in this file, in fact, since we only need a few)
Also, do we need to export these across crate boundary?

src/backend/sev/snp/firmware/linux.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use std::fs::{File, OpenOptions};
+use std::marker::PhantomData;
+use std::os::unix::io::{AsRawFd, RawFd};
+
+// These enum ordinal values are defined in the Linux kernel
+// source code: include/uapi/linux/psp-sev.h
+impl_const_id! {
+    pub Id => u32;
+    // […]
```
rvolosatovs, line 18: We will also need `GetId = 8` here for #1020 , but I can move that myself from SEV crate in the PR

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub platform_version: TcbVersion,
+
+    /// Reported TCB version.
+    pub reported_version: TcbVersion,
+}
+
+/// Information regarding the SEV-SNP platform's current status.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Status {
+    /// The build number.
```
rvolosatovs, line 88: Is this comment still valid given the type of the field?

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub build: Build,
+
+    /// The platform's current state.
+    pub state: State,
+
+    /// IsRmpInitiailzied
+    pub is_rmp_init: bool,
+
+    /// MaskChipId
+    pub mask_chip_id: bool,
```
rvolosatovs, line 98: nit; would be nice to turn these 2 comments into human-readable sentences.

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// The platform state.
+    pub state: u8,
+
+    /// IsRmpInitiailzied
+    pub is_rmp_init: u8,
+
+    /// The platform build ID.
+    pub build_id: u32,
+
+    /// MaskChipId
```
rvolosatovs, line 162: IDEM, less relevant for this I guess

src/backend/sev/builder.rs
haraldh: Update the comment to:
```rust
        // FIXME: use `platform_status()` to prevent warnings about unused code
        // because we want to use  `platform_status()` later in a sub command.
```

src/backend/sev/snp/firmware/mod.rs
rvolosatovs: `build` is of type `Build` defined here https://github.com/enarx/enarx/pull/1026/files#diff-7b889b31bec995c5f7126623e4b70f668d334d68188035dc1274a807a45ff949R107-R116, but comment states it's a *number*

rvolosatovs (approved):

src/backend/sev/snp/firmware/mod.rs
haraldh: changed to 
```rust
    /// The build information.
```

npmccallum (approved):

rvolosatovs (approved):

PR #1039, opened 2021-10-21
fix(shim-sev): reduce the code stack size

haraldh:
Use the same size as the SGX backend for the code stack size.


haraldh:
interesting test failure

PR #1040, opened 2021-10-21
Add `enarx sev vcek` subcommand

### LINKED ISSUE #1020: Tool to get SEV-SNP certificate chain

rvolosatovs:
Closes #1020 

```
$ enarx sev vcek -h
enarx-sev-vcek 0.1.0
Download VCEK certificates for SEV platform and print to stdout in PEM format

USAGE:
    enarx sev vcek

FLAGS:
```


src/backend/sev/snp/firmware/mod.rs
```diff
@@ -62,6 +62,12 @@ impl From<Identifier> for Vec<u8> {
     }
 }
 
+impl From<Vec<u8>> for Identifier {
+    fn from(vec: Vec<u8>) -> Self {
+        Identifier(vec)
```
rvolosatovs, line 67: This does no validation that `vec` represents an actual valid identifier, but I am not sure we can actually assert that in any meaningful way.
We could check that it's at most 64 character long, since that seems like the maximum length of it according to the spec, but that seems to be about all we could do here and not sure if that's worth it, especially since the functionality is only used for testing.

Here's what the SEV spec has to say about it (SNP spec references SEV spec):
image

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -134,7 +150,7 @@ pub struct TcbVersion {
     /// Current PSP OS version.
     /// SVN of PSP operating system.
     pub tee: u8,
-    _reserved: [u8; 4],
+    pub _reserved: [u8; 4],
```
rvolosatovs, line 153: We could avoid needing to do this by introducing a builder/constructor of some sort, but not sure if that's really necessary, since this struct is meant to be filled by C

src/backend/sev/snp/firmware/mod.rs
npmccallum: The ID can be any length. We can't do any validation of it. The contents are opaque.

PR #1072, opened 2021-10-29
Fix a clippy warning

npmccallum:


wgwoods (approved):
I agree with clippy, this is a good change.

haraldh (approved):

PR #1081, opened 2021-11-01
SGX improvements

npmccallum:


haraldh:
stderr is normally not line buffered for obvious reasons... you also don't have to call flush, e.g.

haraldh:
Although the `debug!` and `debugln!` could call some `flush()` like routine here.

internal/shim-sgx/src/handler/mod.rs
```diff
@@ -106,7 +132,7 @@ impl<'a> Handler<'a> {
                     OP_SYSCALL => h.handle_syscall(),
                     OP_CPUID => h.handle_cpuid(),
                     r => {
-                        debugln!(h, "unsupported opcode: {:?}", r);
+                        debugln!(h, "unsupported opcode: 0x{:04x}", r);
```
haraldh, line 135:
```suggestion
                        debugln!(h, "unsupported opcode: {:#04x}", r);
```

haraldh (changes requested):

npmccallum:
I'm rejecting this PR.

PR #1091, opened 2021-11-09
fix(wasmldr): make zerooneone work

haraldh:
Disable `config.dynamic_memory_reserved_for_growth`.

Different values for `config.dynamic_memory_reserved_for_growth` produce
different results:

0..=260KiB: works

261..=268KiB: fails with ud2

269..=272KiB: fails with:
```
panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: cannot recursively acquire mutex', library/std/src/sys/wasi/../unsupported/mutex.rs:23:9
panicked after panic::always_abort(), aborting.
```

273..unknown: fails with ud2

1MiB: fails with ud2

2MiB: works

etc..

Requires/Includes: https://github.com/enarx/enarx/pull/1089

### DIFF
--- a/internal/shim-sev/src/lib.rs
+++ b/internal/shim-sev/src/lib.rs
@@ -9,7 +9,7 @@
 #![deny(clippy::all)]
 #![deny(clippy::integer_arithmetic)]
 #![deny(missing_docs)]
-#![feature(asm, naked_functions)]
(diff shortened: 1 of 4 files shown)


npmccallum (approved):
Please file this against b0.1.z.

PR #1102, opened 2021-11-19
fix: use `tempfile` crate

haraldh:
instead of unmaintained `tempdir` crate.

Also:
- cargo update for more recent packages to counter `cargo deny` failures
- cargo deny for the whole workspace


wgwoods (approved):
LGTM!

PR #1172, opened 2021-12-21
fix: sallyport dependency

haraldh:
commit 6ea4aefabb590ea37da32b4648ffee7d3d3e09ae specified a sallyport
version on a distinct branch without the `v2` subcrate, because
`version` was not used and I thought cargo was confused.

With `version` specified cargo does the right thing and selects the
sallyport version from the main directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,7 +182,7 @@ dependencies = [
  "protobuf",
  "protobuf-codegen-pure",
  "reqwest",
- "sallyport 0.1.0 (git+https://github.com/enarx/sallyport?rev=167a7b5a626e88f42083d2644ff5423e8e66b17a)",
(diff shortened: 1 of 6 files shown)


rvolosatovs (approved):
LGTM, just wondering can't we drop the `rev`?

haraldh:
> LGTM, just wondering can't we drop the `rev`?

Well, I want to to force sync all deps on the same version. It's basically a reminder for me.

rvolosatovs:
> > LGTM, just wondering can't we drop the `rev`?
> 
> Well, I want to to force sync all deps on the same version. It's basically a reminder for me.

Got it, thanks for explanation!

npmccallum (approved):

PR #1238, opened 2022-01-17
ci: Add wasm integration tests

haraldh:
Similar to the `integration/simple` tests, use a cargo runner script, to
simply add integration tests.

### DIFF
--- a/integration/wasm/.cargo/config
+++ b/integration/wasm/.cargo/config
@@ -1,0 +1,5 @@
+[build]
+target = "wasm32-wasi"
+
+[target.wasm32-wasi]
+runner = "../../helper/test-enarx-wasm.sh"
--- a/integration/wasm/src/memspike.rs
+++ b/integration/wasm/src/memspike.rs
@@ -8,6 +8,6 @@ use std::collections::TryReserveError;
 
 fn main() -> Result<(), TryReserveError> {
     let mut alloc: Vec<u8> = Vec::new();
-    let _ = alloc.try_reserve(40_000_000)?;
+    let _ = alloc.try_reserve(16_000_000)?;
     Ok(())
 }
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use process_control::{ChildExt, Output, Timeout};
-use std::io::Write;
+use std::io::{stderr, Write};
 use std::path::Path;
 use std::process::{Command, Stdio};
 use std::time::Duration;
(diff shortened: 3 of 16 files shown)


integration/wasm/src/memspike.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! `memspike`'s primary intention is to trigger memory ballooning in
+//! VM-based keeps. This will help test the ballooning itself as well
+//! as memory pinning for SEV.
+
+fn main() {
+    let _alloc: Vec<u8> = Vec::with_capacity(16_000_000);
```
bstrie, line 8: When we run these tests are they being compiled with optimizations? Because if so, then I suspect this line is being entirely optimized away and thus can't cause a memory spike. We'd want to use `std::hint::black_box` to force the optimizer not to remove it (although black_box is unstable, I think it would be trivial to stabilize).

tests/common/mod.rs
```diff
@@ -80,10 +80,10 @@ pub fn keepldr_exec<'a>(bin: &str, input: impl Into<Option<&'a [u8]>>) -> Output
 
 /// Returns a handle to a child process through which output (stdout, stderr) can
 /// be accessed.
-pub fn keepldr_exec_crate<'a>(
+pub fn keepldr_exec_crate(
     crate_name: &str,
     bin: &str,
-    input: impl Into<Option<&'a [u8]>>,
+    input: impl Into<Option<Vec<u8>>>,
```
bstrie, line 86: Heh, I see you hit a problem where a spawned thread wanted a 'static lifetime so you had to do all this rigamarole to convert it into an owned value. I'll ask about accepting the `thread::scoped` RFC which would have prevented the need for this, but in the meantime this is fine.

tests/common/mod.rs
```diff
@@ -118,6 +119,15 @@ pub fn keepldr_exec_crate<'a>(
         .unwrap_or_else(|e| panic!("failed to run `{}`: {:#?}", bin, e))
         .unwrap_or_else(|| panic!("process `{}` timed out", bin));
 
+    if let Some(input_thread) = input_thread {
+        if let Err(_) = input_thread.join() {
+            unsafe {
+                eprintln!("{}", String::from_utf8_unchecked(output.stderr));
```
bstrie, line 125: I'm curious about the use of `from_utf8_unchecked` here as opposed to just `from_utf8`.

integration/wasm/.cargo/config
```diff
@@ -0,0 +1,5 @@
+[build]
+target = "wasm32-wasi"
+
+[target.wasm32-wasi]
+runner = "../../helper/test-enarx-wasm.sh"  # Searches `PATH` for `foo`.
```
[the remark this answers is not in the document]
haraldh: yes, here: https://doc.rust-lang.org/cargo/reference/config.html#config-relative-paths)
will remove

integration/wasm/src/memspike.rs
haraldh: Yeah, it's not running optimized... `black_box` would be cool to be stabilized.
Will fix the test with `try_reserve()`.

tests/common/mod.rs
haraldh: yes :+1: well spotted

tests/common/mod.rs
haraldh: oops.. debug leftover. Thanks!

Should be:
```rust
let _unused = stderr().write_all(&output.stderr);
```

haraldh:
@bstrie please rereview and do `Resolve conversation` where it is fixed. Thank you!

bstrie (approved):

npmccallum (approved):

PR #1247, opened 2022-01-18
sev: Migrate from KvmEncRegion to kvm_enc_region

jarkkojs:
Use kvm_enc_region provided by the standard library in the place of ad-hoc
KvmEncRegion.

Signed-off-by: Jarkko Sakkinen <jarkko.sakkinen@iki.fi>

PR #1250, opened 2022-01-20
Enarx updates

jarkkojs:
The first one migrates to kvm_enc_region and other one fixes a compilation issues in Arch Linux.

### DIFF
--- a/src/backend/sev/mod.rs
+++ b/src/backend/sev/mod.rs
@@ -14,8 +14,9 @@ use data::{
     dev_kvm, dev_sev, dev_sev_readable, dev_sev_writable, has_reasonable_memlock_rlimit,
     kvm_version, sev_enabled_in_kernel, CPUIDS,
 };
+
+use kvm_bindings::bindings::kvm_enc_region;
 use kvm_ioctls::VmFd;
-use snp::launch::linux::KvmEncRegion;
 
 mod builder;
 mod cpuid_page;
@@ -29,7 +30,11 @@ struct SnpKeepPersonality {
 
 impl KeepPersonality for SnpKeepPersonality {
     fn map(vm_fd: &mut VmFd, region: &Region) -> std::io::Result<()> {
-        KvmEncRegion::new(region.backing()).register(vm_fd)?;
+        let memory_region = kvm_enc_region {
+            addr: region.backing().as_ptr() as _,
+            size: region.backing().len() as _,
+        };
+        vm_fd.register_enc_memory_region(&memory_region).unwrap();
         Ok(())
     }
 }
--- a/src/backend/sev/snp/launch/mod.rs
+++ b/src/backend/sev/snp/launch/mod.rs
@@ -17,6 +17,8 @@ use std::marker::PhantomData;
 use std::os::unix::io::AsRawFd;
 
 use bitflags::bitflags;
+use kvm_bindings::kvm_enc_region;
+use kvm_ioctls::VmFd;
 
 /// Launcher type-state that indicates a brand new launch.
 pub struct New;
@@ -25,30 +27,30 @@ pub struct New;
 pub struct Started;
 
 /// Facilitates the correct execution of the SEV launch process.
-pub struct Launcher<T, U: AsRawFd, V: AsRawFd> {
-    vm_fd: U,
+pub struct Launcher<T, V: AsRawFd> {
+    vm_fd: VmFd,
     sev: V,
     state: PhantomData<T>,
 }
 
-impl<T, U: AsRawFd, V: AsRawFd> AsRef<U> for Launcher<T, U, V> {
+impl<T, V: AsRawFd> AsRef<VmFd> for Launcher<T, V> {
     /// Give access to the vm fd to create vCPUs or such.
-    fn as_ref(&self) -> &U {
+    fn as_ref(&self) -> &VmFd {
         &self.vm_fd
     }
 }
 
-impl<T, U: AsRawFd, V: AsRawFd> AsMut<U> for Launcher<T, U, V> {
+impl<T, V: AsRawFd> AsMut<VmFd> for Launcher<T, V> {
     /// Give access to the vm fd to create vCPUs or such.
-    fn as_mut(&mut self) -> &mut U {
+    fn as_mut(&mut self) -> &mut VmFd {
         &mut self.vm_fd
     }
 }
 
-impl<U: AsRawFd, V: AsRawFd> Launcher<New, U, V> {
+impl<V: AsRawFd> Launcher<New, V> {
     /// Begin the SEV-SNP launch process by creating a Launcher and issuing the
     /// KVM_SNP_INIT ioctl.
-    pub fn new(vm_fd: U, sev: V) -> Result<Self> {
+    pub fn new(vm_fd: VmFd, sev: V) -> Result<Self> {
         let mut launcher = Launcher {
             vm_fd,
             sev,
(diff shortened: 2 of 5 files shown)


jarkkojs:
Hey, please denote me if kvm_enc_region patch still causes compilation issues. I was able to compile it locally. The only way I can test that the build server does not fail is by a PR.

src/backend/sev/snp/launch/mod.rs
```diff
@@ -18,6 +18,11 @@ use std::os::unix::io::AsRawFd;
 
 use bitflags::bitflags;
 
+use kvm_bindings::bindings::kvm_enc_region;
+use kvm_ioctls::Kvm;
+
+use libc;
```
rvolosatovs, line 24: https://github.com/enarx/enarx/pull/1247#discussion_r788699935

rvolosatovs (changes requested):
How about we reopen https://github.com/enarx/enarx/pull/1247 with just the SEV-specific changes (i.e. 2nd commit) and open a separate one for the build change (1st commit)?
Since you opened both PRs from `main` branch, you'd have to close this PR and push the build update to a different (new) branch
You can reopen the original PR by clicking "reopen" and then force-pushing the branch. If you have force-pushed the branch already before, then you will have to force-push it back to the latest-known hash ( `1c2ea7b` ), then reopen and force-push again to the most recent changes. This way we can keep the conversation and review comments, which are still not addressed and now scattered around between the two PRs.

src/backend/sev/snp/launch/mod.rs
jarkkojs: I've now fixed this!

src/backend/sev/mod.rs
```diff
@@ -14,8 +14,10 @@ use data::{
     dev_kvm, dev_sev, dev_sev_readable, dev_sev_writable, has_reasonable_memlock_rlimit,
     kvm_version, sev_enabled_in_kernel, CPUIDS,
 };
+
+use kvm_bindings::bindings::kvm_enc_region;
+
 use kvm_ioctls::VmFd;
```
rvolosatovs, line 20: Please follow the guidelines https://github.com/enarx/enarx/blob/main/STYLE.md#use-declarations

src/backend/sev/snp/launch/mod.rs
```diff
@@ -18,37 +18,41 @@ use std::os::unix::io::AsRawFd;
 
 use bitflags::bitflags;
 
+use kvm_bindings::kvm_enc_region;
+
+use kvm_ioctls::VmFd;
+
```
rvolosatovs, line 24: IDEM

jarkkojs:
> How about we reopen #1247 with just the SEV-specific changes (i.e. 2nd commit) and open a separate one for the build change (1st commit)? Since you opened both PRs from `main` branch, you'd have to close this PR and push the build update to a different (new) branch You can reopen the original PR by clicking "reopen" and then force-pushing the branch. If you have force-pushed the branch already before, then you will have to force-push it back to the latest-known hash ( `1c2ea7b` ), then reopen and force-push again to the most recent changes. This way we can keep the conversation and review comments, which are still not addressed and now scattered around between the two PRs.

Why that would make sense? They are separate commit with different closes tags.

src/backend/sev/mod.rs
jarkkojs: I'm not sure how that does not follow the guide line. It's in its own group.

rvolosatovs:
> > How about we reopen #1247 with just the SEV-specific changes (i.e. 2nd commit) and open a separate one for the build change (1st commit)? Since you opened both PRs from `main` branch, you'd have to close this PR and push the build update to a different (new) branch You can reopen the original PR by clicking "reopen" and then force-pushing the branch. If you have force-pushed the branch already before, then you will have to force-push it back to the latest-known hash ( `1c2ea7b` ), then reopen and force-push again to the most recent changes. This way we can keep the conversation and review comments, which are still not addressed and now scattered around between the two PRs.
> 
> Why that would make sense? They are separate commits with different closes tags. I think it'd be just extra bureaucracy without any point.

That was a response to the original state of the PR when it was identical to #1247, but with the unadressed review comments hidden (since it's a separate PR). It does not matter much at this point indeed, since the contents of the PR have since changed and most of the comments are addressed.

jarkkojs:
> > > How about we reopen #1247 with just the SEV-specific changes (i.e. 2nd commit) and open a separate one for the build change (1st commit)? Since you opened both PRs from `main` branch, you'd have to close this PR and push the build update to a different (new) branch You can reopen the original PR by clicking "reopen" and then force-pushing the branch. If you have force-pushed the branch already before, then you will have to force-push it back to the latest-known hash ( `1c2ea7b` ), then reopen and force-push again to the most recent changes. This way we can keep the conversation and review comments, which are still not addressed and now scattered around between the two PRs.
> > 
> > 
> > Why that would make sense? They are separate commits with different closes tags. I think it'd be just extra bureaucracy without any point.
> 
> That was a response to the original state of the PR when it was identical to #1247, but with the unadressed review comments hidden (since it's a separate PR). It does not matter much at this point indeed, since the contents of the PR have since changed and most of the comments are addressed.

OK got it. Yeah, I got the point about issues and PRs but I think it'd be better idea to go forward as the changes IMHO are in good shape. You could even argue that the first change is logically a legit part of the overall patch set given that it enables compilation, which was broken before that.

src/backend/sev/mod.rs
jarkkojs: So: can you just say how this is wrong and how it should be and then perhaps we can refine the documentation to make it less likely to make the same mistake?

src/backend/sev/snp/launch/mod.rs
```diff
@@ -18,37 +18,41 @@ use std::os::unix::io::AsRawFd;
 
 use bitflags::bitflags;
 
+use kvm_bindings::kvm_enc_region;
+
+use kvm_ioctls::VmFd;
```
haraldh, line 23:
```suggestion
use bitflags::bitflags;
use kvm_bindings::kvm_enc_region;
use kvm_ioctls::VmFd;
```

src/backend/sev/mod.rs
haraldh, line 20:
```suggestion

use kvm_bindings::bindings::kvm_enc_region;
use kvm_ioctls::VmFd;
```

haraldh (changes requested):

src/backend/sev/mod.rs
jarkkojs: OK, will do, thanks to everyone for patience :-)

src/backend/sev/snp/launch/mod.rs
jarkkojs: OK so it should be good now :-)

src/backend/sev/mod.rs
```diff
@@ -14,8 +14,10 @@ use data::{
     dev_kvm, dev_sev, dev_sev_readable, dev_sev_writable, has_reasonable_memlock_rlimit,
     kvm_version, sev_enabled_in_kernel, CPUIDS,
 };
+
```
rvolosatovs, line 17:
```suggestion
```

jarkkojs: I think it should be good now.

src/backend/sev/snp/launch/mod.rs
jarkkojs: BTW, why there is empty line between use std::* and use::bitflags? Anyway the issue is sorted now. I'm not still exactly sure when to add a newline and when not.

src/backend/sev/mod.rs
rvolosatovs: It's still not addressed.

There essentially should be only 2 groups of imports - standard library imports and external dependency imports.

npmccallum (approved):

npmccallum:
@haraldh and @rvolosatovs It looks like we're just waiting for your approvals on this one.

rvolosatovs (approved):

haraldh (approved):

PR #1285, opened 2022-01-27
Show CPU model in `enarx info`

### LINKED ISSUE #1207: Report cpu model in `enarx info`

definitelynobody:
Closes #1207

image

With --json:

```json
{
        "name": "CPU",
        "pass": true,
        "info": "Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz | GenuineIntel",
        "mesg": null
}
```


jarkkojs:
I'd write the short summary "Show CPU model in enarx info" because cpuid is not expected to constantly to fail. Using the "attempt" might give a wrong picture. I'd also write a long description.

E.g.

```
kvm: Show CPU model in enarx info

Invoke CPUID opcode leaf EAX=0, which returns the CPU model ASCII 
string concatenated from EBX, EDX and ECX in this order. E.g. in
the case of Intel CPU:

- EBX: "Genu"
- EDX: "ineI"
- ECX: "ntel"
[taken from Table 3-8 of Intel SDM Volume 2A]

CpuId provides shenanigans to parse all this, and maps the
string as convenient enumerations Vendor::Intel and Vendor::Amd.
```

jarkkojs:
My commit message is lacking one thing tho, which I did not comprehend from the PR: where, when and how this information is showed?

jarkkojs (changes requested):
Please, refine the commit message with the information I provided,
and further with the missing information.

definitelynobody:
Sure thing I'll add more detail to the commit messages

jarkkojs:
@definitelynobody I think the code change itself is acceptable and great quality, just needs a description to matches it.

jarkkojs (approved):

src/backend/probe/x86_64.rs
```diff
@@ -20,8 +26,8 @@ impl Vendor {
         let name = std::str::from_utf8(&name[..]).unwrap();
 
         Ok(match name {
-            "AuthenticAMD" => Self::Amd,
-            "GenuineIntel" => Self::Intel,
+            AUTHENTIC_AMD => Self::Amd,
+            GENUINE_INTEL => Self::Intel,
```
haraldh, line 30: I know it is old code, but could you please also fix the `get()` method into something like this, which does not have `transmute` and `from_utf8().unwrap`

```rust
    pub fn get() -> Result<Self> {
        let res = unsafe { __cpuid_count(0x00000000, 0x00000000) };

        let name = [
            res.ebx.to_le_bytes(),
            res.edx.to_le_bytes(),
            res.ecx.to_le_bytes(),
        ]
        .concat();

        if name.eq(AUTHENTIC_AMD.as_bytes()) {
            Ok(Self::Amd)
        } else if name.eq(GENUINE_INTEL.as_bytes()) {
            Ok(Self::Intel)
        } else {
            Err(Error::new(
                ErrorKind::Other,
                format!("unsupported vendor: {:#?}", name),
            ))
        }
    }
```

src/backend/probe/x86_64.rs
```diff
@@ -40,6 +53,53 @@ pub struct CpuId {
...
+            let res = unsafe { __cpuid(0x00000000) };
+            let name: [u8; 12] = unsafe { transmute([res.ebx, res.edx, res.ecx]) };
+            from_utf8(&name[..]).unwrap().to_string()
+        };
+        let supported_cpu = required_vendor
+            .map(|v| v.id() == manufacturer)
+            .unwrap_or(true);
+        let identifier_fmt = |name: Option<&str>| {
+            format!("{} | {}", name.unwrap_or("[unknown model]"), manufacturer)
+        };
```
haraldh, line 72:
```suggestion
        let manufacturer = {
            let res = unsafe { __cpuid(0x00000000) };
            [
                res.ebx.to_le_bytes(),
                res.edx.to_le_bytes(),
                res.ecx.to_le_bytes(),
            ]
            .concat()
        };
        let supported_cpu = required_vendor
            .map(|v| v.id().as_bytes() == manufacturer)
            .unwrap_or(true);
        let identifier_fmt = |name: Option<&str>| {
            format!(
                "{} | {}",
                name.unwrap_or("[unknown model]"),
                String::from_utf8_lossy(manufacturer.as_slice())
            )
        };
```

src/backend/probe/x86_64.rs
```diff
@@ -40,6 +53,53 @@ pub struct CpuId {
...
+            let string_part: [u8; CHUNK_SIZE] =
+                unsafe { transmute([res.eax, res.ebx, res.ecx, res.edx]) };
+            // Copy all 4 registers into the model name.
+            let start = i * CHUNK_SIZE;
+            let destination = &mut model[start..start + CHUNK_SIZE];
+            destination.clone_from_slice(&string_part);
+        }
+
+        let model_name = from_utf8(&model).unwrap().trim_matches(char::from(0));
+        (supported_cpu, Some(identifier_fmt(Some(model_name))))
```
haraldh, line 99:
```suggestion
        let model_name = String::from_utf8_lossy(&model);
        (supported_cpu, Some(identifier_fmt(Some(&model_name))))
```

src/backend/probe/x86_64.rs
haraldh: again, I don't like the `transmute()` and `unwrap()`

src/backend/probe/x86_64.rs
haraldh: again, I don't like the `unwrap()`

haraldh (changes requested):

src/backend/probe/x86_64.rs
```diff
@@ -40,6 +53,53 @@ pub struct CpuId {
...
+        required_vendor: Option<Vendor>,
+    ) -> (bool, Option<String>) {
+        let manufacturer = {
+            let res = unsafe { __cpuid(0x00000000) };
+            let name: [u8; 12] = unsafe { transmute([res.ebx, res.edx, res.ecx]) };
+            from_utf8(&name[..]).unwrap().to_string()
+        };
+        let supported_cpu = required_vendor
+            .map(|v| v.id() == manufacturer)
+            .unwrap_or(true);
```
haraldh, line 69: why `true` on `unwrap()` ?

haraldh (changes requested):

haraldh (changes requested):
I think these 9 commits can be squashed into one.

haraldh:
perfect... and thanks for the long commit message

haraldh (approved):

PR #1305, opened 2022-02-01
Display system info using `uname` in `enarx info`.

### LINKED ISSUE #1206: Add uname -a output to enarx info
# Acceptance Criteria:
- Use the `uname()` syscall instead of shelling out.

definitelynobody:
- Uses the `uname` syscall to retrieve the system info.
- Adds a `errno` dependency for displaying errors nicely.

Closes #1206

image

### DIFF
--- a/src/backend/probe/common.rs
+++ b/src/backend/probe/common.rs
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::backend::Datum;
+
+use libc::{uname, utsname};
+use std::{ffi::CStr, io, mem::MaybeUninit, os::raw::c_char, str::Utf8Error};
+
+pub fn system_info() -> Datum {
+    fn system_info_string(utsname: utsname) -> Result<String, Utf8Error> {
+        fn array_to_str(array: &'_ [c_char; 65]) -> Result<&'_ str, Utf8Error> {
+            unsafe { CStr::from_ptr(array.as_ptr()) }.to_str()
+        }
+
+        Ok(format!(
+            "{} {} {} {} {} {}",
+            array_to_str(&utsname.sysname)?,
+            array_to_str(&utsname.nodename)?,
+            array_to_str(&utsname.release)?,
+            array_to_str(&utsname.version)?,
+            array_to_str(&utsname.machine)?,
+            array_to_str(&utsname.domainname)?,
+        ))
+    }
+
+    let mut utsname = MaybeUninit::uninit();
+
+    Datum {
+        name: "System Info".to_string(),
+        pass: true,
+        info: if unsafe { uname(utsname.as_mut_ptr()) } != 0 {
+            Some(format!("[{}]", io::Error::last_os_error()))
+        } else {
+            Some(
+                system_info_string(unsafe { utsname.assume_init() })
+                    .unwrap_or_else(|e| format!("[utf8 error: {}]", e)),
+            )
+        },
+        mesg: None,
+    }
+}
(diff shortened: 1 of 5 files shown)


npmccallum:
@definitelynobody Please fix the failing test.

src/backend/probe/common.rs
```diff
@@ -0,0 +1,45 @@
...
+    // If uname returns non-zero then an error occurred.
+    let pass = unsafe { uname(utsname.as_mut_ptr()) } == 0;
+
+    Datum {
+        name: "System Info".to_string(),
+        pass,
+        info: Some(if !pass {
+            // Retrieve the last errno value and display it
+            // as a human-readable string.
+            format!("[{}]", errno())
```
haraldh, line 38: Can't we use [`std::io::Error::last_os_error()`](https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error) ??

src/backend/probe/common.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use std::{ffi::CStr, mem::MaybeUninit, os::raw::c_char, str::Utf8Error};
+
+use errno::errno;
+
+use libc::{uname, utsname};
+
+pub fn system_info() -> Datum {
+    fn system_info_string(mut utsname: utsname) -> Result<String, Utf8Error> {
+        fn array_to_str(array: &'_ mut [c_char; 65]) -> Result<&'_ str, Utf8Error> {
```
rvolosatovs, line 13: Does this need to be `mut`?

src/backend/probe/common.rs
definitelynobody: yeah I didn't know about that! I'll update it

src/backend/probe/common.rs
```diff
@@ -0,0 +1,45 @@
...
+        name: "System Info".to_string(),
+        pass,
+        info: Some(if !pass {
+            // Retrieve the last errno value and display it
+            // as a human-readable string.
+            format!("[{}]", errno())
+        } else {
+            system_info_string(unsafe { utsname.assume_init() })
+                .unwrap_or_else(|e| format!("[utf8 error: {}]", e))
+        }),
```
rvolosatovs, line 42: nit; How about something like this? That `Some` constructor at the beginning is a bit confusing IMO and `pass` is redundant
```suggestion
        info: if unsafe { uname(utsname.as_mut_ptr()) } != 0 {
            // Retrieve the last errno value and display it
            // as a human-readable string.
            Some(format!("[{}]", errno()))
        } else {
            Some(system_info_string(unsafe { utsname.assume_init() })
                .unwrap_or_else(|e| format!("[utf8 error: {}]", e)))
        }),
```

src/backend/probe/common.rs
definitelynobody: I don't think it does let me check

src/backend/probe/common.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::backend::Datum;
+
+use std::{ffi::CStr, mem::MaybeUninit, os::raw::c_char, str::Utf8Error};
+
+use errno::errno;
+
+use libc::{uname, utsname};
+
+pub fn system_info() -> Datum {
+    fn system_info_string(mut utsname: utsname) -> Result<String, Utf8Error> {
```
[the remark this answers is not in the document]
definitelynobody: what does "idem" mean, sorry for confusion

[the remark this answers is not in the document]
definitelynobody: do you think the `system_info_string` should be inlined because I'm kinda leaning in that direction

haraldh (approved):

src/backend/probe/common.rs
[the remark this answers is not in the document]
haraldh: https://en.wikipedia.org/wiki/Idem

idem is a Latin term meaning "the same"

@rvolosatovs was referencing to the `mut`

Issue #1212, opened 2022-01-12
sgx: Reburnish get_attestation() implementation


platten:
1. Import src/backend/sgx/attestation.rs either from 95dcb02ae6e4 (@platten) or 
   7f691c91783c (@haraldh). They provide the exact same version of the file.</del>
   - **Nothing** done. Upstream attestation.rs is the same, which means that this step is not required.
2. Import sgx crate with 41f347cf2a39 **reverted** (not cherry-picked) on top and make it compile together with the project.
   1. Fix compilation issues.
   2. Implement missing to_vec() to Attributes.
   3. Fix alignment of Attributes from 4 to 8.
   4. Add `ProductId` and `SecureVersionNumber` types. 
3. Reburnish internal/shim-sgx/src/handler/enarx.rs attestation so that it builds and functions correctly.

Note from @npmccallum to @jarkkojs:
 * DON'T resurrect the code from the `sgx` crate in the `sgx` crate. Move that code to the `enarx` crate for now.

# Acceptance Criteria

* [x] Remove https://github.com/enarx/enarx-shim-sgx and https://github.com/enarx/sgx crates from the Enarx Github project.
* [x] Fix aesmd has an availability issue: https://lore.kernel.org/linux-sgx/YfPYfsnjYXQsY6Jz@iki.fi/T/#t.
  * A cgroup does not fix this because cgroups do not support EPC limits.
  * The only way to pin pages for architectural enclaves is to put aesmd to a VM, as `/dev/sgx_vepc` is based on partitioning.
* [ ] Verify that an enclave is running on a genuine Intel CPU.

npmccallum:
Note that the code in the `sgx` crate probably doesn't belong there (at least yet). It should probably belong in the test for now.

jarkkojs:
What is `getatt()`? Not aware of such syscall.

npmccallum:
@jarkkojs https://github.com/enarx/enarx/wiki/Enarx-syscall:-get_attestation

jarkkojs:
Thanks.

jarkkojs:
This issue is blocked by https://github.com/enarx/enarx/issues/1331

jarkkojs:
All tests succeed, if I revert `55caa95c`

PR #1312, opened 2022-02-01
shim-sgx: Introduce REPORT microarchitectural data structure

jarkkojs:
Define a SGX REPORT representation as a Rust data structure but use Intel
SDM naming conventions and explicitly do not enforce documentation. It
makes the whole thing more convoluted, e.g. when comparing your definition
to the Intel SDM. Use only primitive types (integers, arrays) for field
because it is the most portable way to define these structs. This is not
the place to be over the top with Rust idioms :-)

Collect all uarch structs to a single file called sgx_uarch.rs, which
will be located for the moment in the root source directory of shim_sgx,
but will be eventually probably moved to some more "generic" location.

Link: #1212
Signed-off-by: Jarkko Sakkinen <jarkko.sakkinen@iki.fi>

### DIFF
--- a/internal/shim-sgx/src/lib.rs
+++ b/internal/shim-sgx/src/lib.rs
@@ -18,6 +18,7 @@
 pub mod entry;
 pub mod handler;
 pub mod heap;
+pub mod uarch;
 
 use sgx::parameters::{Attributes, Features, MiscSelect, Xfrm};
 
--- a/internal/shim-sgx/src/uarch.rs
+++ b/internal/shim-sgx/src/uarch.rs
@@ -1,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Contains microarchitecture structures of SGX. Keep the functionality
+//! minimal and stdlib free. Use the Intel SDM struct naming conventions,
+//! instead of Rust naming conventions.
+//!
+//! Chapter 34 of the Intel SDM contains detailed documentation for these
+//! structs.
+//!
+//! Guidelines for updating this file:
+//! * Only use primitive types and arrays for the fields.
+//! * Only use u8 arrays for reserved fields and padding.
+//! * Document the structs at top-level but never the fields.
+//! * Be super conservative when adding new functionality (e.g. trait
+//!   implementations).
+
+#![allow(non_camel_case_types)]
+#![allow(missing_docs)]
+
+use core::arch::asm;
+
+/// Description of the local attestation source enclave contents.
+#[derive(Clone, Copy)]
+#[repr(C)]
+pub struct REPORT_PAYLOAD {
+    pub cpusvn: [u64; 2],
+    pub miscselect: u32,
+    pub cet_attributes: u8,
+    reserved1: [u8; 11],
+    pub isvextnprodid: [u64; 2],
+    pub attributes: [u64; 2],
+    pub mrenclave: [u8; 32],
+    reserved2: [u8; 32],
+    pub mrsigner: [u8; 32],
+    reserved3: [u8; 32],
+    pub configid: [u8; 64],
+    pub isvprodid: u16,
+    pub isvsvn: u16,
+    pub configsvn: u16,
+    reserved4: [u8; 42],
+    pub isvfamilyid: [u64; 2],
+    pub reportdata: [u8; 64],
+}
+
+impl Default for REPORT_PAYLOAD {
+    fn default() -> Self {
+        REPORT_PAYLOAD {
+            cpusvn: [0, 0],
+            miscselect: 0,
+            cet_attributes: 0,
+            reserved1: [0; 11],
+            isvextnprodid: [0, 0],
+            attributes: [0, 0],
+            mrenclave: [0; 32],
+            reserved2: [0; 32],
+            mrsigner: [0; 32],
+            reserved3: [0; 32],
+            configid: [0; 64],
+            isvprodid: 0,
+            isvsvn: 0,
+            configsvn: 0,
+            reserved4: [0; 42],
+            isvfamilyid: [0, 0],
+            reportdata: [0; 64],
+        }
+    }
+}
+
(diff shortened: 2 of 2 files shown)


npmccallum (changes requested):
1. The `sgx` part of the `sgx_uarch` name is redundant. This crate is only for `sgx`. Just call it `uarch`.

2. Please configure your editor to run `cargo fmt` on every file save so that you don't trigger the test failures.

internal/shim-sgx/src/sgx_uarch.rs
```diff
@@ -0,0 +1,80 @@
...
+
+//! Contains microarchitecture structures of SGX. Keep the functionality
+//! minimal and stdlib free. Use the Intel SDM struct naming conventions,
+//! instead of Rust naming conventions.
+//!
+//! Chapter 34 of the Intel SDM contains detailed documentation for these
+//! structs.
+
+
+#[allow(non_camel_case_types)]
```
npmccallum, line 11: This doesn't do what you intend it to do.

If you want this to apply to all types in this file, it needs to be `#![allow(non_camel_case_types)]`. As it is currently coded, this only applies to `REPORT_BODY`.

internal/shim-sgx/src/sgx_uarch.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Contains the fields that are part of the CMAC, signed with the SGX report
+/// key.
+#[allow(missing_docs)]
```
npmccallum, line 15: This is not needed since you actually provide documentation.

internal/shim-sgx/src/sgx_uarch.rs
```diff
@@ -0,0 +1,80 @@
...
+            isvsvn: 0,
+            configsvn: 0,
+            reserved4: [0; 42],
+            isvfamilyid: [0, 0],
+            reportdata: [0; 64],
+        }
+    }
+}
+
+#[allow(missing_docs)]
```
npmccallum, line 62: Please just provide some documentation.

npmccallum (changes requested):

internal/shim-sgx/src/sgx_uarch.rs
jarkkojs: I can document the struct, but documenting separate fields does more harm than is a benefit.

npmccallum: @jarkkojs Fine. Just put `#![allow(missing_docs)]` at the top of the file.

jarkkojs: I'm about to push with TI included.

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -1,9 +1,35 @@
...
+    unsafe {
+        asm!(
+            "xchg       {RBX}, rbx",
+            "enclu",
+            "mov        rbx, {RBX}",
+
+            RBX = inout(reg) targetinfo => _,
+            in("rax") EREPORT,
+            in("rcx") reportdata.as_ptr(),
+            in("rdx") report.as_mut_ptr(),
```
npmccallum, line 26: Either clean up this function now or file a debt issue.

internal/shim-sgx/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 pub mod entry;
 pub mod handler;
 pub mod heap;
+pub mod sgx_uarch;
```
npmccallum, line 21: Rename to `uarch`.

internal/shim-sgx/src/sgx_uarch.rs
```diff
@@ -0,0 +1,118 @@
...
+impl Default for REPORT {
+    fn default() -> Self {
+        REPORT {
+            body: REPORT_PAYLOAD::default(),
+            keyid: [0; 32],
+            mac: [0, 0],
+            padding: [0; 79],
+        }
+    }
+}
```
npmccallum, line 118: The order should be:

1. `struct REPORT`
2. `impl Default for REPORT`
3. `struct TARGETINFO`
4. `impl Default for TARGETINFO`

npmccallum (changes requested):

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -1,9 +1,35 @@
...
+        );
+    }
+
+    unsafe { report.assume_init() }
+}
+
```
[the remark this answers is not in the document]
jarkkojs: Right, that will also remove the need for `dead_code`. It's cool.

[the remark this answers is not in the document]
jarkkojs: I'll still keep this instead of old `get_report`. It's not as fancy but it's very documentation, easy to trace (and grep) function name. You know exactly where it is coming from with this name.

internal/shim-sgx/src/handler/enarx.rs
jarkkojs: I create a deb issue, and my reasoning is:

- Before end-to-end my past experience test that it does not hurt to go over the top with not changing semantics (even obvious changes).
- It's a good first issue, right?

internal/shim-sgx/src/uarch.rs
```diff
@@ -0,0 +1,144 @@
...
+            miscselect: 0,
+            reserved2: [0; 8],
+            configid: [0; 64],
+            reserved3: [0; 384],
+        }
+    }
+}
+
+impl TARGETINFO {
+    pub fn enclu_ereport(&self, reportdata: &[u8; 64]) -> crate::uarch::REPORT {
```
npmccallum, line 122: `crate::uarch::REPORT` => `REPORT`

npmccallum (changes requested):
One last nitpick!

npmccallum (approved):

Issue #1208, opened 2022-01-12
Report enarx version to `exarx info` output


platten:

moksh-pathak:
Hey! I would like to take up this issue.

npmccallum:
@bstrie You most recently touched this code. Could you help @moksh-pathak get started?

npmccallum:
We will want to have the enarx version in both the json and non-json outputs. We probably want another backend-like object to report global information.

PR #1320, opened 2022-02-02
Enarx info enarx version

moksh-pathak:
Closes Issue #1208

Enarx version is printed whenever `enarx info` or `enarx info --json` is run on terminal.
Made changes in src/cli/info.rs
Made a struct Info, which is used when the `--json` flag is true. It prints the version no. and the backend.
When `--json` is not used, version no. is printed using `env!("CARGO_PKG_VERSION")`.

### DIFF
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -2,8 +2,11 @@
 
 use crate::backend::BACKENDS;
 use crate::cli::{Result, StructOpt};
+use crate::Backend;
+use core::fmt::Formatter;
+use serde::Serialize;
+use std::fmt;
 use std::ops::Deref;
-
 /// Show details about backend support on this system
 #[derive(StructOpt, Debug)]
 pub struct Options {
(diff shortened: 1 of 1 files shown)


src/cli/info.rs
```diff
@@ -20,8 +27,14 @@ impl Options {
...
+            let info = Info {
+                version: env!("CARGO_PKG_VERSION"),
+                backends,
+            };
+
+            println!("{}", serde_json::to_string_pretty(&info)?);
         } else {
+            println!("Version {} starting up", env!("CARGO_PKG_VERSION"));
             for backend in backends {
                 println!("Backend: {}", backend.name());
```
haraldh, line 39: You could implement `Display` also for `Info` and use it in both paths.

haraldh (changes requested):

npmccallum:
@moksh-pathak Could you paste an example of the json output?

src/cli/info.rs
moksh-pathak: I shall look into that.

moksh-pathak:
@npmccallum the example of json output is as follows
Screenshot from 2022-02-03 10-13-39

and the example of non json output is as follows
Screenshot from 2022-02-03 10-14-19


Hope this helps!

haraldh:
Lint is failing... https://github.com/enarx/enarx/runs/5068774406?check_suite_focus=true

haraldh:
please run `cargo fmt` before pushing

moksh-pathak:
> please run `cargo fmt` before pushing

Thanks for the suggestion @haraldh , same has been noted and implemented.

haraldh:
If the tests pass, you want to squash your commits.

haraldh (changes requested):
Please squash your commits to one, as it is small enough and corrections to previous commits should be squashed into them.

moksh-pathak:
Thanks for merging my PR @haraldh :)

PR #1329, opened 2022-02-03
Sgx attestation v2

jarkkojs:


npmccallum (changes requested):
Please rebase on main. Also, please fix the `cargo fmt` error. You should set up your editor to always run `cargo fmt` on save to prevent these issues.

npmccallum:
@jarkkojs The SGX attestation tests are failing.

haraldh (changes requested):
First we need this additional error check in the tests:
```diff
diff --git a/tests/c-tests/sgx_get_att_quote.c b/tests/c-tests/sgx_get_att_quote.c
index 8e729a1d..e77a0fb6 100644
--- a/tests/c-tests/sgx_get_att_quote.c
+++ b/tests/c-tests/sgx_get_att_quote.c
@@ -29,13 +29,13 @@ int main(void) {
 
     ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
 
+    if (size < 0)
+        return 1;
+
     /* this test is SGX-specific, so just return success if not running on SGX */
     if (technology != TEE_SGX)
         return 0;
 
-    if (size < 0)
-        return 1;
-
     /* check beginning of quote matches expected value */
     for (i = 0; i < 32; i++)
     {
diff --git a/tests/c-tests/sgx_get_att_quote_size.c b/tests/c-tests/sgx_get_att_quote_size.c
index c88e524a..1611dc5e 100644
--- a/tests/c-tests/sgx_get_att_quote_size.c
+++ b/tests/c-tests/sgx_get_att_quote_size.c
@@ -16,6 +16,9 @@ int main(void) {
 
     ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
 
+    if (size < 0)
+        return 0;
+
     /* this test is SGX-specific, so just return success if not running on SGX */
     if (technology != TEE_SGX)
         return 0;
```

Second, the host part of `SYS_ENARX_GETATT` is not yet wired up. It needs to talk to the aesmd via protobuf, IIRC.

 `SYS_ENARX_GETATT` has to be handled in the match of `src/backend/sgx/thread.rs:132`
 https://github.com/enarx/enarx/blob/4dc5ea568c8caa27d48467959eb2f79c8711e30b/src/backend/sgx/thread.rs#L132

jarkkojs:
I got full pass with my local machine and icelake when I revert AVX512 change.

haraldh:
> I got full pass with my local machine and icelake when I revert AVX512 change.

yes, because it basically returns an error with `size == -1`

haraldh:
basically:

```rust
        let req = request!(SYS_ENARX_GETATT => 0, 0, shim_buf_ptr.as_ptr(), REPORT_LEN);
```

will fail, because it's not implemented on the host side.

internal/shim-sgx/src/lib.rs
```diff
@@ -29,14 +29,7 @@ pub const ENCL_SIZE_BITS: u8 = 31;
...
 
-const XFRM: Xfrm = Xfrm::from_bits_truncate(
-    Xfrm::X87.bits()
-        | Xfrm::SSE.bits()
-        | Xfrm::AVX.bits()
-        | Xfrm::OPMASK.bits()
-        | Xfrm::ZMM_HI256.bits()
-        | Xfrm::HI16_ZMM.bits(),
-);
+const XFRM: Xfrm = Xfrm::from_bits_truncate(Xfrm::X87.bits() | Xfrm::SSE.bits());
```
rvolosatovs, line 32: Could you motivate this change? Perhaps we want to add a comment about it here as well?

npmccallum:
@jarkkojs Please fix the `cargo clippy` checks.

rvolosatovs (changes requested):

src/backend/sgx/thread.rs
```diff
@@ -14,6 +14,8 @@ use sgx::enclu::{EENTER, EEXIT, ERESUME};
 use sgx::ssa::Vector;
 use vdso::Symbol;
 
+use crate::backend::sgx::attestation::get_attestation;
```
rvolosatovs, line 17: nit; wrong ordering of imports

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+                msg.set_buf_size(SGX_QUOTE_SIZE as u32);
+                req.set_getQuoteExReq(msg);
+                Ok(req)
+            }
+        }
+    }
+
+    fn send_request(&self, req: Request, mut stream: UnixStream) -> Result<Response, Error> {
+        // Set up writer
+        let mut buf_wrtr = vec![0u8; size_of::<u32>()];
```
rvolosatovs, line 124: nit; this could be just `buf`

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+                };
+                let akid = match akid {
+                    Some(a) => a,
+                    None => {
+                        return Err(Error::new(
+                            ErrorKind::Other,
+                            "no attestation key ID provided for setting Get Quote Ex request",
+                        ));
+                    }
+                };
```
rvolosatovs, line 110: https://github.com/enarx/enarx/pull/1360#discussion_r802698938

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+            ReqType::KeySize => {
+                let akid = match akid {
+                    Some(a) => a,
+                    None => {
+                        return Err(Error::new(
+                            ErrorKind::Other,
+                            "no attestation key ID provided for setting Init Quote Ex request to get key size",
+                        ));
+                    }
+                };
```
rvolosatovs, line 84: idem, see review in https://github.com/enarx/enarx/pull/1360#pullrequestreview-877517987

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -1,8 +1,14 @@
...
+use sallyport::request;
+use sallyport::syscall::SGX_DUMMY_QUOTE;
+use sallyport::syscall::SGX_DUMMY_TI;
+use sallyport::syscall::SYS_ENARX_GETATT;
 use sallyport::syscall::{BaseSyscallHandler, EnarxSyscallHandler, SGX_QUOTE_SIZE, SGX_TECH};
 use sallyport::untrusted::{UntrustedRef, UntrustedRefMut, ValidateSlice};
-use sallyport::Reply;
+
+use crate::uarch::REPORT;
+use crate::uarch::TARGETINFO;
```
rvolosatovs, line 11: It's very easy to confuse these for constants, how about we rename these using "normal" Rust names on import to prevent this confusion? https://github.com/enarx/enarx/pull/1360#discussion_r802702282

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -1,8 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use sallyport::request;
+use sallyport::syscall::SGX_DUMMY_QUOTE;
+use sallyport::syscall::SGX_DUMMY_TI;
+use sallyport::syscall::SYS_ENARX_GETATT;
```
rvolosatovs, line 6: nit; these should be merged

internal/shim-sgx/src/handler/enarx.rs
haraldh: Originally they have been named correctly:
https://github.com/enarx/sgx/blob/e4d5cbe232b51b40c8003dedc7226e167e5224de/src/attestation_types/report.rs#L169-L172
and
https://github.com/enarx/sgx/blob/e4d5cbe232b51b40c8003dedc7226e167e5224de/src/attestation_types/ti.rs#L10-L13

nicely aligned and checked with `testaso!`

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+            0 => {
+                out_buf.copy_from_slice(&SGX_DUMMY_TI);
+                return Ok(SGX_TI_SIZE);
+            }
+            // Returns dummy Quote
+            _ => {
+                out_buf.copy_from_slice(&SGX_DUMMY_QUOTE);
+                return Ok(SGX_QUOTE_SIZE);
+            }
+        }
```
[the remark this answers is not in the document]
haraldh: This is copied code from https://github.com/enarx/enarx/blob/7f691c91783c841c807d7fc72c8038a2210d8007/src/backend/sgx/attestation.rs
don't blame this code to @jarkkojs

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+        let mut buf_wrtr = vec![0u8; size_of::<u32>()];
+        match req.write_to_writer(&mut buf_wrtr) {
+            Ok(_) => {}
+            Err(e) => {
+                return Err(Error::new(
+                    ErrorKind::Other,
+                    format!("invalid protobuf Request: {:?}. Error: {:?}", req, e),
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
haraldh: This is copied code from https://github.com/enarx/enarx/blob/7f691c91783c841c807d7fc72c8038a2210d8007/src/backend/sgx/attestation.rs
don't blame this code to @jarkkojs

src/backend/sgx/attestation.rs
haraldh: This is copied code from https://github.com/enarx/enarx/blob/7f691c91783c841c807d7fc72c8038a2210d8007/src/backend/sgx/attestation.rs
don't blame this code to @jarkkojs

src/backend/sgx/attestation.rs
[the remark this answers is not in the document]
rvolosatovs: We could file a debt issue with a reference here to clean-up the implementation, but we should do something about this regardless, right?
Contributors to our project will look at this code and this PR as an example and we should ensure they follow best practices.

internal/shim-sgx/src/handler/enarx.rs
jarkkojs: These names are aligned with SDM (see the comment in the file header).

jarkkojs: Inline comments in uarch structure fields make notorious to modify anything as it makes the struct more cluttered. Sometimes comments are harmful.

rvolosatovs: @jarkkojs I do not suggest renaming the types, but rather names they are imported as

jarkkojs: I updated struct names.

src/backend/sgx/attestation.rs
jarkkojs: `Write` trait is not implemented for array, e.g.

```
-        let mut buf_wrtr = vec![0u8; size_of::<u32>()];
+        let mut buf_wrtr: &[u8] = &[0, 0, 0, 0];
```

will result:

```
error[E0277]: the trait bound `&[u8]: std::io::Write` is not satisfied
   --> src/backend/sgx/attestation.rs:132:35
    |
132 |         match req.write_to_writer(&mut buf_wrtr) {
    |                                   ^^^^^^^^^^^^^ the trait `std::io::Write` is not implemented for `&[u8]`
    |
    = help: the following implementations were found:
              <&mut [u8] as std::io::Write>
    = note: `std::io::Write` is implemented for `&mut [u8]`, but not for `&[u8]`
    = note: required for the cast to the object type `dyn std::io::Write`
```

jarkkojs: For the record, same result comes with `let mut buf_wrtr: [u8; 4] = [0, 0, 0, 0];`.

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -12,18 +16,11 @@ impl<'a> EnarxSyscallHandler for super::Handler<'a> {
...
+        buf: UntrustedRefMut<'_, u8>,
+        buf_len: libc::size_t,
     ) -> sallyport::Result {
-        self.trace("get_att", 0);
-
-        match hash.validate_slice(hash_len, self) {
-            None => Reply::from(Ok([SGX_QUOTE_SIZE.into(), SGX_TECH.into()])),
-            Some(..) => Reply::from(Err(libc::ENOSYS)),
-        }
-        .into()
```
haraldh, line 24: please don't remove the validation, even, if it currently does nothing.

haraldh (changes requested):

src/backend/sgx/attestation.rs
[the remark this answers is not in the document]
jarkkojs:
```
        // Set up writer
        req.write_to_writer(&mut buf_wrtr).ok_or_else(|| {
            Error::new(
                ErrorKind::Other,
                format!("Invalid protobuf request: {:?}", req),
            )
        })?;
```

This will give compilation error:

```
   --> src/backend/sgx/attestation.rs:111:44
    |
111 |         req.write_to_writer(&mut buf_wrtr).ok_or_else(|| {
    |                                            ^^^^^^^^^^ help: there is an associated function with a similar name: `map_or_else`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `enarx` due to previous error
```

internal/shim-sgx/src/handler/enarx.rs
jarkkojs: I cannot find a commit that would anything to validation between `mainline/main` and `origin/sgx_attestation_v2`.

jarkkojs:
```
        let hash = match hash.validate_slice(hash_len, self) {
            None => {
                let rep: sallyport::Reply = Ok([SGX_QUOTE_SIZE.into(), SGX_TECH.into()]).into();
                return sallyport::Result::from(rep);
            }
            Some(h) => {
                if h.len() != 64 {
                    return Err(libc::EINVAL);
                }
                let mut hash = [0u8; 64];
                hash.copy_from_slice(h);
                hash
            }
        };
```
Doesn't this already take care of the validation?

src/backend/sgx/attestation.rs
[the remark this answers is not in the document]
jarkkojs: Now that all validation is fixed in `set_request()` can we close this? Not the prettiest code but it is still **working** code.

internal/shim-sgx/src/handler/enarx.rs
haraldh: yes... I apologize... I shouldn't do reviews on the mobile after waking up.

jarkkojs: NP :-)

internal/shim-sgx/src/uarch.rs
```diff
@@ -70,17 +68,17 @@ impl Default for REPORT_PAYLOAD {
...
     pub mac: [u64; 2],
     padding: [u8; 80],
 }
 
-impl Default for REPORT {
+impl Default for Report {
     fn default() -> Self {
-        REPORT {
-            body: REPORT_PAYLOAD::default(),
+        Report {
```
rvolosatovs, line 80: nit; this could also be
```suggestion
        Self {
```

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+        &self,
+        report: Option<&[u8; 432]>,
+        akid: Option<Vec<u8>>,
+        size: Option<usize>,
+    ) -> Result<Request, Error> {
+        let mut req = Request::new();
+
+        match self {
+            ReqType::AkIdNum => {
+                // removed
```
rvolosatovs, line 45: For a person unfamiliar with internals of this, the comment does not make it clearer - what exactly is removed? From where? It looks like there should be a `TODO` with an issue reference instead

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+                    )
+                })?;
+                let size = size.ok_or_else(|| {
+                    Error::new(ErrorKind::Other, "Missing key size from InitQuoteExRequest")
+                })?;
+                let mut msg = Request_InitQuoteExRequest::new();
+                msg.set_timeout(AESM_REQUEST_TIMEOUT);
+                msg.set_b_pub_key_id(true);
+                msg.set_att_key_id(akid);
+                msg.set_buf_size(size as u64);
```
rvolosatovs, line 66: not sure about internals of this, but if these fields always have to be set for `Request_InitQuoteExRequest` to be valid, then they should be parameters in the constructor and these setters should be removed. (I'm fine with a debt issue if that's the case)

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+                    Error::new(
+                        ErrorKind::Other,
+                        "Missing REPORT structure from GetQuoteExRequest",
+                    )
+                })?;
+                let mut msg = Request_GetQuoteExRequest::new();
+                msg.set_timeout(AESM_REQUEST_TIMEOUT);
+                msg.set_report(report.to_vec());
+                msg.set_att_key_id(akid);
+                msg.set_buf_size(SGX_QUOTE_SIZE as u32);
```
rvolosatovs, line 101: idem

src/backend/sgx/attestation.rs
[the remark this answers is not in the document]
rvolosatovs: Can't we do something like this?
```suggestion
       req.write_to_writer(&mut buf_wrtr).map_err(|e| Error::new(
                    ErrorKind::Other,
                    format!("invalid protobuf Request: {:?}. Error: {:?}", req, e),
                ))?;
```

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+                req.set_getQuoteExReq(msg);
+                Ok(req)
+            }
+        }
+    }
+
+    fn send_request(&self, req: Request, stream: &mut UnixStream) -> Result<Response, Error> {
+        // Set up writer
+        let mut buf_wrtr = vec![0u8; size_of::<u32>()];
```
rvolosatovs, line 110: It is implemented for a mutable borrow of a slice https://doc.rust-lang.org/std/primitive.slice.html#impl-Write
And https://doc.rust-lang.org/std/primitive.array.html#method.as_mut on an array returns a `&mut [u8]`, therefore `buf_wrtr.as_mut()` implements a `Writer`, where `let buf_wrtr = [0u8; 4]`

Also, `_wrtr` suffix only makes it less clear what it is, `buf` would be way more conventional here

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+    let pb_msg: Response = r.send_request(req, &mut stream)?;
+
+    let res = pb_msg.get_getSupportedAttKeyIDNumRes();
+
+    if res.get_errorCode() != 0 {
+        panic!(
+            "Received error code {:?} in GetSupportedAttKeyIDNum",
+            res.get_errorCode()
+        );
+    }
```
rvolosatovs, line 159: Do we really want to panic here? At the very least, there should be a comment explaining why we're panicking in production code

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    let num_key_ids = res.get_att_key_id_num();
+
+    let expected_buffer_size: u32 = num_key_ids * SGX_KEY_ID_SIZE;
+    assert_eq!(1, num_key_ids);
```
rvolosatovs, line 164:
```suggestion
    debug_assert_eq!(1, num_key_ids);
```

Unless we want to panic in production

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+    msg.set_timeout(AESM_REQUEST_TIMEOUT);
+    msg.set_buf_size(expected_buffer_size);
+    req.set_getSupportedAttKeyIDsReq(msg);
+
+    let pb_msg: Response = r.send_request(req, &mut stream)?;
+
+    let res = pb_msg.get_getSupportedAttKeyIDsRes();
+
+    if res.get_errorCode() != 0 {
+        panic!(
```
rvolosatovs, line 180: idem

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+    let mut stream = UnixStream::connect(AESM_SOCKET)?;
+
+    let r = ReqType::TInfo;
+    let pb_req = r.set_request(None, Some(akid), Some(size))?;
+    let mut pb_msg: Response = r.send_request(pb_req, &mut stream)?;
+
+    let res: Response_InitQuoteExResponse = pb_msg.take_initQuoteExRes();
+    let ti = res.get_target_info();
+
+    assert_eq!(
```
rvolosatovs, line 211: idem

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    assert!(!key_ids.is_empty());
+
+    Ok(key_ids.get(0).unwrap().clone())
+}
+
+/// Fills the Target Info of the QE into the output buffer specified and
+/// returns the number of bytes written.
+fn get_ti(akid: Vec<u8>, size: usize, out_buf: &mut [u8]) -> Result<usize, Error> {
+    assert_eq!(out_buf.len(), SGX_TI_SIZE, "Invalid size of output buffer");
```
rvolosatovs, line 200: idem

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+        );
+    }
+
+    let key_ids_blob = res.get_att_key_ids();
+    let key_ids: Vec<Vec<u8>> = key_ids_blob
+        .chunks_exact(SGX_KEY_ID_SIZE as usize)
+        .map(Vec::from)
+        .collect();
+
+    assert!(!key_ids.is_empty());
```
rvolosatovs, line 192: idem

rvolosatovs (dismissed):

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -55,37 +52,27 @@ impl<'a> EnarxSyscallHandler for super::Handler<'a> {
...
         // create Report from TargetInfo.
-        let mut ti = [0u8; 512];
-        let ti_len = ti.len();
+        let mut ti_buf = [0u8; 512];
+        let ti_len = ti_buf.len();
         let c = self.new_cursor();
 
         unsafe {
-            c.copy_into_slice(buf_len, &mut ti[..ti_len])
+            c.copy_into_slice(buf_len, &mut ti_buf[..ti_len])
```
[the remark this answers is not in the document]
jarkkojs: You're absolutely right.

internal/shim-sgx/src/uarch.rs
```diff
@@ -104,9 +102,9 @@ pub struct TARGETINFO {
     reserved3: [u8; 384],
 }
 
-impl Default for TARGETINFO {
+impl Default for TargetInfo {
     fn default() -> Self {
-        TARGETINFO {
+        TargetInfo {
```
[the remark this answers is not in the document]
jarkkojs: It's already like that.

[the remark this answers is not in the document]
jarkkojs: But perhaps also `Self {` here?

internal/shim-sgx/src/uarch.rs
jarkkojs: I'll fix it.

jarkkojs:
> Please squash the "cleanup" commits together with the commits where implementation is introduced

I tend to completely disagree with this when it comes to SGX attestation. In this case we do not want to take even a minor risk to break working code.

jarkkojs:
> > Please squash the "cleanup" commits together with the commits where implementation is introduced
> 
> I tend to completely disagree with this when it comes to SGX attestation. In this case we do not want to take even a minor risk to break working code. I can squash the remaining clean ups to the existing commits where I have the clean ups so far but refuse to do this.

I'll address remaining comments, create a git tag, and after that squash clean/fixups to feature commits. After that I'll check with `git diff` that no unintentional changes have been caused. This should be a sane way to close this PR in a manner that we keep attestation code not breaking. This was my main concern.

internal/shim-sgx/src/handler/enarx.rs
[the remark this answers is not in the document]
jarkkojs: `ti_len` is completely unnecessary.

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -55,37 +52,27 @@ impl<'a> EnarxSyscallHandler for super::Handler<'a> {
 
         // Retrieve TargetInfo from sallyport block and call EREPORT to
         // create Report from TargetInfo.
-        let mut ti = [0u8; 512];
-        let ti_len = ti.len();
+        let mut ti_buf = [0u8; 512];
```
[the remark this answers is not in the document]
jarkkojs: I'm looking at this code:
```
        // Generate Report
        let mut target_info: TargetInfo = Default::default();
        let mut f = [0u8; 8];
        let mut x = [0u8; 8];
        f.copy_from_slice(&ti_buf[32..40]);
        x.copy_from_slice(&ti_buf[40..48]);
        let f = u64::from_le_bytes(f);
        let x = u64::from_le_bytes(x);
        target_info.mrenclave.copy_from_slice(&ti_buf[0..32]);
        target_info.attributes = [f, x];
```

@haraldh why cannot just take a direct copy of `ti_buf`?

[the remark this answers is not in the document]
haraldh: I have no idea... never reviewed that code.. I know that is doing that, because of wanting to fill out the structure in a non-unsafe way and the `attributes` are `packed` and mis-aligned u64.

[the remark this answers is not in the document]
haraldh:
```
#[repr(C, packed(4))]
pub struct Attributes {
...
```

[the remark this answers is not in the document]
haraldh: needed for:
```
#[repr(C)]
pub struct Measurement {
    misc: Masked<MiscSelect>,
    reserved0: [u8; 20],
    attr: Masked<Attributes>,
    mrenclave: [u8; 32],
    reserved1: [u8; 32],
    isv_prod_id: isv::ProdId,
    isv_svn: isv::Svn,
}
```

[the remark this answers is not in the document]
haraldh: If `Measurement` is not needed, one could remove the `packed` attribute from the `.attributes` field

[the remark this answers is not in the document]
jarkkojs: What do you think:
```rust
        let target_info =
            unsafe {
                let mut target_info_buf = [0u8; 512];
                c.copy_into_slice(512, &mut target_info_buf[..]).or(Err(libc::EFAULT))?;
                transmute::<[u8; 512], TargetInfo>(target_info_buf)
            };
```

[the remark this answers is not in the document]
jarkkojs: I'll go with this for the next version.

src/backend/sgx/attestation.rs
jarkkojs:
```
    if num_key_ids != 1 {
        return Err(Error::new(
            ErrorKind::Other,
            format!("Unexpected number of key IDs: {} != 1", num_key_ids),
        ));
    }
```
I think this is a sign of not having a legit fallback, because theoretically this could very well happen in production (if components out of our control give invalid data).

src/backend/sgx/attestation.rs
jarkkojs:
```
    if res.get_errorCode() != 0 {
        return Err(Error::new(
            ErrorKind::Other,
            format!("Received error code {:?} in GetSupportedAttKeyIDNum", res.get_errorCode());
        );
    }
```

src/backend/sgx/attestation.rs
jarkkojs:
```rust
```
    if res.get_errorCode() != 0 {
        return Err(Error::new(
            ErrorKind::Other,
            format!("Received error code {:?} in GetSupportedAttKeyIDNum", res.get_errorCode());
        );
    }
```

File: src/backend/sgx/attestation.rs (hunk 19 above)
Comment (jarkkojs) [reply src/backend/sgx/attestation.rs:211]: IMHO this assert is obsolete but in the beginning of the function there is `assert_eq()` that I converted as `debug_assert_eq()`. I'll just remove this one.

Comment (jarkkojs) [reply src/backend/sgx/attestation.rs:211]: Also the copy operation is after this check so the error message does not make any sense.

File: src/backend/sgx/attestation.rs (hunk 12 above)
Comment (jarkkojs) [reply src/backend/sgx/attestation.rs:45]: I added FIXME.

File: src/backend/sgx/attestation.rs [hunk 25]
```diff
@@ -0,0 +1,365 @@
...
+    ) -> Result<Request, Error> {
+        let mut req = Request::new();
+
+        match self {
+            ReqType::AkIdNum => {
+                // removed
+                unimplemented!()
+            }
+            ReqType::AkId => {
+                // removed
```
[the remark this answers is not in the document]
jarkkojs: ditto

src/backend/sgx/attestation.rs
jarkkojs: I think this is not possible, as it is generated code by protobuf AFAIK. @haraldh ?

src/backend/sgx/attestation.rs
jarkkojs:
```
        // Set up writer
        let mut buf_wrtr = [0u8; 4].as_mut();

        req.write_to_writer(buf_wrtr).map_err(|e| Error::new(
            ErrorKind::Other,
            format!("Invalid protobuf request: {:?}. Error: {:?}", req, e),
        ))?;
```

Result:

```
   Compiling enarx v0.2.0 (/home/jarkko/Projects/Enarx/enarx)
error[E0277]: the trait bound `[u8]: std::io::Write` is not satisfied
   --> src/backend/sgx/attestation.rs:112:29
    |
112 |         req.write_to_writer(buf_wrtr).map_err(|e| Error::new(
    |                             ^^^^^^^^ the trait `std::io::Write` is not implemented for `[u8]`
    |
    = help: the following implementations were found:
              <&mut [u8] as std::io::Write>
    = note: required for the cast to the object type `dyn std::io::Write`

error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> src/backend/sgx/attestation.rs:112:29
    |
112 |         req.write_to_writer(buf_wrtr).map_err(|e| Error::new(
    |                             ^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `[u8]`
    = note: required for the cast to the object type `dyn std::io::Write`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `enarx` due to 2 previous errors
```

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        // Receive Response
+        let mut res_len_bytes = [0u8; 4];
+        stream.read_exact(&mut res_len_bytes)?;
+        let res_len = u32::from_le_bytes(res_len_bytes);
+
+        let mut res_bytes = vec![0; res_len as usize];
+        stream.read_exact(&mut res_bytes)?;
+
+        Ok(Message::parse_from_bytes(&res_bytes)?)
```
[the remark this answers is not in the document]
jarkkojs: I split it.

src/backend/sgx/attestation.rs
jarkkojs: IMHO (or at least in kernel as of today) both panic or assert are strictly forbidden so that e.g. you can do forensics to a kernel even if the problem is because of programming error. E.g. you are guaranteed to get a NAK if you use `BUG_ON()` macro.

jarkkojs:
@haraldh, @rvolosatovs, @npmccallum A new review round :-)

src/backend/sgx/attestation.rs
npmccallum: I believe this constructor is generated by protobuf.

jarkkojs: Yeah, so I put this resolved. I checked the whole compilation process to verify this.

jarkkojs:
Does SEV failure have something to do with these commits?

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,346 @@
...
+        return Err(Error::new(
+            ErrorKind::Other,
+            format!(
+                "GeSupportedAttKeyIDs: invalid count: {} != 1",
+                key_ids.len()
+            ),
+        ));
+    }
+
+    Ok(key_ids.get(0).unwrap().clone())
```
jarkkojs, line 205: Prbly not worth of NAK but could be opened into two separate statements as @rvolosatovs proposed in some other occasion.

haraldh:
```diff
diff --git a/internal/shim-sgx/src/handler/enarx.rs b/internal/shim-sgx/src/handler/enarx.rs
index 163a0788..f0d92d9d 100644
--- a/internal/shim-sgx/src/handler/enarx.rs
+++ b/internal/shim-sgx/src/handler/enarx.rs
@@ -2,8 +2,6 @@
 
 use crate::uarch::{Report, ReportData, TargetInfo};
 
-use core::mem::transmute;
-
 use sallyport::request;
 use sallyport::syscall::{
     BaseSyscallHandler, EnarxSyscallHandler, SGX_QUOTE_SIZE, SGX_TECH, SYS_ENARX_GETATT,
@@ -41,32 +39,31 @@ impl<'a> EnarxSyscallHandler for super::Handler<'a> {
         };
 
         // Used internally for buffer size to host when getting TargetInfo
-        const REPORT_LEN: usize = 512;
+        const TARGETINFO_SIZE: usize = core::mem::size_of::<TargetInfo>();
 
         // Validate output buf memory
         let buf = buf.validate_slice(buf_len, self).ok_or(libc::EFAULT)?;
 
         // Request TargetInfo from host by passing nonce as 0
         let c = self.new_cursor();
-        let (_, shim_buf_ptr) = c.alloc::<u8>(buf_len).or(Err(libc::EMSGSIZE))?;
-        let req = request!(SYS_ENARX_GETATT => 0, 0, shim_buf_ptr.as_ptr(), REPORT_LEN);
+        let (_, shim_buf_ptr) = c.alloc::<u8>(TARGETINFO_SIZE).or(Err(libc::EMSGSIZE))?;
+        let req = request!(SYS_ENARX_GETATT => 0, 0, shim_buf_ptr.as_ptr(), TARGETINFO_SIZE);
         unsafe { self.proxy(req)? };
 
         // Retrieve TargetInfo from sallyport block and call EREPORT to
         // create Report from TargetInfo.
         let c = self.new_cursor();
+        let (_c, src) = c.alloc::<u8>(TARGETINFO_SIZE).or(Err(libc::EFAULT))?;
 
-        // Generate Report
-        let target_info = unsafe {
-            let mut target_info_buf = [0u8; 512];
-            c.copy_into_slice(512, &mut target_info_buf[..])
-                .or(Err(libc::EFAULT))?;
-            transmute::<[u8; 512], TargetInfo>(target_info_buf)
-        };
+        // TargetInfo is align(512), but the sallyport buffer is not.
+        let target_info = unsafe { (src.as_ptr() as *const TargetInfo).read_unaligned() };
 
+        // Generate Report
         let report: Report = target_info.enclu_ereport(&ReportData(hash));
 
         // Request Quote from host
+
+        // Report is align(512), but for sallyport serialize to unaligned
         let report_slice = &[report];
         let report_bytes = unsafe {
             core::slice::from_raw_parts(
```

haraldh (changes requested):
Proposed changes:
https://github.com/jarkkojs/enarx/pull/17/files

haraldh:
> Proposed changes: https://github.com/jarkkojs/enarx/pull/17/files

force pushed correction

haraldh (changes requested):
Please squash my proposed changes into the other commits and rebase

jarkkojs:
> Please squash my proposed changes into the other commits and rebase

I will do this.

src/backend/sgx/attestation.rs
rvolosatovs: This is fine, since it's not a fallible operation anymore.

jarkkojs:
> > Please squash my proposed changes into the other commits and rebase
> 
> I will do this.

It's done.

haraldh (dismissed):

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,317 @@
...
+pub fn get_attestation(
+    nonce: usize,
+    nonce_len: usize,
+    buf: usize,
+    buf_len: usize,
+) -> Result<usize, Error> {
+    let out_buf: &mut [u8] = unsafe { from_raw_parts_mut(buf as *mut u8, buf_len) };
+
+    if nonce == 0 {
+        let akid = get_attestation_key_id().expect("error obtaining attestation key id");
```
rvolosatovs, line 295: nit; this could be outside the `if`

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,317 @@
...
+/// Returns the number of bytes written to the output buffer. Depending on
+/// whether the specified nonce is NULL, the output buffer will be filled with the
+/// Target Info for the QE, or a Quote verifying a Report.
+pub fn get_attestation(
+    nonce: usize,
+    nonce_len: usize,
+    buf: usize,
+    buf_len: usize,
+) -> Result<usize, Error> {
+    let out_buf: &mut [u8] = unsafe { from_raw_parts_mut(buf as *mut u8, buf_len) };
```
rvolosatovs, line 292: nit;
```suggestion
    let out_buf = unsafe { from_raw_parts_mut(buf as *mut u8, buf_len) };
```
compiler should be able to infer the type here

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,317 @@
...
+                SGX_QUOTE_SIZE
+            ),
+        ));
+    }
+
+    let mut req = AesmRequest::new();
+
+    let mut msg = Request_GetQuoteExRequest::new();
+    msg.set_timeout(AESM_REQUEST_TIMEOUT);
+    msg.set_report(report[0..432].to_vec());
```
rvolosatovs, line 250: Do we want to extract this (`432`) into a constant?

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,317 @@
...
+        ));
+    }
+
+    let mut req = AesmRequest::new();
+    let mut msg = Request_InitQuoteExRequest::new();
+
+    msg.set_timeout(AESM_REQUEST_TIMEOUT);
+    msg.set_b_pub_key_id(true);
+    msg.set_att_key_id(akid);
+    msg.set_buf_size(size as u64);
```
rvolosatovs, line 176: nit; compiler should be able to infer the type here
```suggestion
    msg.set_buf_size(size as _);
```

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,317 @@
...
+                res.get_errorCode()
+            ),
+        ));
+    }
+
+    Ok(res.get_att_key_id_num())
+}
+
+fn get_key_ids(num_key_ids: u32) -> Result<Vec<Vec<u8>>, Error> {
+    let expected_buffer_size: u32 = num_key_ids * SGX_KEY_ID_SIZE;
```
rvolosatovs, line 104:
```suggestion
    let expected_buffer_size = num_key_ids * SGX_KEY_ID_SIZE;
```

src/backend/sgx/attestation.rs
haraldh:
```
pub struct Report {
    pub payload: ReportPayload,
    pub keyid: [u8; 32],
    pub mac: [u64; 2],
    padding: [u8; 80],
}
```
So, the `432` is the size of `Report` minus the padding bytes ... `\o/`

rvolosatovs: At the very least there should be a comment here

haraldh: agreed

jarkkojs: I added `SGX_REPORT_SIZE` to `attestation.rs`.

src/backend/sgx/attestation.rs
rvolosatovs: Alright, so what does this actually do?

First, we write the raw protobuf bytes into the buffer. We don't know the size of the message we're going to read, therefore we *do* need the vector. https://github.com/enarx/enarx/blob/81fbf186cb33c08054352b74a025d0ebbb9bf0ef/src/protobuf/aesm-proto.proto#L37-L138

After that, on L59 we overwrite the first 4 bytes of the buffer with length of the request.

So I guess the intention is to store the length of the request in the first 4 bytes and append the actual protobuf message to the vector?

Why do we need this, if we already have a protobuf? Is this some specific binary format?
This most definitely lacks some documentation.

rvolosatovs: Looks like we should use https://docs.rs/protobuf/latest/protobuf/trait.Message.html#method.write_length_delimited_to_vec here instead

rvolosatovs (approved):
LGTM, great work @jarkkojs !

haraldh (dismissed):

jarkkojs:
> LGTM, great work @jarkkojs !

Ah, thank you :-) Sorry I missed this (just requested a new review).

haraldh (approved):

rvolosatovs (approved):

npmccallum (approved):

PR #1332, opened 2022-02-04
Add an opt-in script for pre-push fmt/clippy hooks

### LINKED ISSUE #1286: [Feature]: Setup git precommit hooks

### Description

It would be helpful to create a precommit hook to run formatting and linting checks before you commit.
If you want to ignore the precommit when you commit something you can just pass `-n` and it will be skipped so it's not too invasive.

### Acceptance Criteria

When commits are made a git precommit hook will run which will run various read-only checks.

### Suggestions for a technical implementation

I've used https://crates.io/crates/rusty-hook in the past for this.

.rusty-hook.toml

```
[hooks]
pre-commit = "cargo fmt -- --check && cargo clippy [ARGS HERE] && ..."
```

Add this to any Cargo.toml

```
[dev-dependencies]
rusty-hook = "^0.11.2"
```

We could run clippy a bunch of times to match what the pipeline does? (I use https://github.com/casey/just at home and run a lot of precommit hooks with `just precommit`, maybe we could consider that in the future or something similar)

definitelynobody:
This will run `cargo clippy` and `cargo fmt` checks in a read-only
way on `enarx` and internal crates.

Run `.git-hooks/toggle.sh` to try it.

Related to #1286

### DIFF
--- a/.git-hooks/toggle.sh
+++ b/.git-hooks/toggle.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+# This script allows users to easily opt-in to git hooks.
+set -e
+
+if [ -f ".git/hooks/pre-push" ]; then
+    echo "It looks like a pre-push hook is already installed."
+    read -e -p "Would you like to delete it? [y/N] " choice
+    [[ "$choice" == [Yy]* ]] || exit 0
+    echo "Removing hook"
+    rm .git/hooks/pre-push
+else
+    echo "This will install a pre-push Git hook for cargo fmt/clippy (read-only checks)."
+    echo "This will prevent you from pushing if either fail (but can be bypassed with --no-verify)."
+    read -e -p "Would you like to install it? [y/N] " choice
+    [[ "$choice" == [Yy]* ]] || exit 0
+    echo "Installing hook"
+    cp .git-hooks/pre-push .git/hooks/
+fi
--- a/.git-hooks/pre-push
+++ b/.git-hooks/pre-push
@@ -1,0 +1,16 @@
+#!/bin/sh
+set -xe
+
+# Run linting checks (not completely exhaustive)
+cargo clippy --all-features --manifest-path=Cargo.toml -- -D warnings
+cargo clippy --target=x86_64-unknown-linux-musl --all-features --manifest-path=internal/shim-sgx/Cargo.toml -- -D warnings
+cargo clippy --target=x86_64-unknown-linux-musl --all-features --manifest-path=internal/shim-sev/Cargo.toml -- -D warnings
+cargo clippy --target=x86_64-unknown-linux-musl --all-features --manifest-path=internal/wasmldr/Cargo.toml -- -D warnings
+
+# Run formatting tests
+cargo fmt -- --check
+cargo fmt --manifest-path=internal/shim-sgx/Cargo.toml -- --check
+cargo fmt --manifest-path=internal/shim-sev/Cargo.toml -- --check
+cargo fmt --manifest-path=internal/wasmldr/Cargo.toml -- --check
+
+exit 0


.git-hooks/toggle.sh
```diff
@@ -0,0 +1,18 @@
...
+    [[ "$choice" == [Yy]* ]] || exit 0
+    echo "Removing hook"
+    rm .git/hooks/pre-push
+else
+    echo "This will install a pre-push Git hook for cargo fmt/clippy (read-only checks)."
+    echo "This will prevent you from pushing if either fail (but can be bypassed with --no-verify)."
+    read -e -p "Would you like to install it? [y/N] " choice
+    [[ "$choice" == [Yy]* ]] || exit 0
+    echo "Installing hook"
+    cp .git-hooks/pre-push .git/hooks/
```
rvolosatovs, line 17: Depending on the implementation of `cp` this may silently overwrite the already-existing-hook, right?

.git-hooks/toggle.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# This script allows users to easily opt-in to git hooks.
+set -e
+
+if [ -f ".git/hooks/pre-push" ]; then
+    echo "It looks like a pre-push hook is already installed."
+    read -e -p "Would you like to delete it? [y/N] " choice
```
rvolosatovs, line 7: I think we should be explicit about deleting the file

.git-hooks/toggle.sh
```diff
@@ -0,0 +1,18 @@
...
+    [[ "$choice" == [Yy]* ]] || exit 0
+    echo "Removing hook"
+    rm .git/hooks/pre-push
+else
+    echo "This will install a pre-push Git hook for cargo fmt/clippy (read-only checks)."
+    echo "This will prevent you from pushing if either fail (but can be bypassed with --no-verify)."
+    read -e -p "Would you like to install it? [y/N] " choice
```
rvolosatovs, line 14: I think we should be explicit about possibly overwriting the file

rvolosatovs (commented):
Overall, I like it, but I am not sure about the `.git-hooks` directory and the toggle script value.

One of my concerns is, for example: If I already had custom hooks set up for the repository, I couldn't "just" use this - therefore I think it would be nicer to append rather than just copy (could be a follow-up).

As the end (future) result, I think having something like this would my ideal solution:
- `$tool hooks pre-push` - print `pre-push` hook to STDOUT (ditto for other types of hooks if we introduce any)
- `$tool hooks install` - create hook files, if not already present and then append after checking if the appropriate hook already exists per each file
- `$tool hooks remove` remove the previously-installed hooks (e.g. check for some commented-out tag added by `hooks install`, and remove that section)

Problem is that we don't have a `$tool` (yet?) to use AFAIK, so I guess that's not really an option.

I feel like I would prefer either:
a. simply adding contents of this hook to the wiki with steps on how to install it (or optional script to just copy-paste) - that's my preferred option, tbh
b. creating `contrib/git/hooks/pre-push` (or something other than `contrib`), where such "helper" files could live, we could then move the script to `contrib/git/install-hooks.sh`, which would append, preferably.

What do you think?

.git-hooks/toggle.sh
definitelynobody: I think it shouldn't because of the check above ` if [ -f ".git/hooks/pre-push" ]; then` right? If a file does exist it shouldn't go into this path.

definitelynobody:
@rvolosatovs I think we could definitely just move this to documentation maybe https://enarx.dev/docs/Contributing/Coding-Style or on a new page right after it for now considering it's not too hard to install manually.

It would be cool to have a tool be able to generate the pre-push script by finding all Cargo.toml manifests/features so we don't have to have the user do anything so maybe I could look into doing that at some point. We could add checks for merge commits, custom styling requirements, etc.

For now maybe I'll just put it in the docs?

rvolosatovs:
> For now maybe I'll just put it in the wiki?

Yes, please - I think that'd be the perfect place for this

PR #1360, opened 2022-02-09
Sgx attestation v2

haraldh:

### DIFF
--- a/internal/shim-sgx/src/handler/enarx.rs
+++ b/internal/shim-sgx/src/handler/enarx.rs
@@ -1,14 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use sallyport::request;
-use sallyport::syscall::SGX_DUMMY_QUOTE;
-use sallyport::syscall::SGX_DUMMY_TI;
 use sallyport::syscall::SYS_ENARX_GETATT;
 use sallyport::syscall::{BaseSyscallHandler, EnarxSyscallHandler, SGX_QUOTE_SIZE, SGX_TECH};
 use sallyport::untrusted::{UntrustedRef, UntrustedRefMut, ValidateSlice};
 
-use crate::uarch::REPORT;
 use crate::uarch::TARGETINFO;
+use crate::uarch::{ReportData, REPORT};
 
 impl<'a> EnarxSyscallHandler for super::Handler<'a> {
     // NOTE: The 'nonce' field is called 'hash' here, as it is used to pass in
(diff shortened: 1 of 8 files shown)


src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,372 @@
...
+            ReqType::TInfo => {
+                let akid = match akid {
+                    Some(a) => a,
+                    None => {
+                        return Err(Error::new(
+                            ErrorKind::Other,
+                            "no attestation key ID provided for setting Init Quote Ex request",
+                        ));
+                    }
+                };
```
rvolosatovs, line 52: How about something like this instead? https://doc.rust-lang.org/nightly/core/option/enum.Option.html#method.ok_or_else
```suggestion
                let akid = akid.ok_or_else(|| Error::new(
                               ErrorKind::Other,
                               "no attestation key ID provided for setting Init Quote Ex request",
                 ))?;
```

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,372 @@
...
+                let size = match size {
+                    Some(s) => s,
+                    None => {
+                        return Err(Error::new(
+                            ErrorKind::Other,
+                            "no key size provided for setting Init Quote Ex request",
+                        ));
+                    }
+                };
```
rvolosatovs, line 61: idem, match is unnecessary

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,372 @@
...
+                let mut msg = Request_InitQuoteExRequest::new();
+                msg.set_timeout(TIMEOUT);
+                msg.set_b_pub_key_id(true);
+                msg.set_att_key_id(akid);
+                msg.set_buf_size(size as u64);
+                req.set_initQuoteExReq(msg);
+                Ok(req)
+            }
+            ReqType::KeySize => {
+                let akid = match akid {
```
rvolosatovs, line 71: idem

internal/shim-sgx/src/handler/enarx.rs
```diff
@@ -1,8 +1,14 @@
...
+use sallyport::request;
+use sallyport::syscall::SGX_DUMMY_QUOTE;
+use sallyport::syscall::SGX_DUMMY_TI;
+use sallyport::syscall::SYS_ENARX_GETATT;
 use sallyport::syscall::{BaseSyscallHandler, EnarxSyscallHandler, SGX_QUOTE_SIZE, SGX_TECH};
 use sallyport::untrusted::{UntrustedRef, UntrustedRefMut, ValidateSlice};
-use sallyport::Reply;
+
+use crate::uarch::REPORT;
+use crate::uarch::TARGETINFO;
```
rvolosatovs, line 11: Is there a specific reason these are capitalized? If that was something specific to that particular submodule, how about we just rename these to "normal" Rust struct names (i.e. `Report` and `TargetInfo`) to avoid confusion?

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,372 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// Credit to: https://github.com/fortanix/rust-sgx/tree/master/aesm-client
```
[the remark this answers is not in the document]
jarkkojs: Updated.

PR #1380, opened 2022-02-14
fix(examples/tcp_server): promptly echo back input

haraldh:
The original `OpenBSD netcat` version does not issue
`shutdown(_, SHUT_WR)`, when the input pipe is closed.

Therefore the tcp_server input loop is never ended and nothing is echo'd
back.

To end with `netcat` the README.md is updated accordingly.

### DIFF
--- a/examples/tcp_server/src/main.rs
+++ b/examples/tcp_server/src/main.rs
@@ -161,6 +161,9 @@ fn handle_connection_event(
                     break;
                 }
                 Ok(n) => {
+                    // echo back what we received and ignore any errors while doing so
+                    let _ = connection.write_all(&received_data[bytes_read..(bytes_read + n)]);
+
                     bytes_read += n;
                     if bytes_read == received_data.len() {
                         received_data.resize(received_data.len() + 1024, 0);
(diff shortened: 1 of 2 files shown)


rvolosatovs (approved):

examples/tcp_server/src/main.rs
```diff
@@ -161,6 +161,9 @@ fn handle_connection_event(
                     break;
                 }
                 Ok(n) => {
+                    // echo back what we received and ignore any errors while doing so
+                    let _ = connection.write_all(&received_data[bytes_read..(bytes_read + n)]);
```
bstrie, line 165: This is just an example so it hardly matters, but I'm curious why we're ignoring errors here rather than using `?` like in the previous version.

bstrie (approved):

examples/tcp_server/src/main.rs
haraldh: If the other does a shutdown on the receiving side, we still want all what it sent and display that.

npmccallum (approved):

PR #1382, opened 2022-02-14
DRAFT: Rust sgx attestation test

haraldh:


jarkkojs:
Looks quite good to me. For Report etc. uarch stuff I think the redundancy for the moment if fine. Once I push SGX uarch to x86_64 crate we can then just migrate our stack to use it.

PR #1391, opened 2022-02-16
sgx: zero entire rflags, including alignment check (AC) flag.

jovanbulck:
Opening a PR, a discussed with @MikeCamel and @npmccallum via email:

I'm reaching out to you for a (minor) issue we spotted in Enarx's shim-sgx assembly entry code.

**Problem.** When reviewing the assembly code in enarx/internal/shim-sgx/src/main.rs I noticed that you currently only use `add/cld` to cleanse the x86 `rflags` register:

https://github.com/enarx/enarx/blob/99352a16ff0e0f070d8492c5deb8b173050e17bc/internal/shim-sgx/src/main.rs#L96-L98

However, there's also some more funny flags in there that can be abused, notably the `rflags.ac` flag. In short, when the AC flag in the `rflags` x86 configuration register is set, any unaligned data accesses performed by the enclave will force an #AC exception, forming a subtle controlled-channel attack primitive for the untrusted operating system. More info on this attack vector is in our earlier paper:

https://jovanbulck.github.io/files/ccs19-tale.pdf

Note that not clearing AC is "only" subtle side-channel leakage, so definitely not a major issue. However, #AC side-channel leakage can be easily prevented in software and has been addressed in all major production runtimes (Intel SDK, OpenEnclave, Rust-EDP, Gramine).

**Fix.** This PR proposes a minimal patch that fixes the issue by sanitizing `rflags` in the simplest and most complete way. Make sure to double-check and critically review my proposed patch (I didn't run it)!

On a related note, but this is somewhat less clear, my patch adds a `cld` instruction early-on, as it seems that you may call compiler-generated relocation Rust code _before_ `clearx` executes. This may be dangerous as the compiler is free to emit code that uses x86 string instructions (e.g., calls to `memcpy`/`memset`) and those will be corrupted in the wrong direction if `rflags.df` is set by the attacker(!)

(As discussed via email, it would be great if this commit would still land upstream within the next days, if possible. Then, I could still acknowledge and reference it in an overview table I'm preparing as part of the camera-ready version of a short discussion paper on ABI sanitization in main SGX runtimes.)

Let me know if you have any questions/comments!

Best regards,
Jo


npmccallum:
@jovanbulck Looks like you used the wrong asm syntax and thus the patch actually produces a build error.

internal/shim-sgx/src/main.rs
```diff
@@ -93,9 +93,11 @@ extern "sysv64" fn clearx() {
             "xor    r10,    r10",
             "xor    r11,    r11",
 
-            // Clear CPU flags
-            "add    r11,    r11",
-            "cld",
+            // Clear CPU state bits and DF/AC flags
+            // Note: we can simply popfq an all-zero value, as system flags and
+            // reserved bits are not writable from the user-space enclave
+            "pushq $0",
```
haraldh, line 99:
```suggestion
            "push DWORD PTR 0",
```

haraldh (changes requested):

jovanbulck:
Good catch, thanks! Sorry, I was not aware to use Intel syntax instead of AT&T.

I amended the commit with proper Intel syntax and verified that `cargo build` runs fine now on my machine.

npmccallum (approved):

haraldh (approved):

npmccallum:
@jovanbulck Thank you! Good luck on the paper.

npmccallum:
@jovanbulck Note that we will be releasing a point release today that includes this patch. @platten Please warm up the release engine. :)

jovanbulck:
Awesome! Thanks for the quick response on this :)

Issue #1395, opened 2022-02-17
[Bug]: SNP initialization is too fragile


npmccallum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

Enarx tries to bring up an SNP keep but spuriously fails with a "too busy" error. This really painful in CI because it causes the whole test to fail. We should retry a bit with a backoff delay, maybe up to 1-2 seconds.

### Expected Behaviour

Don't be so fragile.

### Environment Information

N/A

PR #1397, opened 2022-02-17
fix(sev): retry SNP `Launcher::init`

haraldh:
If there is a CI storm, where a lot of VMs are started, `/dev/sev` is sometimes busy:

```
Error: SNP Launcher init failed

Caused by:
    0: I/O Error
    1: Device or resource busy (os error 16)
```

Try to reopen before the CI process has to be rekicked.

To be discussed:
```rust
const SEV_RETRIES: usize = 3;
const SEV_RETRY_SLEEP_MS: u64 = 500;
```

Fixes: https://github.com/enarx/enarx/issues/1395


src/backend/sev/builder.rs
```diff
@@ -9,34 +9,71 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
+const SEV_RETRY_SLEEP_MS: u64 = 500;
+
 pub struct Builder {
     kvm_fd: Kvm,
     launcher: Launcher<Started, Firmware>,
     regions: Vec<Region>,
     sallyports: Vec<Option<VirtAddr>>,
 }
 
+pub fn retry<O, E, F>(func: F) -> Result<O, E>
```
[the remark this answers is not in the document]
haraldh: removed, although we are not a library, so nobody cares anyway

src/backend/sev/builder.rs
```diff
@@ -9,34 +9,71 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
+            Err(e) => {
+                if retries > 0 {
+                    retries -= 1;
+                    eprintln!("Error {:#?}.\nRetrying soon.", e);
+                    let millis =
+                        time::Duration::from_millis(SEV_RETRY_SLEEP_MS + rng.gen::<u8>() as u64);
+                    thread::sleep(millis);
+                    continue;
+                }
+                return Err(e);
```
rvolosatovs, line 53: nit; We could maybe get rid of the nested `if` by adding a guard? e.g. something like this:
```suggestion
            Err(e) if retries == 0 => {
                return Err(e);
            }
            Err(e) if retries > 0 {
                retries -= 1;
                eprintln!("Error {:#?}.\nRetrying soon.", e);
                let millis =
                    time::Duration::from_millis(SEV_RETRY_SLEEP_MS + rng.gen::<u8>() as u64);
                thread::sleep(millis);
                continue;
            }
```

Would be nice to log the retry number as well, maybe?

rvolosatovs (approved):

src/backend/sev/builder.rs
haraldh: addressed

rvolosatovs (dismissed):

src/backend/sev/builder.rs
```diff
@@ -9,34 +9,76 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
 }
 
+fn retry<O, E, F>(func: F) -> Result<O, E>
+where
+    F: Fn() -> Result<O, E>,
+    E: Debug,
+{
+    let mut retries = SEV_RETRIES;
+    let mut rng = thread_rng();
+    // try to open /dev/sev several times
```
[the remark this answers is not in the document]
haraldh: right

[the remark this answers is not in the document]
haraldh: corrected

src/backend/sev/builder.rs
```diff
@@ -9,34 +9,77 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
+            // try to open /dev/sev and start the Launcher several times
+
+            let kvm_fd = Kvm::new().context("Failed to open '/dev/kvm'")?;
+            let vm_fd = kvm_fd
+                .create_vm()
+                .context("Failed to create a virtual machine")?;
+
+            let sev = retry(|| Firmware::open().context("Failed to open '/dev/sev'"))?;
+            Launcher::new(vm_fd, sev)
+                .map(|launcher| (kvm_fd, launcher))
```
bstrie, line 80: I'm curious why this is a `.map` instead of using the `let launcher` like previously and just returning a tuple of `(kvm_fd, launcher)` at the end of the function.

rvolosatovs (approved):
LGTM

src/backend/sev/builder.rs
```diff
@@ -9,34 +9,71 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
 }
 
+fn retry<O, E, F>(func: F) -> Result<O, E>
+where
+    F: Fn() -> Result<O, E>,
+    E: Debug,
+{
+    let mut retries = SEV_RETRIES;
+    let mut rng = thread_rng();
+    // try to open /dev/sev several times
```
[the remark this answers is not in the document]
rvolosatovs: already addressed

src/backend/sev/builder.rs
haraldh: stuff is moved... try yourself

src/backend/sev/builder.rs
```diff
@@ -9,34 +9,71 @@ use crate::backend::kvm::builder::kvm_try_from_builder;
...
+    loop {
+        match func() {
+            Err(e) => {
+                if retries > 0 {
+                    retries -= 1;
+                    eprintln!("Error {:#?}.\nRetrying soon.", e);
+                    let millis =
+                        time::Duration::from_millis(SEV_RETRY_SLEEP_MS + rng.gen::<u8>() as u64);
```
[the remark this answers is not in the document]
haraldh: well, whenever there is a resource problem and all parties wait the same time... you will have the problem again later.

src/backend/sev/builder.rs
```diff
@@ -10,33 +10,75 @@ use crate::backend::kvm::mem::Region;
...
 pub struct Builder {
     kvm_fd: Kvm,
     launcher: Launcher<Started, Firmware>,
     regions: Vec<Region>,
     sallyports: Vec<Option<VirtAddr>>,
 }
 
+fn retry<O, F>(func: F) -> anyhow::Result<O>
+where
+    F: Fn() -> anyhow::Result<O>,
```
rvolosatovs, line 35: Can't we do this instead?
```suggestion
fn retry<O>(func: impl Fn() -> anyhow::Result<O>) -> anyhow::Result<O>
```

src/backend/sev/builder.rs
```diff
@@ -10,33 +10,75 @@ use crate::backend::kvm::mem::Region;
...
+                .create_vm()
+                .context("Failed to create a virtual machine")?;
+
+            let sev = retry(|| Firmware::open().context("Failed to open '/dev/sev'"))?;
 
-        let sev = Firmware::open().context("Failed to open '/dev/sev'")?;
+            let launcher = Launcher::new(vm_fd, sev).context("SNP Launcher init failed")?;
 
-        let launcher = Launcher::new(vm_fd, sev).context("SNP Launcher init failed")?;
+            Ok((kvm_fd, launcher))
```
rvolosatovs, line 80: not that it really matters, but I actually preferred the original version with a map
Not entirely sure what we gain with this refactor commit

bstrie: I find it poor for readability to smuggle the returned tuple inside a map, especially when what should logically be the "returned" thing is immediately followed by a error handling for an operation unrelated to the return. I'd prefer to keep the launcher creation logic self-contained, and return the final tuple as its own statement.

The reason that the map was used in the first place was for type-system-tetris to avoid a type annotation which arose from being overly generic on the return type of `retry`. This was itself an instance of poor style, since we had both `std::result::Result` and `anyhow::Result` in scope as simply `Result` and were relying on the number of type parameters to distinguish them, but that makes for poor readability. In any case, we didn't need to be using `std::result::Result` as the return type of `retry` since we only use `retry` with `anyhow::Result`. In general it's just good style not to have two types in scope with the exact same name; given usual Rust naming conventions this usually only occurs for Result types (and their variants), so it is encouraged to use them namespaced as `foo::Result` to reduce confusion. And by fixing this style nit (by just using `anyhow::Result` everywhere) we also remove the impetus for using `.map` to avoid the type inference failure.

src/backend/sev/builder.rs
rvolosatovs: I think if we can avoid naming generic type parameters, then we should, especially in cases like this when there's only 1 such parameter

rvolosatovs (dismissed):
If @haraldh agrees, then still LGTM, whether with @bstrie commit or without.

bstrie:
Yes, I'll let haraldh have the chance to give feedback on my commit. I don't want to approve in case it decides to merge automatically...

src/backend/sev/builder.rs
haraldh: fixed with your PR.. thanks

src/backend/sev/builder.rs
haraldh: done

haraldh:
@bstrie @rvolosatovs next round of bike shedding :smile:

rvolosatovs (approved):

npmccallum (approved):

Issue #1402, opened 2022-02-17
[Bug]: SGX attestation tests being executed on non-SGX platforms


platten:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

The SGX attestation test is executed when running `cargo test` on a SEV-SNP/Milan system:

```console
test backend::sev::snp::tests::test_const_id_macro ... ok
test backend::sev::snp::certs::test::test_vcek_url ... ok
test backend::sgx::attestation::tests::request_target_info ... FAILED
test cli::sev::tests::test_merge_vcek_stack ... ok
test workldr::wasmldr::test::is_builtin ... ok

failures:

---- backend::sgx::attestation::tests::request_target_info stdout ----
thread 'backend::sgx::attestation::tests::request_target_info' panicked at 'error obtaining attestation key id: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/backend/sgx/attestation.rs:294:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    backend::sgx::attestation::tests::request_target_info

test result: FAILED. 4 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass '--bin enarx'
```

### Expected Behaviour

Test should be skipped on non-SGX machines.

### Environment Information

```console
Version 0.2.1 starting up
Backend: sgx
 ✔ System Info: Linux milan.sev.lab.enarx.dev 5.14.0-snp6 #1 SMP Tue Jan 18 18:29:40 UTC 2022 x86_64 (none)
 ✗ Driver: /dev/sgx_enclave
 ✗ CPU: AMD EPYC 7543 32-Core Processor                 | AuthenticAMD
 ✗  SGX Support
 ✗   Version 1
 ✗   Version 2
 ✗   FLC Support
 ✗   Max Size (32-bit)
 ✗   Max Size (64-bit)
 ✗   MiscSelect
 ✗   Features
 ✗   Xfrm
 ✗   EPC Size
Backend: sev
 ✔ System Info: Linux milan.sev.lab.enarx.dev 5.14.0-snp6 #1 SMP Tue Jan 18 18:29:40 UTC 2022 x86_64 (none)
 ✔ Driver: /dev/sev
 ✔  SEV-SNP is enabled in host kernel
 ✔  /dev/sev is readable by user
[...]

PR #1404, opened 2022-02-18
fix[sgx]: attestation test

jarkkojs:
Check that SGX is enabled before running the test.

Closes: https://github.com/enarx/enarx/issues/1402

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
-
 use std::collections::HashMap;
 use std::ffi::OsStr;
-use std::fs::OpenOptions;
+use std::fs;
+use std::os::unix::fs::FileTypeExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use walkdir::WalkDir;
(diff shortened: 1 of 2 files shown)


haraldh (approved):

haraldh:
@rvolosatovs it should only be used for tests then... and should be probably named: `test_can_use_sgx` or so

jarkkojs:
updated:
```
    #[test]
    fn request_target_info() {
        use crate::backend::sgx::data::dev_sgx_enclave;
        use std::path::Path;

        if !dev_sgx_enclave().pass || !Path::new(AESM_SOCKET).exists() {
            return;
        }

        let output = [1u8; SGX_TI_SIZE];
        assert_eq!(
            get_attestation(0, 0, output.as_ptr() as usize, output.len()).unwrap(),
            SGX_TI_SIZE
        );
    }
```

jarkkojs:
Updated based on @rvolosatovs feedback.`build.rs` has now `has_sgx`.

rvolosatovs (dismissed):

rvolosatovs:
Please make sure tests pass, otherwise LGTM

npmccallum (approved):

PR #1422, opened 2022-02-21
build: read prebuilt internal bins from `ENARX_PREBUILT_*`

veehaitch:
The Enarx build script also builds the internal binaries (`shim-sev`,
`shim-sgx`, `wasmldr`) to embed them in the final `enarx` binary.

This approach makes it difficult to build the project if the build phase
does not have any network access. The Nix package manager is a notable
example. Therefore, it is useful to have the option to build the
internal dependencies independently before building `enarx`.

To make that possible, this commit reads the `ENARX_PREBUILT_*`
environment variable when building an internal binary. If the
environment variable is set, it should point to the corresponding
prebuilt binary. In this case, it copies the binary to the expected
output path (`out_bin`).

If the variable is not set, cargo builds the respective binary as
before.

Signed-off-by: Vincent Haupert <mail@vincent-haupert.de>

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -81,6 +81,20 @@ fn cargo_build_bin(
     target_name: &str,
     bin_name: &str,
 ) -> std::io::Result<()> {
+    // And here's where we'd like to place the final (stripped) binary
+    let out_bin = out_dir.join("bin").join(bin_name);
+
+    // Don't run the build if ENARX_PREBUILT_${bin_name} is set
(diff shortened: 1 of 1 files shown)


npmccallum:
@veehaitch Thanks for the PR! I don't have a problem with this strategy and have assigned @haraldh to review it. However, please note that we are sponsoring this PR: https://github.com/rust-lang/cargo/pull/9992 . So our custom hacky stuff will be going away hopefully soon.

haraldh (approved):

PR #1443, opened 2022-02-22
fix(test): use `/usr/bin/env` shebang

veehaitch:
On NixOS, neither `/bin/sh` nor `/usr/bin/bash` exists. This means, when
developing on NixOS with the given `flake.nix`, calling `cargo test`
partly fails as the given path in the helper scripts' shebang does not
exist.

For some compatibility, NixOS offers `/usr/bin/env`. Therefore, this
commit uses `/usr/bin/env` to call `sh` and `bash` relative to `PATH`.

Signed-off-by: Vincent Haupert <mail@vincent-haupert.de>

### DIFF
--- a/helper/parse-trace.sh
+++ b/helper/parse-trace.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 SHIM="$1"
 PAYLOAD="$2"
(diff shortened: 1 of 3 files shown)


rvolosatovs (dismissed):

npmccallum:
@veehaitch See #1446 for a fix for the test failures.

PR #1455, opened 2022-02-23
build(nix): add `enarx` package

veehaitch:
This PR adds a package `enarx` to the project's flake. It allows building `enarx` with full 
Nix sandboxing and gives strong reproducibility guarantees. Also, it makes it easy to use the `enarx` binary on Nix-enabled systems.

The following command builds the package through Nix and brings the `enarx` binary into scope:

```ShellSession
# nix shell 'github:yaxitech/enarx/nix-build'
...
# enarx info
Version 0.2.1 starting up
Backend: sgx
 ✔ System Info: Linux sgx-cont 5.4.0-99-generic #112-Ubuntu SMP Thu Feb 3 13:50:55 UTC 2022 x86_64 (none)
 ✔ Driver: /dev/sgx_enclave
 ✔ CPU: Intel(R) Core(TM) i5-9400 CPU @ 2.90GHz | GenuineIntel
 ✔  SGX Support
 ✔   Version 1
 ✗   Version 2
 ✔   FLC Support
 ✔   Max Size (32-bit): 2 GiB
 ✔   Max Size (64-bit): 64 GiB
 ✔   MiscSelect: (empty)
 ✔   Features: DEBUG | MODE64BIT | PROVISIONING_KEY | EINIT_KEY
 ✔   Xfrm: X87 | SSE | AVX | YMM | BNDREG | BNDCSR
 ✔   EPC Size: 94 MiB
...
```

The PR also does some refactoring regarding the Nix development shell (`nix develop`).

I'm also happy to extend [the documentation](https://github.com/enarx/enarx.github.io/blob/main/docs/Installation/Enarx.md#install-from-nix) accordingly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,3 +91,4 @@ jobs:
     - uses: cachix/install-nix-action@v16
     - run: nix flake check
     - run: nix develop --ignore-environment -c cargo build
+    - run: nix build -L
--- a/flake.lock
+++ b/flake.lock
@@ -53,6 +53,26 @@
         "type": "github"
       }
     },
+    "naersk": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1639947939,
+        "narHash": "sha256-pGsM8haJadVP80GFq4xhnSpNitYNQpaXk4cnA796Cso=",
+        "owner": "nix-community",
+        "repo": "naersk",
+        "rev": "2fc8ce9d3c025d59fee349c1f80be9785049d653",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "naersk",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1645433236,
@@ -74,6 +94,7 @@
         "fenix": "fenix",
         "flake-compat": "flake-compat",
         "flake-utils": "flake-utils",
+        "naersk": "naersk",
         "nixpkgs": "nixpkgs"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -83,6 +83,9 @@
 
             nativeBuildInputs = [ pkgs.pkg-config ];
             buildInputs = [ pkgs.openssl ];
+
+            cargoTestOptions = x: x ++ [ "-- --skip check_listen_fd" ];
+            doCheck = true;
           };
 
         defaultPackage = self.packages.${system}.enarx;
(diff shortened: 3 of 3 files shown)


rvolosatovs (approved):
Awesome, thank you for contribution!

veehaitch:
I also tried to run the tests in the Nix build. They run fine with one exception: `check_listen_fd`. [The documentation](https://enarx.dev/docs/Installation/Enarx) states that there's a problem with `nix-shell` opening FD 3 and 4. Could this be the reason?

rvolosatovs:
> I also tried to run the tests in the Nix build. They run fine with one exception: `check_listen_fd`. [The documentation](https://enarx.dev/docs/Installation/Enarx) states that there's a problem with `nix-shell` opening FD 3 and 4. Could this be the reason?

Yes, that's exactly the case, therefore we should use `nix develop -c cargo test` instead.
However, we can't run the tests in CI (that's what I originally tried to do in #1441), because we have no backend to run on (i.e. KVM, SGX or SEV). We could do it on a self-hosted runner, but installing Nix there is problematic, because `USER` is not set

veehaitch:
Rebased.

> Yes, that's exactly the case, therefore we should use nix develop -c cargo test instead.
However, we can't run the tests in CI (that's what I originally tried to do in https://github.com/enarx/enarx/pull/1441), because we have no backend to run on (i.e. KVM, SGX or SEV).

I also added a commit to run the Cargo tests during the Nix build. It skips the `check_listen_fd` test for the reasons you outlined, @rvolosatovs.

Even with a runner which has support for either KVM, SGX or SEV, those features probably wouldn't be visible inside a Nix build (for example, the SGX devices are not available in the build's namespace). It still makes sense to run the tests we can.

veehaitch:
The build now checks for KVM support. If it is supported, it only skips the integration test `check_listen_fd`. If KVM is not supported, it skips all the integration tests and only runs the unit tests.

I still wonder if a similar logic for KVM would make sense:

https://github.com/enarx/enarx/blob/500c0b8feaa267b559227694c469d2b1e3b12388/build.rs#L267-L285

rvolosatovs (approved):
LGTM, thank you for contribution!

veehaitch:
While updating the docs I noticed that we also have to add a `default.nix` which leverages `flake-compat` similar to `shell.nix` to allow building the `enarx` package with legacy Nix (i.e., `nix-build`).

I'm not sure if it is still a sensible choice to continue supporting `nix-shell` and `nix-build`, particularly given that the docs mention `cargo test` is broken when using `nix-shell`. Therefore, I'd rather propose to remove support of legacy Nix.

rvolosatovs (approved):

rvolosatovs:
Sorry for late response, had been sick with COVID and still recovering

veehaitch:
Don't worry, get well soon!

veehaitch:
> The test always fails now... at least in my PR #1460

Looks like it's running in a `nix` bug: https://github.com/NixOS/nix/pull/4638

Unfortunately, I don't see an obvious solution to fix the issue. Therefore:

- Disable the CI jobs for `nix flake check` and `nix build .#enarx` until we have a proper solution.
- If `sallyport` `3.0.0` lands on crates.io, the issue goes away but will appear again if someone adds any Git dependency which doesn't have a `master` branch. 

Sorry for the fuss. I didn't expect that.

PR #1456, opened 2022-02-23
test(gh-actions): enable auto-merge on all approved PRs

platten:


npmccallum:
@platten I think we want `test(gh-actions)` since `fix` describes a user-facing fix to be included in the release notes.

platten:
> @platten I think we want `test(gh-actions)` since `fix` describes a user-facing fix to be included in the release notes.

Addressed.

platten:
I think that one more thing that needs to happen here is to make sure that the `pr/auto-merge` action is not run until all of the other checks pass. Also I need to check if this action will fail if auto-merge is already enabled for a particular repository.

PR #1466, opened 2022-02-28
chore(deps): update rust toolchain scripts

platten:


platten:
I rolled these into PR #1469 but keeping this open in case we want to break apart that PR.

PR #1474, opened 2022-03-03
fix(backend): require all checks to pass

haraldh:
And error out with a message to run `enarx info` to prevent users from
getting strange error messages later.

### DIFF
--- a/src/backend/kvm/mod.rs
+++ b/src/backend/kvm/mod.rs
@@ -80,7 +80,7 @@ impl crate::backend::Backend for Backend {
 
     #[inline]
     fn have(&self) -> bool {
-        data::dev_kvm().pass
+        self.data().iter().all(|x| x.pass)
     }
 
     fn data(&self) -> Vec<super::Datum> {
--- a/src/backend/sev/mod.rs
+++ b/src/backend/sev/mod.rs
@@ -55,7 +55,7 @@ impl super::Backend for Backend {
 
     #[inline]
     fn have(&self) -> bool {
-        data::dev_sev_writable().pass
+        self.data().iter().all(|x| x.pass)
     }
 
     fn data(&self) -> Vec<super::Datum> {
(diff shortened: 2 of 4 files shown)


src/backend/kvm/mod.rs
```diff
@@ -80,7 +80,7 @@ impl crate::backend::Backend for Backend {
 
     #[inline]
     fn have(&self) -> bool {
-        data::dev_kvm().pass
+        self.data().iter().fold(true, |acc, x| acc & x.pass)
```
bstrie, line 83: How about `self.data().iter().all(|x| x.pass)`.

src/backend/sev/mod.rs
```diff
@@ -55,7 +55,7 @@ impl super::Backend for Backend {
 
     #[inline]
     fn have(&self) -> bool {
-        data::dev_sev_writable().pass
+        self.data().iter().fold(true, |acc, x| acc & x.pass)
```
bstrie, line 58: Same here.

src/backend/sgx/mod.rs
```diff
@@ -39,7 +39,7 @@ impl crate::backend::Backend for Backend {
 
     #[inline]
     fn have(&self) -> bool {
-        data::dev_sgx_enclave().pass
+        self.data().iter().fold(true, |acc, x| acc & x.pass)
```
bstrie, line 42: Same here.

src/backend/kvm/mod.rs
haraldh: even better .. thanks

haraldh: done

src/backend/sgx/mod.rs
haraldh: done

src/backend/sev/mod.rs
haraldh: done

rvolosatovs (approved):

PR #1475, opened 2022-03-03
fix(helper/test-enarx.sh): require bash

haraldh:

### DIFF
--- a/helper/test-enarx.sh
+++ b/helper/test-enarx.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


bstrie:
What does this fix?

haraldh:
> What does this fix?

@bstrie `${ENARX_BIN[0]}` is not recognized by a posix shell

haraldh:
@bstrie if `/bin/sh` is not a symlink to `/bin/bash` the tests will fail

rvolosatovs (approved):
LGTM, but we really *have* to use bash here?

Issue #1468, opened 2022-03-01
[Bug]: Error: InternalError when running hello world


juhanurmi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

$ enarx run target/wasm32-wasi/release/hello-world.wasm
Error: InternalError

### Expected Behaviour

Runs the default hello world app as the Enarx installation is in place.

### Environment Information

rustc 1.58.1 (db9d1b20b 2022-01-20)

enarx 0.2.0

Linux juha-desktop 5.11.0-49-generic #55-Ubuntu SMP Wed Jan 12 17:36:34 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

juha@juha-desktop:~$ enarx info
Version 0.2.0 starting up
Backend: sgx
 ✔ System Info: Linux juha-desktop 5.11.0-49-generic #55-Ubuntu SMP Wed Jan 12 17:36:34 UTC 2022 x86_64 (none)
 ✗ Driver: /dev/sgx_enclave
 ✔ CPU:        Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz | GenuineIntel
 ✗  SGX Support
 ✔   Version 1
 ✔   Version 2
 ✗   FLC Support
 ✔   Max Size (32-bit): 1 
 ✔   Max Size (64-bit): 1 
 ✗   MiscSelect
 ✔   Features: INIT
 ✔   Xfrm: (empty)
 ✗   EPC Size
Backend: sev
 ✔ System Info: Linux juha-desktop 5.11.0-49-generic #55-Ubuntu SMP Wed Jan 12 17:36:34 UTC 2022 x86_64 (none)
 ✗ Driver: /dev/sev
 ✗  SEV-SNP is enabled in host kernel
 ✗  /dev/sev is readable by user
 ✗  /dev/sev is writable by user
 ✔ Driver: /dev/kvm
 ✔  API Version: 12
 ✔ MEMLOCK rlimit allows for: ~598 keeps (soft limit = 3140073984 bytes, hard limit = 3140073984 bytes)
 ✗ CPU:        Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz | GenuineIntel
 ✗  Microcode support
 ✗  Secure Memory Encryption (SME)
 ✗   Physical address bit reduction
 ✗   C-bit location in page table entry
[...]

PR #1493, opened 2022-03-08
fix(kvm): add the missing CPUID checks

jarkkojs:
Check for RDRAND, FSGSBASE and ADX with the help of raw-cpuid crate.

Closes: #1468

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,7 @@ dependencies = [
  "protobuf",
  "protobuf-codegen-pure",
  "rand",
+ "raw-cpuid",
  "reqwest",
  "sallyport",
  "semver",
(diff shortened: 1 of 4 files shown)


haraldh:
I will go the other round and enable KVM for these old CPUs

jarkkojs:
> I will go the other round and enable KVM for these old CPUs

For KVM it makes sense. SGX1 does not make much.

haraldh:
Here is my counter proposal PR, which fixes things: https://github.com/enarx/enarx/pull/1503
as I already wrote in https://github.com/enarx/enarx/issues/1468#issuecomment-1058119788 that I will do so.

haraldh:
> > I will go the other round and enable KVM for these old CPUs
> 
> For KVM it makes sense. SGX1 does not make much.

We don't support SGX1 ... and I think SGX2 machines all have these CPU features...

haraldh (changes requested):
For SGX we don't need those checks and for KVM, I opened a compat PR with https://github.com/enarx/enarx/pull/1503

npmccallum:
Closing due to merging @haraldh 's alternative approach.

PR #1495, opened 2022-03-08
feat(sgx): EACCEPT{COPY}

jarkkojs:

### DIFF
--- a/internal/shim-sgx/Cargo.lock
+++ b/internal/shim-sgx/Cargo.lock
@@ -277,6 +277,7 @@ dependencies = [
 name = "shim-sgx"
 version = "0.2.1"
 dependencies = [
+ "bitflags",
  "compiler_builtins",
  "const-default",
  "crt0stack",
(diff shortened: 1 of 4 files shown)


npmccallum (changes requested):
This probably belongs in the https://github.com/enarx/sgx crate. Specifically, it probably belongs in this file: https://github.com/enarx/sgx/blob/main/src/page/sinfo.rs.

npmccallum:
@jarkkojs Also note that we have constants for the leaf instructions already: https://docs.rs/sgx/0.3.0/sgx/enclu/index.html

jarkkojs:
> This probably belongs in the https://github.com/enarx/sgx crate. Specifically, it probably belongs in this file: https://github.com/enarx/sgx/blob/main/src/page/sinfo.rs.

What about Report and TargetInfo?

npmccallum:
@jarkkojs Link?

npmccallum:
@jarkkojs Yes. It probably makes sense to move them to `sgx` immediately after the release.

jarkkojs:
https://github.com/enarx/sgx/pull/62

PR #1497, opened 2022-03-09
feat(sgx): mmap()

jarkkojs:
Shim mmap() handler:
1. Ask host for mmap() syscall.
2. Construct secinfo matching the protection bits.
3. For each page in the address range: secinfo.accept_copy() with a zero page.

Shim mprotect() handler:
1. Ask host for mprotect() syscall.
2. For each page in the address range: secinfo.accept() and secinfo.protect().

Backend mprotect() handler:
1. Invoke ENCLAVE_RESTRICT_PERMISSIONS ioctl for the address range with PROT_NONE.
2. Invoke mmap() syscall.

Trimming and removing pages is excluded because it is a part in the kernel patch set that might still need some work (not as throughly reviewed as augmentation and permissions yet).

### DIFF
--- a/internal/shim-sgx/Cargo.toml
+++ b/internal/shim-sgx/Cargo.toml
@@ -19,7 +19,7 @@ gdbstub_arch = { version = "0.1.1" , default-features = false, optional = true }
 gdbstub = { version = "0.5.0" , default-features = false, optional = true }
 goblin = { version = "0.5", default-features = false, features = [ "elf64" ] }
 primordial = { version = "^0.4.0", features = ["const-default"] }
-x86_64 = { version = "^0.14.8", default-features = false }
+x86_64 = { version = "0.14.8-sgx", git = "https://github.com/jarkkojs/x86_64", branch = "feat/sgx" }
 crt0stack = { version = "0.1", default-features = false }
 sallyport = { version = "0.3.0" }
 spinning = { version = "0.1", default-features = false }
@@ -29,6 +29,7 @@ xsave = { version = "2.0.2" }
 rcrt1 = "1.0.0"
 lset = "0.2"
 sgx = "0.3"
+heapless = "0.7.10"
 
 [dev-dependencies]
 memoffset = "0.6.1"
--- a/internal/shim-sgx/Cargo.lock
+++ b/internal/shim-sgx/Cargo.lock
@@ -3,10 +3,44 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "atomic-polyfill"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee6adc1648f03fbc1bc1b5cf0f2fdfb5edbc96215b711edcfe6ce2641ef9b347"
+dependencies = [
+ "critical-section",
+ "riscv-target",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
+name = "bare-metal"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5deb64efa5bd81e31fcd1938615a6d98c82eafcbcd787162b6f63b91d6bac5b3"
+dependencies = [
+ "rustc_version",
+]
+
+[[package]]
+name = "bare-metal"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8fe8f5a8a398345e52358e18ff07cc17a568fbca5c6f73873d3a62056309603"
 
 [[package]]
 name = "bit_field"
@@ -15,10 +49,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dcb6dd1c2376d2e096796e234a70e17e94cc2d5d54ff8ce42b28cef1d0d359a4"
 
 [[package]]
+name = "bitfield"
+version = "0.13.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "46afbd2983a5d5a7bd740ccb198caf5b82f45c40c09c0eed36052d91cb92e719"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "byteorder"
+version = "1.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "cfg-if"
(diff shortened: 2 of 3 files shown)


jarkkojs:
This depends on sgx crate changes and does not compile I just put it available to show a couple of issues to sort out.

jarkkojs:
Depends on: https://github.com/enarx/sgx/pull/62

internal/shim-sgx/Cargo.toml
```diff
@@ -29,7 +29,7 @@ noted = "^1.0.0"
 xsave = { version = "2.0.2" }
 rcrt1 = "1.0.0"
 lset = "0.2"
-sgx = "0.3"
+sgx = { path = "/home/jarkko/Projects/Enarx/sgx" }
```
haraldh, line 32: Please upload your changes to github and point to that commit here, so we can reproduce it.

src/backend/sgx/thread.rs
```diff
@@ -156,6 +168,21 @@ impl super::super::Thread for Thread {
                         return Ok(Command::Continue);
                     }
 
+                    libc::SYS_mprotect => {
```
haraldh, line 171: please rebase to current main

haraldh (changes requested):

internal/shim-sgx/Cargo.toml
jarkkojs: The pull request is here: https://github.com/enarx/sgx/pull/62.

So I guess it is like `sgx = { git = "https://github.com/jarkkojs/sgx/tree/feat/sinfo/eaccept" }`?

haraldh: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

jarkkojs:
@haraldh, I'll make some sub-issues and smaller PR's. First for the backend. Let this be PoC.

haraldh:
please convert to Draft

jarkkojs:
Converted to draft up until dependencies (sgx crate and sallyport) have been sorted out. I learned today about this draft possibility (thanks @haraldh).

jarkkojs:
Current or future kernel changes will not block this because EAUG is hidden to mmap() and #PF flow, i.e. this is almost guaranteed to stay same after SGX2 support reaches the mainline kernel.

jarkkojs:
Temporarily turning this draft as I refactor this a lot.

jarkkojs:
@npmccallum, @haraldh, @rvolosatovs: I'm stuck from all angles to move this forward. Pull requests that block me:

* https://github.com/enarx/primordial/pull/15
* https://github.com/enarx/sgx/pull/73
* https://github.com/enarx/enarx/pull/1561

For enarx/sgx#73 it is already 4th trial. I've done version of this by now for `x86_64`, `sgx` and `enarx` craes. This is 2nd version for `sgx` based on the version I did for `x86_64`. Most of the versions I've done of the versions have received only little attention, expect `x86_64` version, but only from the corresponding maintainer.

I hold on doing Enarx development up until these get merged. I cannot give any promises on landing SGX2 to 0.4 but if these get sorted out fast, then there is a small chance of that happening.

jarkkojs:
I.e. hold on meaning that of course I fix the issues in these but my hands are kind of tied before these get merged.

jarkkojs:
There's probably still enough time to get it done since all the blocks are there in some form, and I actually have quite clear picture in my head of the implementation but e.g. I have not even peaked at `GlobalAlloc` yet because there's so many other moving parts that are not merged. No bandwidth for that. Does it even require a special implementations if all the syscalls are working. Doesn't the default one work then?

jarkkojs:
> There's probably still enough time to get it done since all the blocks are there in some form, and I actually have quite clear picture in my head of the implementation but e.g. I have not even peaked at `GlobalAlloc` yet because there's so many other moving parts that are not merged. No bandwidth for that. Does it even require a special implementations if all the syscalls are working. Doesn't the default one work then?

Right. It does not exist because `no_stdlib`. Perhaps we can use https://crates.io/crates/alloc-no-stdlib for that? That has calloc version, which can then leverage the syscalls. I think that's the easy way out.

jarkkojs:
> > There's probably still enough time to get it done since all the blocks are there in some form, and I actually have quite clear picture in my head of the implementation but e.g. I have not even peaked at `GlobalAlloc` yet because there's so many other moving parts that are not merged. No bandwidth for that. Does it even require a special implementations if all the syscalls are working. Doesn't the default one work then?
> 
> Right. It does not exist because `no_stdlib`. Perhaps we can use https://crates.io/crates/alloc-no-stdlib for that? That has calloc version, which can then leverage the syscalls. I think that's the easy way out.

The Dropbox solution does not look perfect as my SGX2 implementation provides fully dynamic mmap() et al. Is this somewhere: https://rust-lang.github.io/rfcs/2480-liballoc.html? That's exactly what we need. @bstrie

jarkkojs:
I found a solution for heap allocator, which requires only syscalls working:https://crates.io/crates/mmap-alloc. So GlobalAlloc is cleared.

jarkkojs:
Anyway, there's a bunch of quick solutions to deal with heap allocation so it's not a huge problem as long as syscalls land to shim-sgx.

jarkkojs:
This is now feature complete: https://github.com/jarkkojs/mmledger

It has:

- `extend_region()` for `brk()` system call.
- `find_free_space()` for finding free space.
- `insert_region()` for `mmap()`. Limitations: is able to combine adjacent regions into one but overlaps with existing regions are not at this point supported.
- `set_region_permissions()` for `mprotect()`. Limitations: supports only changing exactly matching regions.

The limitations can be relaxed but I guess this is good starting point. There's some ugly aligned checks but I'd prefer to fix them by using `Address<usize, Page>`. That requires enarx/primordial#15 to be merged first. Does not make sense to refine this further before starting to integrate with Enarx. The whole thing is covered by 14 test cases, so it is somewhat working code already.

Now I'm officially in wait state :-) Cannot progress further than this. Every next step has something blocking.

jarkkojs:
Had already gone to sleep but woke up and realized that the structure does not have to track permissions. If there is something wrong in them, `EACCEPT` will crash the enclave. This renders `set_region_permissions()` obsolete.

The one thing that needs to be done is to check that `prot` matches what was left to SSA. @npmccallum: Is there a check in place for SSA so that the parameters match? This ensures that the host does not do any malicious changes. That is obviously important to have.

`mprotect()` is straight-forward as it needs only the syscall and `EACCEPT` + `EMODPE` loop. `mprotect() does not need to consult **any** sort of structure as it does not change the allocation. The only thing it needs to consult is SSA. Even the normal `mprotect()` works in a way that it executes and `SIGSEGV`'s if it goes "out of region".

Thus, I can simplify the API as:

* `allocate_region(size)`
* `extend_region(addr)`
* `insert_region(addr, size)`

Having implemented a working draft for address space, I'm guaranteed that permission tracking is not required.

In the kernel adjacent VMA's can be merged obviously only if their permissions match, which initially led me to thinking that it would be required here. Obviously it is not required. This structure can merge anything adjacent because we only care where there is unallocated area.

I originally had this grazy idea that what if I just use a bitmap where each bit means a reserved page. With 128 kB bitmap you can represent 4 GB address space. Scanning 32 pages in worst case is **insignificant** amount of time compared to everything else happening in the sequence. And it simplifies the operations and makes them super robust.

I first thought that this bitmap idea that I had is way way too over the top, because it was so unconventional, but now I think it is actually quite cool and innovative way to, and it removes almost all the special cases from the code.

I got the idea from 2D graphics engines that I use something like that for collision detection. This is just 1D graphics engine :-)

jarkkojs:
This is also not purely conceptual. I used a bit similar approach in my recent bug fix for the page reclaimer the SGX driver: https://lore.kernel.org/linux-sgx/Yio814lJVgutumd8@iki.fi/T/#t. I.e. I have recent experience implementing something along the lines. Also in this case this prevented having to maintain a tree structure for reserved areas.

jarkkojs:
This version starts to be very near something that should work for us in API level: https://github.com/jarkkojs/mmledger/releases/tag/v0.1.11. I.e. no permissions involved. Simplifies the scheme a lot.

@npmccallum Here I have another SSA question that I asked above (but there was lot of my spam too): how the syscall parameters come to the guest handler? Are they taken from SSA, or at least verified against those in SSA? For mprotect() that should be sufficient and otherwise it does not any checks.

jarkkojs:
Yeah, actually I'm not worried getting land to this. The permission sudoku is a lot easier than  I thought. This is simple interval management and only that. Let the CPU do the hard work with permissions!

PR #1500, opened 2022-03-10
Major wasmldr rework

npmccallum:
cc @haraldh @rvolosatovs


rvolosatovs:
I will test this myself still, but when I tried to implement `WasiFile` trait directly without "extending" an existing implementation, I ran into issues, which I attributed to `wasmtime` performing a downcast on `<impl>.as_any()` and matching to concrete type of the fd. I will add a reference here later. Curious if you found a way to work around that somehow.

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -45,47 +48,29 @@ impl WasiFile for Stream {
...
-    async fn write_vectored<'a>(&mut self, bufs: &[io::IoSlice<'a>]) -> Result<u64, Error> {
-        let (tcp, tls) = &mut *self.0.write().unwrap();
-
-        let n = tls
-            .writer()
-            .write_vectored(bufs)
-            .map_err(|e| Error::io().context(e))
-            .context("could not write plaintext to TLS buffer ciphertext")?;
+    async fn write_vectored<'a>(&mut self, bufs: &[IoSlice<'a>]) -> Result<u64, Error> {
+        let n = self.tls.writer().write_vectored(bufs)?;
```
rvolosatovs, line 70: These need to be wrapped as was done before, because otherwise they will trigger a crash in wasmtime (it crashes on unknown errors). Good to have context as well.

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,146 @@
...
+            self.tls.process_new_packets()?;
+        }
+
+        let n = match self.tls.reader().read_vectored(bufs) {
+            Err(e) if e.kind() == ErrorKind::UnexpectedEof => 0,
+            Err(e) => return Err(e.into()),
+            Ok(n) => n,
+        };
+
+        Ok(n as u64)
```
rvolosatovs, line 66: Why cast and not `.try_into()` conversion as in original draft?

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[wiggle::async_trait]
+impl WasiFile for Listener {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    async fn sock_accept(&mut self, fdflags: FdFlags) -> Result<Box<dyn WasiFile>, Error> {
+        // Accept the connection.
+        let (mut tcp, ..) = self.listener.accept()?;
```
rvolosatovs, line 117: These errors will crash wasmtime if not wrapped

rvolosatovs: See https://github.com/bytecodealliance/wasmtime/blob/acee06aa1641b32b3fe9bb78b29fd20a0a1d20ba/crates/wasi-common/src/error.rs#L3-L24

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: I think the issue is more complex than that. We definitely don't want to crash wasmtime. However, we need to be able to trigger the right wasm return values too. For example, nonblocking doesn't really work in part because we can't bubble up `ErrorKind::WouldBlock` properly. I removed the wrapping because I was concerned that our wrapping would "hide" the appropriate return values from wasmtime.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: We can had `.try_into()` if you want. But all conversions from `usize` to `u64` are infallible on all existing hardware platforms. And I don't see us getting 128bit CPUs anytime soon.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: We really need some application tests to exercise these.

internal/wasmldr/src/loader/compiled/tls.rs
rvolosatovs: But this is an error from rustls, not wasmtime, right?

npmccallum:
This PR is now ready for review. All tests pass!

However, note that there is currently no test for TLS connectivity (though I have tested it manually). Basically, to add TLS tests we have to have some way to communicate the self-generated certs between the tester and testee. But this isn't currently possible. The easiest solution is to have a TLS test that connects to a remote server. But this requires networking to run.

I propose we accept this as technical debt.

internal/wasmldr/src/main.rs
```diff
@@ -67,61 +67,128 @@ use std::os::unix::io::FromRawFd;
...
+    Ok(())
+}
 
-    // Choose an appropriate exit code from our result
-    std::process::exit(match result {
-        Ok(_) => 0,
-        Err(e) => i32::from(e),
-    });
+#[cfg(test)]
+pub(crate) mod test {
```
bstrie, line 125: Does this need to be public?

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+
+impl<T> From<T> for Forgotten<T> {
+    fn from(value: T) -> Self {
+        Self(Some(value))
+    }
+}
+
+impl<T> Drop for Forgotten<T> {
+    fn drop(&mut self) {
+        forget(self.0.take());
```
bstrie, line 40: Is the entire purpose of this type just to leak anything it wraps? What does this achieve? A comment explaining the purpose of this type would be useful.

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+
+impl From<Stream> for Box<dyn WasiFile> {
+    fn from(value: Stream) -> Self {
+        Box::new(value)
+    }
+}
+
+impl Stream {
+    fn new(tcp: CapStream, tls: Connection) -> Self {
+        let cap = unsafe { CapStream::from_raw_fd(tcp.as_raw_fd()) }.into();
```
bstrie, line 57: Safety comment.

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+    async fn set_times(
+        &self,
+        atime: Option<wasi_common::SystemTimeSpec>,
+        mtime: Option<wasi_common::SystemTimeSpec>,
+    ) -> Result<(), Error> {
+        self.any.set_times(atime, mtime).await
+    }
+
+    async fn read_vectored<'a>(&self, bufs: &mut [IoSliceMut<'a>]) -> Result<u64, Error> {
+        let (cap, tls) = &mut *self.lck.write().unwrap();
```
bstrie, line 136: This needs write access, not just read access?

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+
+pub struct Listener {
+    cap: Forgotten<CapListener>,
+    any: AnyListener,
+    cfg: Arc<ServerConfig>,
+}
+
+impl Listener {
+    pub fn new(tcp: cap_std::net::TcpListener, cfg: Arc<ServerConfig>) -> Self {
+        let cap = unsafe { CapListener::from_raw_fd(tcp.as_raw_fd()) }.into();
```
bstrie, line 213: Safety comment.

internal/wasmldr/src/loader/requested.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut body = Vec::new();
+        response.into_reader().read_to_end(&mut body)?;
+
+        // Decode the certificate chain.
+        let path = PkiPath::from_der(&body)?;
+        path.0
+            .iter()
+            .rev()
+            .map(|c| Ok(c.to_vec()?))
+            .collect::<Result<Vec<_>, _>>()
```
bstrie, line 46: Is this turbofish necessary? I would expect it to be inferred from the return type.

internal/wasmldr/src/loader/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub struct Connected {
+    wstore: Store<WasiCtx>,
+    linker: Linker<WasiCtx>,
+}
+
+pub struct Completed {
+    values: Vec<Val>,
+}
```
bstrie, line 60: How would you feel about defining the above structs in their respective files?

bstrie (changes requested):
Looks beautiful. In general it would be nice to have some module-level comments to ease navigation.

internal/wasmldr/src/loader/compiled/null.rs
```diff
@@ -0,0 +1,111 @@
...
+    async fn read_vectored_at<'a>(
+        &self,
+        _bufs: &mut [std::io::IoSliceMut<'a>],
+        _offset: u64,
+    ) -> Result<u64, Error> {
+        Ok(0)
+    }
+
+    async fn write_vectored<'a>(&self, bufs: &[std::io::IoSlice<'a>]) -> Result<u64, Error> {
+        Ok(bufs.iter().map(|b| b.len()).fold(0, |a, x| a + x as u64))
```
rvolosatovs, line 77:
```suggestion
        Ok(bufs.iter().map(|b| b.len()).sum::<usize>() as _)
```

internal/wasmldr/src/loader/compiled/null.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    async fn write_vectored_at<'a>(
+        &self,
+        bufs: &[std::io::IoSlice<'a>],
+        _offset: u64,
+    ) -> Result<u64, Error> {
+        Ok(bufs.iter().map(|b| b.len()).fold(0, |a, x| a + x as u64))
```
rvolosatovs, line 85:
```suggestion
        Ok(bufs.iter().map(|b| b.len()).sum::<usize>() as _)
```

internal/wasmldr/src/loader/compiled/tls.rs
rvolosatovs: Suggestion: https://github.com/enarx/enarx/pull/1499/files#diff-83ecd81920414b13ce57bbe3bd3bfb49a21359a18829b1f57a32f928c6653c4aR213-R214

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        // Create a new TLS connection.
+        let mut tls = Connection::Server(
+            ServerConnection::new(self.cfg.clone())
+                .map_err(|e| Error::io().context(e))
+                .context("could not create new TLS connection")?,
+        );
+
+        // Perform handshake.
+        cap.set_nonblocking(true)?;
```
rvolosatovs, line 243: Shouldn't this come from `fdflags`, why set all streams to nonblocking? at the very least there should be a comment here explaining what's going on and point out the fact that actual requested `fdflags` are going to be set below on the resulting stream

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+        tls.complete_io(&mut cap)
```
rvolosatovs, line 244: Didn't we want to handle this within read and write implementations?
I believe we could just use `complete_io` in those

internal/wasmldr/src/loader/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+pub struct Loader<T>(T);
+
+impl Loader<Acquired> {
+    #[cfg(test)]
```
rvolosatovs, line 65: I wish we could make the fact that this is only used for testing more apparent.
Maybe we should add a comment?

internal/wasmldr/src/loader/pki.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn sign(&self, body: &[u8], algo: AlgorithmIdentifier<'_>) -> Result<Vec<u8>>;
+}
+
+impl<'a> PrivateKeyInfoExt for PrivateKeyInfo<'a> {
+    fn generate(oid: ObjectIdentifier) -> Result<Zeroizing<Vec<u8>>> {
+        let rand = ring::rand::SystemRandom::new();
+
+        let doc = match oid {
+            P256 => {
+                use ring::signature::{EcdsaKeyPair, ECDSA_P256_SHA256_ASN1_SIGNING as ALG};
```
rvolosatovs, line 55: Why not just import `EcdsaKeyPair` at top-level, since it's used in both branches?
Also, not sure there is much benefit from renaming the other import to `ALG`, could also have just been imported at top-level and used here directly with default name

internal/wasmldr/src/loader/pki.rs
```diff
@@ -0,0 +1,110 @@
...
+            (ECPK, Some(P384)) => Ok(ES384),
+            _ => return Err(anyhow!("unsupported")),
+        }
+    }
+
+    fn sign(&self, body: &[u8], algo: AlgorithmIdentifier<'_>) -> Result<Vec<u8>> {
+        let rng = ring::rand::SystemRandom::new();
+        match (self.algorithm.oids()?, algo) {
+            ((ECPK, Some(P256)), ES256) => {
+                use ring::signature::{EcdsaKeyPair, ECDSA_P256_SHA256_ASN1_SIGNING as ALG};
```
rvolosatovs, line 96: Same, duplicate import statement, this could all be at top-level without renaming

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+        let cap = unsafe { CapStream::from_raw_fd(tcp.as_raw_fd()) }.into();
+        let any = AnyStream::from_cap_std(tcp);
+        Self {
+            lck: RwLock::new((cap, tls)),
+            any,
+        }
+    }
+
+    pub fn connect(
+        mut tcp: cap_std::net::TcpStream,
```
rvolosatovs, line 66: This is already imported at top-level as `CapStream`

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+pub struct Listener {
+    cap: Forgotten<CapListener>,
+    any: AnyListener,
+    cfg: Arc<ServerConfig>,
+}
+
+impl Listener {
+    pub fn new(tcp: cap_std::net::TcpListener, cfg: Arc<ServerConfig>) -> Self {
```
rvolosatovs, line 212: Already imported as `CapListener`

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    async fn writable(&self) -> Result<(), Error> {
+        self.any.writable().await
+    }
+}
+
+pub struct Listener {
+    cap: Forgotten<CapListener>,
+    any: AnyListener,
```
rvolosatovs, line 207: Neither the field name, nor import name this is renamed to make it clear what is the purpose of this.
At the very least there should be a comment here.
Also, I'm not `AnyListener` rename helps anything here, IMO just using fully-qualified name would be way clearer here

internal/wasmldr/src/loader/compiled/tls.rs
rvolosatovs: I think previous approach with `tcp` rather than `cap` was clearer

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -0,0 +1,340 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::Any;
+use std::io::{ErrorKind, IoSlice, IoSliceMut, Read, Write};
+use std::mem::forget;
+use std::ops::{Deref, DerefMut};
+use std::sync::{Arc, RwLock};
+
+use cap_std::net::{TcpListener as CapListener, TcpStream as CapStream};
```
rvolosatovs, line 9: This could have just been imported as `TcpListener`(it's the "actual" listener we use) if we avoid importing `wasmtime_wasi::net` primitives at top level. Same for stream

internal/wasmldr/src/loader/requested.rs
```diff
@@ -0,0 +1,157 @@
...
+            TLS13_AES_128_GCM_SHA256,
+            TLS13_CHACHA20_POLY1305_SHA256,
+        ];
+
+        // Set up the server config.
+        let srvcfg = ServerConfig::builder()
+            .with_cipher_suites(cipher_suites)
+            .with_kx_groups(kx_groups)
+            .with_protocol_versions(protocol_versions)?
+            .with_no_client_auth() // TODO: Enable client auth
```
rvolosatovs, line 130: We should file an issue about this and add a reference here

internal/wasmldr/src/loader/compiled/null.rs
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::Any;
+
+use wasi_common::file::{Advice, FdFlags, FileType, Filestat};
+use wasi_common::{Error, ErrorExt, WasiFile};
+
+pub struct Null;
```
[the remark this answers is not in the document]
npmccallum: Yes. That's my plan. I'll file an issue.

tests/wasm/rust-tests/src/check_fd.rs
```diff
@@ -0,0 +1,41 @@
...
+        .expect("Failed to parse FD_COUNT to i32");
+
+    let fd_names = std::env::var("FD_NAMES").expect("No FD_NAMES");
+
+    assert_eq!(fd_names, "stdin:stdout:stderr:LISTEN:CONNECT");
+    assert_eq!(fd_count, 5);
+
+    // Set up the environment sockets.
+    let connect = unsafe { TcpStream::from_raw_fd(4) };
+    let listen = unsafe { TcpListener::from_raw_fd(3) };
```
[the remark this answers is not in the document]
npmccallum: @bstrie We basically did parse `fd_names` on line 22.

internal/wasmldr/src/loader/configured.rs
```diff
@@ -0,0 +1,75 @@
...
+use const_oid::{db::rfc5912::SECP_256_R_1, AssociatedOid, ObjectIdentifier};
+use pkcs8::PrivateKeyInfo;
+use x509::der::{asn1::BitString, Any, Decodable, Encodable};
+use x509::request::{CertReq, CertReqInfo, ExtensionReq};
+use x509::{attr::Attribute, ext::Extension, name::RdnSequence};
+
+impl Loader<Configured> {
+    const KVM: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.6.1.4.1.58270.1.1");
+    //const SGX: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.6.1.4.1.58270.1.2");
+    //const SNP: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.6.1.4.1.58270.1.3");
```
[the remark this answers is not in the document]
npmccallum: They are TODOs which we will do immediately. But they require implementations below. I'm communicating these special reserved numbers to whomever implements this.

internal/wasmldr/src/loader/mod.rs
npmccallum: Defining them in their respective files means that I need to make all the fields `pub(super)`. Both have drawbacks.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: Done. However, I think there are subtle bugs here in nonblocking mode. We shall see...

internal/wasmldr/src/loader/pki.rs
npmccallum: 1. I think Arm uses RSA, which will be a different code flow. We are positioned now for that. Also, this code is copy/paste from steward.

2. The renaming like this prevents awkward line-wrapping.

internal/wasmldr/src/loader/pki.rs
npmccallum: See above.

internal/wasmldr/src/loader/attested.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use std::{io::Read, os::unix::prelude::FromRawFd};
+
+use anyhow::Result;
+
+impl Loader<Attested> {
+    pub fn next(self) -> Result<Loader<Acquired>> {
+        // TODO: get workload from drawbridge.
+
+        // FIXME: temporary hack to read wasm from fd 3.
```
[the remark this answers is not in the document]
npmccallum: In this particular case, the `FIXME` is closely related to the `TODO` above. i.e. we're going to get this workload from the drawbridge. All this has to get refactored in the next release.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: Done.

internal/wasmldr/src/main.rs
npmccallum: Fixed.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: Fixed.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: Fixed.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: I removed all this code.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: Fixed.

internal/wasmldr/src/loader/compiled/null.rs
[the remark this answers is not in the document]
npmccallum: Filed.

internal/wasmldr/src/main.rs
```diff
@@ -67,61 +67,128 @@ use std::os::unix::io::FromRawFd;
...
 
-    let mut reader = if let Some(module) = opts.module {
-        info!("reading module from {:?}", &module);
-        File::open(&module).expect("Unable to open file")
-    } else {
-        // v0.1.0 KEEP-CONFIG HACK: just assume module is on FD3
-        info!("reading module from fd 3");
-        unsafe { File::from_raw_fd(3) }
+        (None, None) => unsafe { File::from_raw_fd(4) },
+        _ => panic!(),
```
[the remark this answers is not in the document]
npmccallum: Fixed.

internal/wasmldr/src/loader/requested.rs
npmccallum: Removed.

internal/wasmldr/src/loader/compiled/null.rs
npmccallum: Fixed.

internal/wasmldr/src/loader/mod.rs
npmccallum: I added a module comment.

bstrie (dismissed):

internal/wasmldr/src/loader/requested.rs
npmccallum: Done.

bstrie (approved):

bstrie (approved):

PR #1515, opened 2022-03-10
feat(sgx): add `enarx sgx register` command

npmccallum:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ default = ["backend-kvm", "backend-sgx", "backend-sev", "wasmldr"]
 
 backend-sev = ["backend-kvm", "ureq"]
 backend-kvm = ["x86_64", "kvm-bindings", "kvm-ioctls"]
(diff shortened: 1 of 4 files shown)


npmccallum:
@rjzak Could you take over this feature? It doesn't actually work yet. You'll have to see the comment from ScottM @ Intel in `chat.enarx.dev` as to why it doesn't work. Basically, we have to extract the correct data from the `bytes` variable and post that instead.

rjzak:
Fix: `.send_bytes(&bytes[8..])?;`. In the chat, Scott said to start with the GUID. I got HTTP status 201 when I did that, and a byte response `1A5B22F205BDD803E690A23FC9FC4E8A`

npmccallum:
@haraldh This is ready for review. I tested it on icelake and it appears to work (the Intel service doesn't return a failure code). This should enable PCK fetching for this system.

haraldh (approved):

PR #1520, opened 2022-03-11
feat(sgx): SecInfoExt

jarkkojs:
```
Add SecInfoExt trait that provides EACCEPT, EACCEPTCOPY and EMODPE
operations for enclave's use.

Closes: #1519
```

### DIFF
--- a/internal/shim-sgx/src/handler/mod.rs
+++ b/internal/shim-sgx/src/handler/mod.rs
@@ -43,6 +43,8 @@ use sallyport::item::syscall::{ARCH_GET_FS, ARCH_GET_GS, ARCH_SET_FS, ARCH_SET_G
 
 use crate::handler::usermem::UserMemScope;
 use crate::{DEBUG, ENARX_EXEC_END, ENARX_EXEC_START, ENCL_SIZE};
+use sgx::enclu::{EACCEPT, EACCEPTCOPY, EMODPE};
+use sgx::page::SecInfo;
 use sgx::ssa::StateSaveArea;
 use x86_64::structures::idt::ExceptionVector;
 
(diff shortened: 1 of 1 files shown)


jarkkojs:
Would it make sense to add `unsafe fn enclu()` to the trait to avoid repetition?

internal/shim-sgx/src/handler/mod.rs
```diff
@@ -71,6 +73,63 @@ impl<'a> Write for Handler<'a> {
     }
 }
 
+trait SecInfoExt {
+    fn accept(&self, dest: *const u8);
+    fn accept_copy(&self, dest: *const u8, src: *const u8);
+    fn protect(&self, dest: *const u8);
```
haraldh, line 79: do these accept any pointer or only page aligned ones?

haraldh: Shouldn't they return the error code from `rax` ??
Best to create consts for them, too.

jarkkojs: Oops, this is not the latest version. In the newest I uses NonNull. I'll update.

internal/shim-sgx/src/handler/mod.rs
```diff
@@ -71,6 +73,63 @@ impl<'a> Write for Handler<'a> {
     }
 }
 
+trait SecInfoExt {
+    fn accept(&self, dest: NonNull<libc::c_void>);
+    fn accept_copy(&self, dest: NonNull<libc::c_void>, src: NonNull<libc::c_void>);
+    fn protect(&self, dest: NonNull<libc::c_void>);
```
haraldh, line 79: Shouldn't they return the error code from rax ??

haraldh (changes requested):

internal/shim-sgx/src/handler/mod.rs
jarkkojs: * EACCEPT, EACCEPTCOPY: yes
* EMODPE: no

jarkkojs: I fixed this issue and pushed a new commit.

jarkkojs:
Could this be merged soon, or is there still something to fix?

haraldh:
what about pointer page alignment? check or not?

jarkkojs:
> what about pointer page alignment? check or not?

I don't know where I messed things up but now the *const u8's are back to NonNull (just did a new push).

So, there could be page alignment check, yes, but does it matter? CPU will assert exception and thread will thrown out if that ever happens.

jarkkojs:
I updated the commit:
- Use NonNull.
- Return usize instead of Result<usize>. SGX error code is a legit execution i.e. there are only legit results from these kind of.

jarkkojs:
New version pushed now the check is:
```rust
    if std::path::Path::new("/dev/sgx_enclave").exists()
        && fs::metadata("/dev/sgx_enclave")
            .unwrap()
            .file_type()
            .is_char_device()
    {
        println!("cargo:rustc-cfg=host_can_test_sgx");

        if std::path::Path::new(AESM_SOCKET).exists()
            && fs::metadata(AESM_SOCKET).unwrap().file_type().is_socket()
        {
        }
    }
```

PR #1527, opened 2022-03-11
feat: enable SGX attestation tests only if AESMD is running

jarkkojs:
```
    feat: enable SGX attestation tests only if AESMD is running
    
    Attestation tests require attestation infrastructure in place, which is
    not always the situation in your developer platform. Having them enabled
    by default weakens the contributor experience because it makes cargo
    test usage tedious. Improve this by disabling attestation tests only if
    AESMD is running. Completely ignore two ioctl tests, as they are
    unstable, and the functionality will migrate to the sgx crate.
    
    Closes: #1524
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,6 +9,7 @@ use walkdir::WalkDir;
 
 const CRATE: &str = env!("CARGO_MANIFEST_DIR");
 const TEST_BINS_IN: &str = "tests/c-tests";
+const AESM_SOCKET: &str = "/var/run/aesmd/aesm.socket";
 
 fn find_files_with_extensions<'a>(
     exts: &'a [&'a str],
@@ -264,13 +265,18 @@ fn main() {
         }
     }
 
-    if std::path::Path::new("/dev/sgx_enclave").exists() {
-        // Not expected to fail, as the file exists.
-        let metadata = fs::metadata("/dev/sgx_enclave").unwrap();
-        let file_type = metadata.file_type();
+    if std::path::Path::new("/dev/sgx_enclave").exists()
+        && fs::metadata("/dev/sgx_enclave")
+            .unwrap()
+            .file_type()
+            .is_char_device()
+    {
+        println!("cargo:rustc-cfg=host_can_test_sgx");
 
-        if file_type.is_char_device() {
-            println!("cargo:rustc-cfg=host_can_test_sgx");
+        if std::path::Path::new(AESM_SOCKET).exists()
+            && fs::metadata(AESM_SOCKET).unwrap().file_type().is_socket()
+        {
+            println!("cargo:rustc-cfg=host_can_test_attestation");
         }
     }
 
@@ -281,6 +287,7 @@ fn main() {
 
         if file_type.is_char_device() {
             println!("cargo:rustc-cfg=host_can_test_sev");
+            println!("cargo:rustc-cfg=host_can_test_attestation");
         }
     }
 }
--- a/src/backend/sgx/attestation.rs
+++ b/src/backend/sgx/attestation.rs
@@ -286,8 +286,8 @@ pub fn get_quote(report: &[u8], akid: Vec<u8>, out_buf: &mut [u8]) -> Result<usi
 mod tests {
     use super::*;
 
+    #[cfg_attr(not(host_can_test_attestation), ignore)]
     #[test]
-    #[cfg_attr(not(host_can_test_sgx), ignore)]
     fn request_target_info() {
         assert_eq!(std::path::Path::new(AESM_SOCKET).exists(), true);
 
--- a/tests/c_integration_tests.rs
+++ b/tests/c_integration_tests.rs
@@ -126,6 +126,7 @@ fn read_udp() {
     run_test("read_udp", 0, input.as_slice(), input.as_slice(), None);
 }
 
+#[cfg_attr(not(host_can_test_attestation), ignore)]
 #[test]
 #[serial]
 fn get_att() {
@@ -133,6 +134,7 @@ fn get_att() {
 }
 
 #[cfg(feature = "backend-sgx")]
+#[cfg_attr(not(host_can_test_attestation), ignore)]
 #[test]
 #[serial]
 fn sgx_get_att_quote() {
@@ -140,6 +142,7 @@ fn sgx_get_att_quote() {
 }
 
 #[cfg(feature = "backend-sgx")]
+#[cfg_attr(not(host_can_test_attestation), ignore)]
 #[test]
 #[serial]
 fn sgx_get_att_quote_size() {


haraldh:
we have can_test_sgx... why not check for the aesmd socket and have can_attest_sgx

jarkkojs:
> we

I think it makes sense to have a separate config flag for attestation test because they can be also non-SGX tests, like "get_att" we have.

jarkkojs:
> > we have can_test_sgx... why not check for the aesmd socket and have can_attest_sgx
> 
> I think it makes sense to have a separate config flag for attestation test because they can be also non-SGX tests, like "get_att" we have. Given all attestation needs CA etc. infrastructure **you know** when you want attestation testing happening.

Maybe use the same naming convention as for `host_can_test_sgx`, and call this `host_want_test_attestation` instead of `test_att`? For two SGX specific tests I could use then logical conjunction:

```
#[cfg(all(feature = "backend-sgx", host_can_test_sgx, host_want_test_attestation))]
#[test]
#[serial]
fn sgx_get_att_quote_size() {
    run_test("sgx_get_att_quote_size", 0, None, None, None);
}
```

jarkkojs:
> > > we have can_test_sgx... why not check for the aesmd socket and have can_attest_sgx
> > 
> > 
> > I think it makes sense to have a separate config flag for attestation test because they can be also non-SGX tests, like "get_att" we have. Given all attestation needs CA etc. infrastructure **you know** when you want attestation testing happening.
> 
> Maybe use the same naming convention as for `host_can_test_sgx`, and call this `host_want_test_attestation` instead of `test_att`? For two SGX specific tests I could use then logical conjunction:
> 
> ```rust
> #[cfg(all(feature = "backend-sgx", host_can_test_sgx, host_want_test_attestation))]
> #[test]
> #[serial]
> fn sgx_get_att_quote_size() {
>     run_test("sgx_get_att_quote_size", 0, None, None, None);
> }
> ```

On 2nd thought, I'll actually do what you proposed :-)

jarkkojs:
@haraldh, I took an angle more aligned with what you suggested. It's now pushed.

build.rs
```diff
@@ -264,13 +265,28 @@ fn main() {
...
 
-    if std::path::Path::new("/dev/sgx_enclave").exists() {
+    let host_can_test_sgx = if std::path::Path::new("/dev/sgx_enclave").exists() {
         // Not expected to fail, as the file exists.
         let metadata = fs::metadata("/dev/sgx_enclave").unwrap();
         let file_type = metadata.file_type();
 
         if file_type.is_char_device() {
             println!("cargo:rustc-cfg=host_can_test_sgx");
+            true
```
bstrie, line 275: Can probably avoid the ifs and elses by replacing this `true` with `if file_type.is_socket() && std::path::Path::new(AESM_SOCKET).exists() {`

src/backend/sgx/ioctls.rs
```diff
@@ -192,6 +192,7 @@ mod tests {
     use super::*;
 
     #[test]
+    #[ignore]
```
bstrie, line 195: Why are these being ignored rather than using the cfg approach?

bstrie:
Seems reasonable. You've tried this locally and this suffices to prevent you from needing to run the attestation tests? I just want to confirm that cfg flags set by build.rs are visible to the test runner.

jarkkojs:
> Seems reasonable. You've tried this locally and this suffices to prevent you from needing to run the attestation tests? I just want to confirm that cfg flags set by build.rs are visible to the test runner.

Yes because I did this to enable local use :-)

src/backend/sgx/ioctls.rs
jarkkojs: I removed these changes from this patch. `#[tests]` is prepended with `host_can_test_sgx`.

build.rs
jarkkojs: There was actually a bug. I checked file_type() for the device file, not socket.

jarkkojs: This should be correct now:
```rust
    let host_can_test_sgx = std::path::Path::new("/dev/sgx_enclave").exists()
        && fs::metadata("/dev/sgx_enclave")
            .unwrap()
            .file_type()
            .is_char_device();

    if host_can_test_sgx {
        println!("cargo:rustc-cfg=host_can_test_sgx");

        if std::path::Path::new(AESM_SOCKET).exists()
            && fs::metadata(AESM_SOCKET).unwrap().file_type().is_socket()
        {
        }
    }

 ```

**Comment by jarkkojs** [comment]
So the check has formed now into:
```rust
    if std::path::Path::new("/dev/sgx_enclave").exists()
        && fs::metadata("/dev/sgx_enclave")
            .unwrap()
            .file_type()
            .is_char_device()
    {
        println!("cargo:rustc-cfg=host_can_test_sgx");

        if std::path::Path::new(AESM_SOCKET).exists()
            && fs::metadata(AESM_SOCKET).unwrap().file_type().is_socket()
        {
        }
    }
```

bstrie (approved):

bstrie:
Have you confirmed that the tests do run properly on machines where they can?

haraldh:
I would prefer `ignore` for the skipped attestation tests.

jarkkojs:
> Have you confirmed that the tests do run properly on machines where they can?

@haraldh Doesn't CI do attestation tests? I did test locally that non-attestation is working.

jarkkojs:
> I would prefer `ignore` for the skipped attestation tests.

Put ignore in front of them and checks?

haraldh:
```
#[cfg_attr(host_can_attest, ignore)]
```

haraldh:
@jarkkojs this... before `#[test]`

jarkkojs:
> @jarkkojs this... before `#[test]`

OK, I hope I got right.

haraldh:
```
test sgx_get_att_quote ... ignored 

test sgx_get_att_quote_size ... ignored
```

hmmm

haraldh:
in my first version of my comment, I forgot the `not()`

jarkkojs:
> in my first version of my comment, I forgot the `not()`

OK I was puzzled by this :-)

jarkkojs:
> 

```
test get_att ... ignored
test bind ... ok
test close ... ok
test clock_gettime ... ok
test exit_one ... ok
test exit_zero ... ok
test getegid ... ok
test sgx_get_att_quote ... ignored
test sgx_get_att_quote_size ... ignored
```
As expected now in my laptop.

jarkkojs:
@haraldh OK now they should be good.

haraldh (approved):

haraldh:
your enotty slipped in...totally unrelated

haraldh (changes requested):
remove the enotty changes

jarkkojs:
Done.

jarkkojs:
> your enotty slipped in...totally unrelated

was a mistake

haraldh:
> > your enotty slipped in...totally unrelated
> 
> was a mistake

you are testing me, aren't you? 😀

haraldh (approved):

jarkkojs:
> > > your enotty slipped in...totally unrelated
> > 
> > 
> > was a mistake
> 
> you are testing me, aren't you? grinning

I'm switching between my desktop and laptop and it can get confusing sometimes. Thanks for approving!

PR #1537, opened 2022-03-14
refactor(shims): replace libc crate with vendored libc

bstrie:
As a step towards porting the shims to x84_64-unknown-none, remove the libc dependency in favor of the vendored libc definitions exported by sallyport.

### DIFF
--- a/internal/shim-kvm/src/hostcall.rs
+++ b/internal/shim-kvm/src/hostcall.rs
@@ -13,17 +13,20 @@ use crate::snp::{cpuid, snp_active};
 use crate::spin::{RacyCell, RwLocked};
 
 use const_default::ConstDefault;
-use core::ffi::c_void;
+use core::ffi::{c_int, c_size_t, c_uint, c_ulong, c_void};
 use core::mem::size_of;
 use core::ops::{Deref, DerefMut};
 use core::ptr::NonNull;
 use core::slice;
 
-use libc::{c_int, c_uint, c_ulong, off_t, size_t, EINVAL, GRND_NONBLOCK, GRND_RANDOM};
 use sallyport::guest::syscall::Getrandom;
 use sallyport::guest::{self, Handler, Platform, ThreadLocalStorage};
 use sallyport::item::enarxcall::sev::TECH;
 use sallyport::item::syscall;
+use sallyport::libc::{
+    off_t, EAGAIN, EFAULT, EINVAL, EIO, EMSGSIZE, ENOMEM, GRND_NONBLOCK, GRND_RANDOM,
+    MAP_ANONYMOUS, MAP_PRIVATE, PROT_EXEC, PROT_WRITE,
+};
 use sallyport::util::ptr::is_aligned_non_null;
 use sallyport::KVM_SYSCALL_TRIGGER_PORT;
 use spinning::Lazy;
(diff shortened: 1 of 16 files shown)


internal/shim-sgx/src/handler/mod.rs
```diff
@@ -33,13 +33,13 @@ use core::fmt::Write;
...
 use core::ptr::read_unaligned;
 use core::ptr::NonNull;
-use libc::{c_int, c_ulong, off_t, size_t};
 
 use sallyport::guest::Handler as _;
 use sallyport::guest::{self, Platform, ThreadLocalStorage};
 use sallyport::item::enarxcall::sgx::{Report, ReportData, TargetInfo, TECH};
 use sallyport::item::enarxcall::SYS_GETATT;
 use sallyport::item::syscall::{ARCH_GET_FS, ARCH_GET_GS, ARCH_SET_FS, ARCH_SET_GS};
+use sallyport::libc::{self, c_int, c_ulong, off_t, size_t};
```
rvolosatovs, line 42: Why do we not just use the `core_ffi_c` feature (which we use in sallyport anyway, tracking issue https://github.com/rust-lang/rust/issues/94501) and definitions from `core::ffi` here?

bstrie: We could, it doesn't make a difference, they're just type aliases. I just made this patch as minimally-invasive as possible.

internal/shim-kvm/src/hostcall.rs
```diff
@@ -519,17 +519,17 @@ impl Platform for UserMemScope {
...
     /// and registers the memory as borrowed mutably.
     ///
-    /// Returns a mutable borrow if valid, otherwise [`EINVAL`](libc::EINVAL).
+    /// Returns a mutable borrow if valid, otherwise [`EINVAL`](https://man7.org/linux/man-pages/man3/errno.3.html).
     ///
     /// [the ptr module documentation]: core::ptr#safety
     #[inline]
     fn validate_mut<T>(&self, ptr: usize) -> Result<&mut T, c_int> {
-        is_aligned_non_null::<T>(ptr).ok_or(EINVAL)?;
+        is_aligned_non_null::<T>(ptr).ok_or(libc::EINVAL)?;
```
rvolosatovs, line 527: Why this change? As a general rule, I think we should bring these kind of constants (e.g. errno or flags) into scope for clarity.

internal/shim-kvm/src/hostcall.rs
```diff
@@ -492,9 +492,9 @@ impl Handler for HostCall<'_> {
         Ok(())
     }
 
-    fn getrandom(&mut self, buf: &mut [u8], flags: c_uint) -> sallyport::Result<size_t> {
-        if flags & !(GRND_NONBLOCK | GRND_RANDOM) != 0 {
-            return Err(EINVAL);
+    fn getrandom(&mut self, buf: &mut [u8], flags: c_uint) -> sallyport::Result<c_size_t> {
+        if flags & !(libc::GRND_NONBLOCK | libc::GRND_RANDOM) != 0 {
+            return Err(libc::EINVAL);
```
npmccallum, line 497: Same issue as @rvolosatovs noticed here.

npmccallum (changes requested):

internal/shim-kvm/src/hostcall.rs
bstrie: I decided whether or not to bring things into scope based on how many items were being used and how many times each were being used. I don't mind bringing them all into scope if we have a blanket policy for that, though sometimes it might result in very long `use` statements.

internal/shim-kvm/src/hostcall.rs
bstrie: I've favored a policy of bringing these items into scope.

rvolosatovs (approved):

PR #1551, opened 2022-03-15
fix(wasmldr): make examples/tcp_server work with TLS

npmccallum:

### DIFF
--- a/internal/wasmldr/src/loader/compiled/tls.rs
+++ b/internal/wasmldr/src/loader/compiled/tls.rs
@@ -2,7 +2,7 @@
 //! A WasiFile for transparent TLS
 
 use std::any::Any;
-use std::io::{ErrorKind, IoSlice, IoSliceMut, Read, Write};
+use std::io::{IoSlice, IoSliceMut, Read, Write};
 use std::mem::forget;
 use std::ops::{Deref, DerefMut};
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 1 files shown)


internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -83,6 +95,17 @@ impl Stream {
 
         Ok(Self::new(tcp, tls))
     }
+
+    fn complete_io(&self) -> Result<(), Error> {
+        let (cap, tls) = &mut *self.lck.write().unwrap();
+
+        while tls.wants_read() || tls.wants_write() {
```
rvolosatovs, line 102: `complete_io` already does this check internally

internal/wasmldr/src/loader/compiled/tls.rs
```diff
@@ -83,6 +95,17 @@ impl Stream {
...
+            tls.complete_io(cap.deref_mut()).map_err(errmap)?;
+            tls.process_new_packets()?;
```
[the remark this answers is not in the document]
npmccallum: No. It isn't `std::io::Error`.

internal/wasmldr/src/loader/compiled/tls.rs
npmccallum: It only does one read. See the docs.

rvolosatovs (approved):

Issue #1558, opened 2022-03-17
[Bug]: shim-sgx: arch_prctl() does nothing


jarkkojs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

Looking at `arch_prctl()` handler:

```rust
            ARCH_SET_FS => self.ssa.gpr.fsbase = addr,
            ARCH_SET_GS => self.ssa.gpr.gsbase = addr,
````

SSA gets overwritten on AEX and the values come from those in TCS. FS can be only changed with WRFSBASE. For GS we should never do nothing special since it is irrelevant for 64-bit.

haraldh:
fixed now, isn't it?

haraldh:
with #1665

PR #1559, opened 2022-03-17
fix(shim-sgx): update FSGSBASE in arch_prctl() correctly

jarkkojs:
Closes: #1558

### DIFF
--- a/internal/shim-sgx/Cargo.toml
+++ b/internal/shim-sgx/Cargo.toml
@@ -19,7 +19,7 @@ gdbstub_arch = { version = "0.1.1" , default-features = false, optional = true }
 gdbstub = { version = "0.5.0" , default-features = false, optional = true }
 goblin = { version = "0.5", default-features = false, features = [ "elf64" ] }
 primordial = { version = "^0.4.0", features = ["const-default"] }
-x86_64 = { version = "^0.14.8", default-features = false }
+x86_64 = { version = "^0.14.8", default-features = true }
 crt0stack = { version = "0.1", default-features = false }
 sallyport = { version = "0.3.0" }
 spinning = { version = "0.1", default-features = false }
--- a/internal/shim-sgx/src/handler/mod.rs
+++ b/internal/shim-sgx/src/handler/mod.rs
@@ -45,6 +45,9 @@ use crate::handler::usermem::UserMemScope;
 use crate::{DEBUG, ENARX_EXEC_END, ENARX_EXEC_START, ENCL_SIZE};
 use sgx::ssa::StateSaveArea;
 use sgx::ssa::Vector;
+use x86_64::instructions::segmentation::Segment64;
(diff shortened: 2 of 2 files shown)


haraldh (changes requested):
This is not correct and sets FS and GS in the shim context.

jarkkojs:
> This is not correct and sets FS and GS in the shim context.

The existing code does not set them, they're unconditionally taken from TCS.

jarkkojs:
> > This is not correct and sets FS and GS in the shim context.
> 
> The existing code does not set them, they're unconditionally taken from TCS.

Applies to both EENTER and ERESUME:

```
TMP_FSBASE := (DS:RBX).OFSBASE + TMP_SECS.BASEADDR;
TMP_FSLIMIT := (DS:RBX).OFSBASE + TMP_SECS.BASEADDR + (DS:RBX).FSLIMIT;
TMP_GSBASE := (DS:RBX).OGSBASE + TMP_SECS.BASEADDR;
TMP_GSLIMIT := (DS:RBX).OGSBASE + TMP_SECS.BASEADDR + (DS:RBX).GSLIMIT;
```

haraldh:
I'll leave it to @npmccallum to explain again all the SSA level changes happening on a syscall.

jarkkojs:
> I'll leave it to @npmccallum to explain again all the SSA level changes happening on a syscall.

Happy to hear that, since I did this PR to poke a bit before I start pushing anything larger. I was planning to limit this further to allow setting only to the dynamically allocated range, so in that sense this is really important to understand.

jarkkojs:
@npmccallum For this do not eagerly want to get this merged, but I do want to know what is wrong, because I don't understand :-) It's one way to formalize a question kind of.

npmccallum:
@jarkkojs  the handler is operating in CSSA=1. The ssa reference here is to CCCS=0. We can modify it and then when we do ERESUME, the CPU restores its context to the new values we put into the SSA.

jarkkojs:
@npmccallum I might be wrong but looking at pseudo-code from SDM, fsbase and gsbase are always pulled from TCS. You can write other values to SSA but I don't think they have effect. That's why there was also push from SGX side to get WRFSBASE et al supported in the kernel.

jarkkojs:
From `ERESUME` description in the SDM: _"new values are constructed using TCS.OFSBASE/GSBASE (32 and 64-bit mode) and TCS.OFSLIMIT/GSLIMIT (32-bit mode only)."_

npmccallum:
I'm open to my implementation being faulty. However, I think that means we need a musl libc that supports WRFSBASE. Is that correct?

jarkkojs:
> I'm open to my implementation being faulty. However, I think that means we need a musl libc that supports WRFSBASE. Is that correct?

See e.g. https://github.com/occlum/occlum/issues/39. So I guess the answer is yes.

PR #1564, opened 2022-03-19
build(nix): use `buildRustPackage` instead of `naersk`

veehaitch:
- Use native `nixpkgs` tooling to produce `x86_64-unknown-linux-musl` artifacts
- Remove `naersk`
- Re-add the toolchain hash to allow pure evaluation/building (aee24c73f79d12a9818d434af512d5087d868b78)
- Run all tests when building (`nix build`)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,10 +7,8 @@
   inputs.nixpkgs.url = github:NixOS/nixpkgs/nixos-unstable;
   inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
   inputs.fenix.url = github:rvolosatovs/fenix?ref=fix/rustc-patch;
-  inputs.naersk.url = github:nix-community/naersk;
-  inputs.naersk.inputs.nixpkgs.follows = "nixpkgs";
 
-  outputs = { self, nixpkgs, fenix, flake-utils, naersk, ... }:
+  outputs = { self, nixpkgs, fenix, flake-utils, ... }:
     # NOTE: musl is only supported on Linux.
     with flake-utils.lib; eachSystem [ system.x86_64-linux ] (system:
       let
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -7,17 +7,16 @@
...
-  outputs = { self, nixpkgs, fenix, flake-utils, naersk, ... }:
+  outputs = { self, nixpkgs, fenix, flake-utils, ... }:
     # NOTE: musl is only supported on Linux.
     with flake-utils.lib; eachSystem [ system.x86_64-linux ] (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
 
         rust = fenix.packages."${system}".fromToolchainFile {
           file = "${self}/rust-toolchain.toml";
+          sha256 = "sha256-CSZsqsnfDEL4pfyKRvWU/opdMjvx6vnJwFnkJXL/oBI=";
```
rvolosatovs, line 19: The maintenance cost of keeping this up-to-date is too high (i.e. to update the pinned nightly version, this would be an extra step, which requires a `nix` dependency), so let's keep it unset.

rvolosatovs: Maybe we should just have a separate toolchain with the hash pinned and used for builds and have a "dynamic" toolchain to use for the dev shell, what do you think?

veehaitch: Thanks for all the effort!

I totally agree with your assessment; the `fenix` IFD approach is cumbersome to maintain. [`oxalica`](https://github.com/oxalica/rust-overlay) is easier to handle but for whatever reason doesn't work with the `devShell`. Building the package with `nix build` works fine with `oxalica`.

For now, going to mark this as draft as #1571 complicates the Nix build.

platten:
@rvolosatovs @veehaitch is there any progress on this?

veehaitch:
Unfortunately, no. I'm closing this until we come up with a viable solution.

PR #1581, opened 2022-03-22
feat: SNP attestation with VCEK key

haraldh:
Chain the SNP report with the vcek in ASN.1 format, in a format the steward can use right away.

For simplicity the length fields are all u32 encoded, regardless of the actual size.

See also `tests/sev_attestation/src/main.rs`.

Fixes: https://github.com/enarx/sallyport/issues/136

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -842,8 +842,7 @@ checksum = "73b4b750c782965c211b42f022f59af1fbceabdd026623714f104152f1ec149f"
 [[package]]
 name = "sallyport"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f264dba3626016c14186361e82a499d3274d4580305c60e399fd07623af27404"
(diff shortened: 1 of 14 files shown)


npmccallum:
@haraldh Note, you can find a valid instance of the `Evidence` type [here](https://github.com/profianinc/steward/blob/main/src/main.rs#L387-L391). You can encode it to bytes using the `.to_vec()` method. Then you can parse the bytes [here](https://lapo.it/asn1js/) to see which bytes you need to make dynamic.

haraldh:
Got the basics ready. Will post an updated PR tomorrow

npmccallum (changes requested):
@haraldh You are encoding one too many layers here. You are encoding the `Extension` type when you should be encoding the inner `Evidence` type. The wasmldr code will wrap the encoded `Evidence` in an `Extension`.

haraldh:
> @haraldh You are encoding one too many layers here. You are encoding the `Extension` type when you should be encoding the inner `Evidence` type. The wasmldr code will wrap the encoded `Evidence` in an `Extension`.

@npmccallum fixed

npmccallum (approved):

PR #1582, opened 2022-03-22
feat(sev): add vcek caching

haraldh:
Removes the old `sev` sub-command and adds:

```
$ enarx snp vcek --help
SNP VCEK related commands

USAGE:
    enarx snp vcek <SUBCOMMAND>

FLAGS:

SUBCOMMANDS:
    help         Prints this message or the help of the given subcommand(s)
    show         Print the VCEK certificate for this platform to stdout in PEM format
    show-file    Print the VCEK certificate cache file used
    update       Download the VCEK certificate for this platform and save it to a cache file in `/var/cache/amd-sev/` or `$XDG_CACHE_HOME` or `$HOME/.cache/`
```

`enarx snp vcek update` is to be used in system boot scripts to initialize the cache file once.
    
The searched directories are `/var/cache` and then `$XDG_CACHE_HOME` or `$HOME/.cache`.

The searched entry is `$DIR/amd-sev/vcek-${unique vcek identifier}`.
    
The `update` and `show` command will create a cache file with the vcek, if it does not yet exist.
    
The `show-file` prints the cache file used by `show` to stdout.


rvolosatovs (commented):
Overall implementation looks good, but why separate sub-command?
I think we should use the cache on each `vcek` invocation by default, i.e.:
1. Check the cache, if the VCEK is already cached - print to stdout, otherwise 2.
2. Perform a `GET`, store the downloaded VCEK in cache on success
3. Print the downloaded VCEK to stdout
We could also add a `--force` or `--force-download` flag to the `vcek` subcommand to bypass the cache and always issue a GET request updating the cache on success

Additionally, this lacks testing - as complexity grows here, at least some basic unit tests testing the URLs and paths generated would be very helpful

rjzak:
1.  Maybe have an option to cache the key in the user's home directory, like `~/.enarx/` or similar if the user doesn't have permission to write in `/var/cache/amd-sev/`.
2. Maybe have the cache directory something like `/var/cache/enarx/amd-sev/` so it's clear to the system owner why it's there?

rvolosatovs:
Also, is `XDG_CACHE_HOME` a better choice here?

haraldh:
oh right.. fallback to `.local/cache`

haraldh:
`/var/cache/amd-sev` already exists

src/backend/sev/snp/firmware/mod.rs
```diff
@@ -57,6 +57,14 @@ bitflags! {
 pub struct Identifier(Vec<u8>);
 
 impl Identifier {
+    // Get a unique identifier for the VCEK key to be used in linux file names
+    pub fn vcek_cache_path(&self, version: &TcbVersion) -> String {
+        format!(
+            "/var/cache/amd-sev/Milan-{:x}-{:02}-{:02}-{:02}-{:02}",
```
npmccallum, line 63: We can't hard code `Milan` here.

haraldh: @npmccallum  It's already hardcoded in the URL

npmccallum: @haraldh What happens when Genoa is released?

npmccallum: @haraldh My understanding is that we have invented this filename format. I'm not sure why the generation name is even needed here. Is it to avoid collisions after a CPU swap?

npmccallum (changes requested):
I think we should have two commands:
1.  `enarx snp vcek update`
2. `enarx snp vcek show`

The first command should download the VCEK and write it to whatever cache location it has permissions for: the system cache first, then user cache as a fallback. If a file already exists for the bootloader/tee/snp/microcode, do nothing and exit with success. This command can be put in a systemd unit to regularly update the VCEK on disk as things like the microcode change, for example.

The second command should only show the contents of the cache. It should not do any downloading and should error if it is invoked and the cache is missing the current VCEK. This command should show information about the current certificate for the CPU on disk. It should have an option (maybe `--raw` or `--certificate`) to dump the raw DER-encoded certificate to stdout.

src/backend/sev/snp/firmware/mod.rs
haraldh: Prefixed with `vcek-` now.

haraldh:
Ok, you asked for the over engineered version. Here it is. With locking and concurrent test threads.

src/cli/snp.rs
```diff
@@ -0,0 +1,260 @@
...
+            .context(format!("Error getting vcek from URL {}", &url))?;
+        Ok(Box::new(call.into_reader()))
+    })
+}
+
+// get a cached file, or write the file with the reader provided by the contents function
+fn get_or_write<F>(
+    paths: Vec<PathBuf>,
+    name: String,
+    contents: F,
```
rvolosatovs, line 119: `impl Fn() -> Result<Box<dyn Read>>` maybe?

rvolosatovs (commented):
Overall, LGTM, but I'll let @npmccallum to comment on the expected behavior here.
Seems to differ from https://github.com/enarx/enarx/pull/1582#pullrequestreview-918654910

src/cli/snp.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// SNP VCEK related commands
+    Vcek(VcekCommand),
+}
+
+#[derive(StructOpt, Debug)]
+pub enum VcekCommand {
+    /// Print the VCEK certificate for this platform to stdout in PEM format
+    Show,
+    /// Print the VCEK certificate cache file used
+    ShowFile,
```
rvolosatovs, line 24: Define "used" - it seems that we may use different paths on download and on query

src/cli/snp.rs
```diff
@@ -0,0 +1,260 @@
...
+
+pub fn run(cmd: Command) -> Result<()> {
+    match cmd {
+        Command::Vcek(VcekCommand::Show) => {
+            let mut reader = get_vcek_reader()?;
+            io::copy(&mut reader, &mut io::stdout())?;
+            Ok(())
+        }
+        Command::Vcek(VcekCommand::ShowFile) => {
+            let (path, _) = get_vcek_reader_with_paths(paths())?;
```
rvolosatovs, line 38: Does this download and write to fs if file is not present?

haraldh: fixed: not anymore.

src/cli/snp.rs
haraldh: So, `update` might update another file in the hierarchy first, but _only_ if none of the others exist.

src/cli/snp.rs
haraldh: fixed

src/cli/snp.rs
```diff
@@ -0,0 +1,260 @@
...
+
+/// Returns the list of search paths in order that they
+/// will be searched for the VCEK.
+fn paths() -> Vec<PathBuf> {
+    vec![home(), sys()].into_iter().flatten().collect()
+}
+
+/// Return a reader, which provides the VCEK
+pub fn get_vcek_reader() -> Result<Box<dyn Read>> {
+    get_vcek_reader_with_paths(paths()).map(|(_, r)| r)
```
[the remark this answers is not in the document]
haraldh: fixed

src/cli/snp.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// append the `amd-sev` subdir
+fn append_rest(mut path: PathBuf) -> PathBuf {
+    path.push("amd-sev");
+    path
+}
+
+/// Returns the "user-level" search path for the SEV
+/// certificate chain (`$HOME/.cache/amd-sev`).
```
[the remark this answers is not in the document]
haraldh: fixed

haraldh:
> Overall, LGTM, but I'll let @npmccallum to comment on the expected behavior here. Seems to differ from [#1582 (review)](https://github.com/enarx/enarx/pull/1582#pullrequestreview-918654910)

changed the behaviour

src/cli/snp.rs
haraldh: clarified, that it prints the file used by the `show` command

npmccallum:
@rvolosatovs Let me know when you're happy and I'll approve.

rvolosatovs (approved):
Let's not spend more time on this and merge as-is, but the code could benefit from some refactoring later to e.g. reduce nesting

haraldh:
@rvolosatovs checks didn't run on the last PR update... please re-approve

rvolosatovs (approved):

PR #1595, opened 2022-03-27
feat(shim-sgx): enable `mmledger::Ledger`

jarkkojs:
feat(shim-sgx): maintain `Heap` state with `mmledger::Ledger` 

### DIFF
--- a/internal/shim-sgx/Cargo.toml
+++ b/internal/shim-sgx/Cargo.toml
@@ -17,7 +17,7 @@ dbg = []
 gdbstub_arch = { version = "0.1.1" , default-features = false, optional = true }
 gdbstub = { version = "0.5.0" , default-features = false, optional = true }
 goblin = { version = "0.5", default-features = false, features = [ "elf64" ] }
-primordial = { version = "^0.4.0", features = ["const-default"] }
+primordial = { git = "https://github.com/enarx/primordial", rev = "b606b9e472e2238384b003f12fdada077e04d2fe", features = ["const-default"] }
 x86_64 = { version = "^0.14.8", default-features = false }
 crt0stack = { version = "0.1", default-features = false }
 sallyport = { version = "0.3.0", git = "https://github.com/enarx/sallyport", rev = "580480b19b1542bc3da2fdafef08f895f5bb3389" }
@@ -26,8 +26,9 @@ const-default = "1.0"
 noted = "^1.0.0"
 xsave = { version = "2.0.2" }
 rcrt1 = "1.0.0"
-lset = "0.2"
+lset = { git = "https://github.com/enarx/lset", rev = "5c2feed24d85442725f55d812315854e1b29595d"  }
 sgx = "0.3"
+mmledger = { git = "https://github.com/jarkkojs/mmledger", rev = "2b3e55024cc13c8c14b76efa2f61bbae0905f6eb" }
 
 [dev-dependencies]
 memoffset = "0.6.1"
--- a/internal/shim-sgx/src/heap.rs
+++ b/internal/shim-sgx/src/heap.rs
@@ -2,12 +2,14 @@
 
 //! Allocate and deallocate memory on a Heap
 
+use core::cmp::Ordering;
 use core::ffi::{c_int, c_size_t};
 use core::num::NonZeroUsize;
 use core::ops::Range;
 
 use const_default::ConstDefault;
-use primordial::Page;
+use mmledger::{Access, Ledger, Region};
+use primordial::{Address, Offset, Page};
 use sallyport::libc::{
     off_t, EINVAL, ENOMEM, MAP_ANONYMOUS, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE,
 };
@@ -17,9 +19,9 @@ use sallyport::libc::{
 static mut BLOCK: Block<32768> = Block::new();
 
 /// The keep heap
-pub static HEAP: spinning::RwLock<Heap<'_, 32768>> =
+pub static HEAP: spinning::RwLock<Heap<'_>> =
     spinning::RwLock::const_new(spinning::RawRwLock::const_new(), unsafe {
-        Heap::new(&mut BLOCK)
+        Heap::new(&mut BLOCK.0)
     });
 
 /// An allocated block of memory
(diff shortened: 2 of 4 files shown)


jarkkojs:
Do not expect tests pass yet. Just put it visible.

jarkkojs:
> Do not expect tests pass yet. Just put it visible.

Fixing the reported issues.

jarkkojs:
It could be that `heap::tests::mmap_munmap_oneshot` is failing because of `mmledger` bugs.

For reference:

```
cargo test -- --exact test heap::tests::mmap_munmap_oneshot
```

@npmccallum

jarkkojs:
@npmccallum, please check my suggestions here: https://github.com/enarx/mmledger/issues/25

jarkkojs:
I think this example from the heap implementation goes for a rationale.

Now:

```rust
    fn allocate(&mut self, page: usize) {
        let region = Region::new(self.as_address(page), self.as_address(page + 1));

        let record = Record {
            region,
            access: Access::empty(),
        };

        self.ledger.add(record).unwrap();
    }

```

After:

```rust
    fn allocate(&mut self, page: usize) {
        let region = Region::new(self.as_address(page), self.as_address(page + 1));
 
        self.ledger.map(record, Access::empty()).unwrap();
    }
```

jarkkojs:
With https://github.com/enarx/mmledger/pull/32:

```
$ cargo test
    Blocking waiting for file lock on build directory
   Compiling shim-sgx v0.3.0 (/home/jarkko/Projects/enarx/internal/shim-sgx)
    Finished test [unoptimized + debuginfo] target(s) in 6.11s
     Running unittests src/lib.rs (target/debug/deps/shim_sgx-e68e1de884ea1604)

running 2 tests
test heap::tests::mmap_munmap_oneshot ... ok
test heap::tests::mmap_munmap_incremental ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

   Doc-tests shim-sgx

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

jarkkojs:
@npmccallum this fully passes tests, after this is merged I continue to https://github.com/enarx/enarx/pull/1596 (still draft).

jarkkojs:
updated on top of my latest `mmledger` changes.

jarkkojs:
Updated from `jarkkojs/mmledger` to `enarx/mmledger`, as the required changes have been merged.

internal/shim-sgx/src/heap.rs
```diff
@@ -32,14 +35,25 @@ impl<const N: usize> Block<N> {
...
+
 /// A heap
 pub struct Heap<'a, const N: usize>
 where
     [(); N * 64]: Sized,
 {
     blk: &'a mut Block<N>,
-    map: [u64; N * 64],
     brk: Option<NonZeroUsize>,
+    ledger: Ledger<LEDGER_CAPACITY>,
```
npmccallum, line 56: Since we aren't worrying about page alignment on this type anymore, I suspect it might just be easier to pick a capacity number we think is reasonable. I think 64k is probably too big for our needs. I also think it is more reasonable to reason about in number of entries rather than by size. For example, I think 128 entries is probably more than enough for our needs once we have the ability to add pages.

This boils down to my experience with `wasmtime`. We currently have a bunch of efficiency settings turned off. However, by default `wasmtime` will attempt to map a 4GiB block with no permissions. Once we have SGX2, we don't have to allocate any pages for this. Then `wasmtime` allocates various regions within that 4GiB block (basically: rodata, data, text, stack, heap, etc). So the total number of ledger entries ends up being something like <16.

If we run out of ledger entries, we can always just fail the `mmap()` operation back to `wasmtime`. This will be observable during testing. So we should see it pretty quickly.

My vote is:

```suggestion
    ledger: Ledger<127>,
```

internal/shim-sgx/src/heap.rs
```diff
@@ -53,52 +67,26 @@ where
     /// This function is unsafe because it expects the block to be mapped RWX.
     const unsafe fn new(blk: &'a mut Block<N>) -> Self {
         let brk = None;
-        let bitmap = [0; N * 64];
         Self {
             blk,
-            map: bitmap,
             brk,
+            ledger: Ledger::new(Region::new(Address::new(0), Address::new(N * Page::SIZE))),
```
npmccallum, line 73: I think we want absolute addressing here. All the requests (i.e. `brk()`, `mmap()`, etc) are in absolute addressing. Let's avoid unnecessary conversions.

jarkkojs: At this time there is no address.

internal/shim-sgx/src/heap.rs
jarkkojs: I'll set it to 511.

jarkkojs: Fixed, and new version pushed.

jarkkojs:
Is SEV up?

jarkkojs:
Updated:
```rust
    /// Get reference to the heap's memory.
    pub fn memory(&self) -> &[Page] {
        if cfg!(test) {
            unsafe { from_raw_parts::<Page>(self.blk.0.as_ptr() as *const _, N) }
        } else {
            unsafe { from_raw_parts::<Page>(heap_address() as *const Page, N) }
        }
    }
```
And:
```
/// Get the Shim's base address.
///
/// # Safety
///
/// The pointer is not protected by any means by the compiler. Encapsulate for
/// instance to an array.
#[inline]
pub unsafe fn heap_address() -> *const u8 {
    &ENARX_HEAP_ADDRESS as *const _
}
```

internal/shim-sgx/Cargo.toml
```diff
@@ -17,7 +17,7 @@ dbg = []
 gdbstub_arch = { version = "0.1.1" , default-features = false, optional = true }
 gdbstub = { version = "0.5.0" , default-features = false, optional = true }
 goblin = { version = "0.5", default-features = false, features = [ "elf64" ] }
-primordial = { version = "^0.4.0", features = ["const-default"] }
+primordial = { git = "https://github.com/enarx/primordial", branch = "main", features = ["const-default"] }
```
haraldh, line 20: Please use git hash references, so, that if the remote branch gets updated, the enarx builds don't break.
```suggestion
primordial = { git = "https://github.com/enarx/primordial", rev = "b606b9e472e2238384b003f12fdada077e04d2fe", features = ["const-default"] }
```

internal/shim-sgx/Cargo.toml
```diff
@@ -28,6 +28,7 @@ xsave = { version = "2.0.2" }
 rcrt1 = "1.0.0"
 lset = "0.2"
 sgx = "0.3"
+mmledger = { git = "https://github.com/enarx/mmledger" }
```
haraldh, line 31: Please use git hash references, so, that if the remote branch gets updated, the enarx builds don't break.

internal/shim-sgx/src/heap.rs
```diff
@@ -53,52 +58,30 @@ where
...
-            if self.is_allocated(i) {
-                return true;
-            }
+    /// Get reference to the heap's memory.
+    pub fn memory(&self) -> &[Page] {
+        if cfg!(test) {
+            unsafe { from_raw_parts::<Page>(self.blk.0.as_ptr() as *const _, N) }
+        } else {
+            unsafe { from_raw_parts::<Page>(heap_address() as *const Page, N) }
         }
```
haraldh, line 74:
```suggestion
    pub fn memory(&self) -> &[Page] {
        &self.blk.0
```

internal/shim-sgx/layout.ld
```diff
@@ -70,5 +70,6 @@ SECTIONS {
 
     /* HEAP */
     . = ALIGN(128M);
+    HIDDEN(ENARX_HEAP_ADDRESS = .);
```
haraldh, line 73: not needed

internal/shim-sgx/src/heap.rs
```diff
@@ -2,12 +2,17 @@
 
 //! Allocate and deallocate memory on a Heap
 
+use crate::heap_address;
```
haraldh, line 5: not needed

internal/shim-sgx/src/lib.rs
```diff
@@ -70,3 +72,14 @@ extern "C" {
...
+/// Get the heap's base address.
+///
+/// # Safety
+///
+/// The pointer is not protected by any means by the compiler. Encapsulate for
+/// instance to an array.
+#[inline]
+pub unsafe fn heap_address() -> *const u8 {
+    &ENARX_HEAP_ADDRESS as *const _
+}
```
haraldh, line 85: not needed

internal/shim-sgx/src/lib.rs
```diff
@@ -61,6 +61,8 @@ extern "C" {
     pub static ENARX_EXEC_START: u8;
     /// Extern
     pub static ENARX_EXEC_END: u8;
+    /// Extern
+    pub static ENARX_HEAP_ADDRESS: u8;
```
haraldh, line 65: not needed

internal/shim-sgx/src/heap.rs
```diff
@@ -133,23 +116,25 @@ where
...
-        if old < new {
-            for page in old..new {
-                if self.is_allocated(page) {
+        match old.cmp(&new) {
+            Ordering::Less => {
+                let region = Region::new(
+                    Address::new(old * Page::SIZE),
+                    Address::new(new * Page::SIZE),
+                );
+                if self.ledger.map(region, Access::empty()).is_err() {
```
haraldh, line 125: why `Access::empty()` ?

internal/shim-sgx/src/heap.rs
```diff
@@ -166,36 +151,18 @@ where
...
-                    self.allocate(page);
-                }
-
-                return Ok(self.blk.0[range].as_mut_ptr() as *mut T);
-            }
+        if let Some(addr) = self.ledger.find_free_back(Offset::from_items(pages)) {
+            let base = self.memory().as_ptr() as usize;
+            let offset = addr.as_ptr() as usize;
+            let region = Region::new(addr, addr + Offset::from_items(pages));
+            self.ledger.map(region, Access::empty()).unwrap();
```
haraldh, line 163: why `Access::empty()` ?

haraldh (changes requested):

internal/shim-sgx/src/heap.rs
jarkkojs: They don't really matter before SGX2 is stable but does not hurt to set them right, so I'll update this.

jarkkojs:
Added `HeapTestExt` for heap functions that are only used by the unit tests in this module.

jarkkojs:
Changed internal representation to:
```
    memory: &'a mut [Page],
```

jarkkojs:
refactor `N` out of the heap.

internal/shim-sgx/src/heap.rs
```diff
@@ -33,72 +35,47 @@ impl<const N: usize> Block<N> {
...
+    memory: &'a mut [Page],
+    ledger: Ledger<511>,
     brk: Option<NonZeroUsize>,
 }
 
-impl<'a, const N: usize> Heap<'a, N>
+impl<'a> Heap<'a>
 where
-    [(); N * 64]: Sized,
+    [(); 64]: Sized,
```
haraldh, line 49: What is this for?

internal/shim-sgx/src/heap.rs
```diff
@@ -241,14 +216,35 @@ mod tests {
     const PROT: c_int = PROT_READ;
     const FLAGS: c_int = MAP_PRIVATE | MAP_ANONYMOUS;
 
+    trait HeapTestExt {
+        fn is_allocated(&self, page: usize) -> bool;
+    }
+
+    impl<'a> HeapTestExt for Heap<'a>
+    where
+        [(); 64]: Sized,
```
haraldh, line 225: What is this for?

haraldh (changes requested):

jarkkojs:
Remove the `where` clauses.

haraldh (approved):

Issue #1445, opened 2022-02-22
[Infrastructure]: Migrate to Cargo binary dependencies


platten:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Description

Migrate build system to use the new cargo binary artifacts facility.

bstrie:
I'm looking into this again. Previous attempts were stymied by https://github.com/rust-lang/cargo/issues/10431 and https://github.com/rust-lang/cargo/issues/10452 , which are fixed on nightly via https://github.com/rust-lang/cargo/pull/10433 (though we will need to bump our toolchain to get the fix). The first step will be to make wasmldr a binary artifact (since the shims are blocked on making unknown-none a Tier 2 platform: https://github.com/enarx/enarx/issues/1528). Currently I'm seeing some strange build failures that I will need to minimize and diagnose; it would not surprise me if there were still lingering bugs in the implementation of binary artifacts (which is fine; putting this feature through its paces is one of our goals here).

bstrie:
wasmldr has been made a binary artifact, although with complications; see the extensive comment at https://github.com/enarx/enarx/pull/1604 for details. In particular we encountered several bugs with binary dependencies that either needed to be worked around or endured. I've minimized the remaining bugs and filed them with Cargo:

1. https://github.com/rust-lang/cargo/issues/10525
    * This bug is mildly concerning due to how unclear the underlying cause is and the subtle ways in which it is triggered. However, as best we can tell so far its only result is compilation errors rather than anything extremely concerning such as miscompilations. We should not let this block our attempts to continue migrating to binary dependencies, but we should keep it in mind.
3. https://github.com/rust-lang/cargo/issues/10526
    * This bug is a soft blocker for porting the shims to binary dependencies, because we don't want to include both shims in the Enarx binary. However, it's not impossible to work around this, so in theory we could proceed with porting the shims.
5. https://github.com/rust-lang/cargo/issues/10527
    * This bug is something that we've simply decided to endure for the time being. Anyone making changes to any binary dependency will need to keep this in mind while developing. Notably this isn't a problem for anything that builds from scratch, e.g. CI.

PR #1604, opened 2022-03-28
refactor: make wasmldr a binary dependency and workspace member

bstrie:
Part of https://github.com/enarx/enarx/issues/1445 .

Currently wasmldr is a hidden dependency of Enarx that gets built beforehand via a build script that invokes Cargo, so that the wasmldr binary can be `include_bytes`'d into the Enarx binary at compile time. In addition to being hacky, this makes issuing new releases painful. This PR replaces the aforementioned process with a new Cargo feature called "artifact dependencies", that allow projects to depend on other crates while having access to the raw artifacts produced by those crates, in this case the wasmldr binary.

By making wasmldr into a binary dependency we can also integrate wasmldr into the Enarx workspace. This PR moves the `internal/wasmldr` directory to `src/bin/wasmldr`.

We are stressing this new feature of Cargo to its limits. Here are a list of bugs and deficiencies that were encountered while attempting to integrate this feature, and how I worked around them. Most of these bugs do not have official issues filed with Cargo yet; I will be spending this afternoon creating minimal reproductions and filing issues for them.

1. Cargo was panicking while building wasmldr's dependencies. This was reported at https://github.com/rust-lang/cargo/issues/10431 and fixed in https://github.com/rust-lang/cargo/pull/10433 . We resolved this by updating our Rust toolchain: https://github.com/enarx/enarx/pull/1593 .
2. Cargo was reporting type inference errors while compiling the structopt crate, which was a dependency of Enarx. This appears to have been due to the fact that wasmldr has a dependency on clap 3, whereas structopt depends on clap 2. It's unclear (and mildly concerning) that this could possibly be an error, however it was worked around by upgrading Enarx to clap 3, which was a desirable change on its own: https://github.com/enarx/enarx/pull/1592 . I will report this issue today.
3. The Enarx crate was failing to compile as a result of not being able to find the `CARGO_BIN_FILE_WASMLDR` environment variable. This is the environment variable that contains the full path to the built wasmldr binary. The reason for this failure has to do with the fact that wasmldr was an optional but enabled-by-default dependency, which Cargo was failing to take into account. This PR resolves this by making wasmldr a mandatory dependency, which is desirable as the fact that wasmldr was optional is apparently a historical artifact. However, we will need this to be fixed in the future in order to port the shims to being binary artifacts. I will report this issue today.
4. By becoming a member of Enarx's workspace, the wasmldr that is built as a part of the Enarx build process now respects Enarx's `.cargo/config` file rather than its own. This was worked around by deleting wasmldr's own Cargo config, as the features that it enabled were either unnecessary or useful only for debugging. In general this represents a limitation of the current design of artifact dependencies, wherein Cargo doesn't provide many knobs for customizing the artifacts that were built, since historically it has made no sense for many of these options to differ between the top-level artifact and its dependencies. The Cargo team is aware of this pain point and would presumably be receptive to proposals to improve the situation here, however that would require design work.
5. By becoming a member of Enarx's workspace, The Cargo profiles defined in wasmldr's Cargo.toml are now ignored. This was resolved by moving these profiles to Enarx's Cargo.toml and enabling them for the entire Enarx crate. If there were any profile options that needed to be enabled only in one crate but not the other, then that might present a problem in the future: while profile overrides can be used in some cases to control the flags with which dependencies are built, some flags (e.g. `panic`, `lto`) cannot be used in profile overrides because, once again, Cargo was not historically designed to consider the possibility of building multiple disparate artifacts in the same invocation. This would require design work and a proposal to address.
6. Cargo is not precisely tracking when dependencies must be rebuilt as a result of changes to artifact dependencies. This exhibits itself via the following: make a change to wasmldr, then do `cargo build` at the top level to see that wasmldr is rebuilt, but not Enarx, then immediately run `cargo build` again to see that Enarx is now rebuilt. For some reason you must invoke Cargo twice in order to have Enarx reflect any changes in wasmldr. This could make development frustrating for anyone working on wasmldr. I will report an issue for this today.

cc @npmccallum


haraldh (approved):

npmccallum:
@bstrie  while I'm glad that @haraldh  merged this promptly, it means we don't have tracking on the issues in your very good write-up. Could you open issues? This is particularly important for the "build twice bug."

bstrie:
@npmccallum I've opened issues with Cargo and noted them on the tracking issue here: https://github.com/enarx/enarx/issues/1445#issuecomment-1084531507

PR #1613, opened 2022-03-29
refactor(shim-sgx): put heap to a lazy container

jarkkojs:
This will allow to replace it in `main()`. In addition, switch from
`jarkkojs/mmledger` to `enarx/mmledger`.

### DIFF
--- a/internal/shim-sgx/src/heap.rs
+++ b/internal/shim-sgx/src/heap.rs
@@ -13,16 +13,15 @@ use primordial::{Address, Offset, Page};
 use sallyport::libc::{
     off_t, EINVAL, ENOMEM, MAP_ANONYMOUS, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE,
 };
+use spinning::{Lazy, RwLock};
 
 /// This section MUST be marked as RWX in the linker script
 #[link_section = ".enarx.heap"]
 static mut BLOCK: Block<32768> = Block::new();
 
 /// The keep heap
-pub static HEAP: spinning::RwLock<Heap<'_>> =
-    spinning::RwLock::const_new(spinning::RawRwLock::const_new(), unsafe {
-        Heap::new(&mut BLOCK.0)
-    });
+pub static HEAP: Lazy<RwLock<Heap<'_>>> =
+    Lazy::new(|| RwLock::new(unsafe { Heap::new(&mut BLOCK.0) }));
 
 /// An allocated block of memory
 #[repr(C, align(4096))]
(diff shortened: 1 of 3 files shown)


internal/shim-sgx/src/heap.rs
```diff
@@ -13,16 +13,15 @@ use primordial::{Address, Offset, Page};
...
 #[link_section = ".enarx.heap"]
 static mut BLOCK: Block<32768> = Block::new();
 
 /// The keep heap
-pub static HEAP: spinning::RwLock<Heap<'_>> =
-    spinning::RwLock::const_new(spinning::RawRwLock::const_new(), unsafe {
-        Heap::new(&mut BLOCK.0)
-    });
+pub static HEAP: Lazy<RwLock<Heap<'_>>> =
+    Lazy::new(|| RwLock::const_new(RawRwLock::const_new(), unsafe { Heap::new(&mut BLOCK.0) }));
```
npmccallum, line 24: You should be able to use the non-const constructors now.

npmccallum (changes requested):
We are very strict about one logical charge per commit. Please split this into two commits.

jarkkojs:
@npmccallum SEV failing has happened me before without actual problems in the commits. Is milan down?

jarkkojs:
> @npmccallum SEV failing has happened me before without actual problems in the commits. Is milan down?

I.e. I suspect a false positive.

haraldh (changes requested):
The `Cargo.lock` and `Cargo.toml` changes are in different commits

jarkkojs:
> The `Cargo.lock` and `Cargo.toml` changes are in different commits

Ugh, I'll fix it.

jarkkojs:
> > The `Cargo.lock` and `Cargo.toml` changes are in different commits
> 
> Ugh, I'll fix it.

Should be good now.

internal/shim-sgx/src/heap.rs
```diff
@@ -13,16 +13,15 @@ use primordial::{Address, Offset, Page};
...
 #[link_section = ".enarx.heap"]
 static mut BLOCK: Block<32768> = Block::new();
 
 /// The keep heap
-pub static HEAP: spinning::RwLock<Heap<'_>> =
-    spinning::RwLock::const_new(spinning::RawRwLock::const_new(), unsafe {
-        Heap::new(&mut BLOCK.0)
-    });
+pub static HEAP: Lazy<RwLock<Heap<'_>>> =
+    Lazy::new(|| RwLock::new(unsafe { Heap::new(&mut BLOCK.0) }));
```
haraldh, line 24: The previous solution was leaner and more performant :smile:

haraldh (approved):

internal/shim-sgx/src/heap.rs
jarkkojs: @npmccallum asked not to use `const_new` so I changed it.

npmccallum (approved):

PR #1617, opened 2022-03-30
chore(actions): workflows using configured rust toolchain

platten:


.github/workflows/lint.yml
```diff
@@ -12,10 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Get toolchain version
+        id: toolchain_version
+        run: echo "::set-output name=toolchain_version::$(grep channel rust-toolchain.toml | cut -d\" -f2)"
```
haraldh, line 17:
```suggestion
        run: awk -F '[="]' '/^[:space:]*channel/ { print "::set-output name=toolchain_version::"$3 }' rust-toolchain.toml
```

.github/workflows/test.yml
```diff
@@ -16,10 +16,13 @@ jobs:
       - run: sudo apt -o Acquire::Retries=3 update
       - run: sudo apt -o Acquire::Retries=3 install -y musl-tools
       - uses: actions/checkout@v2
+      - name: Get toolchain version
+        id: toolchain_version
+        run: echo "::set-output name=toolchain_version::$(grep channel rust-toolchain.toml | cut -d\" -f2)"
```
haraldh, line 21:
```suggestion
        run: awk -F '[="]' '/^[:space:]*channel/ { print "::set-output name=toolchain_version::"$3 }' rust-toolchain.toml
```

haraldh (changes requested):

rvolosatovs (commented):
I feel like this is way more complex than it should be, is there a reason we're not letting `rustup` to figure out the version, component list etc. from toolchain toml file directly?
If toolchain toml file does not contain some specific components we're using, why not add them?

.github/workflows/lint.yml
```diff
@@ -12,10 +12,13 @@ jobs:
...
       - uses: actions-rs/toolchain@v1
         with:
           components: rustfmt
-          toolchain: nightly-2022-03-23
+          toolchain: ${{ steps.toolchain_version.outputs.toolchain_version }}
           profile: minimal
           override: true
```
rvolosatovs, line 23: we should be able to replace this whole thing by `run: rustup show`
See https://github.com/profianinc/drawbridge/blob/39ea7b34576ce5edafa7e81aa0333c9298c4bf90/.github/workflows/test.yml#L8-L9

`rustup` is already installed on `ubuntu` images by default, `rustup show` will parse the `rust-toolchain.toml`
`rust-toolchain.toml` would need to contain all the components and targets for it to work

platten:
> I feel like this is way more complex than it should be, is there a reason we're not letting `rustup` to figure out the version, component list etc. from toolchain toml file directly? If toolchain toml file does not contain some specific components we're using, why not add them?

While this is a bit slower, it is cleaner, so changes applied. We still needed to add the toolchain manually in the sev/sgx hosted runner builds though since `rustup` was missing there.

.github/workflows/test.yml
```diff
@@ -15,13 +15,14 @@ jobs:
...
         with:
           target: x86_64-unknown-linux-musl
-          toolchain: nightly-2022-03-23
-          override: true
-          components: rust-src
+          toolchain: stable
+          profile: minimal
+      - uses: actions/checkout@v2
+      - name: Setup Rust toolchain
+        run: rustup show
```
rvolosatovs, line 25:
```suggestion
```

Unnecessary, since we're using the `toolchain` action above

rvolosatovs (commented):
Just a small comment

rvolosatovs:
> > I feel like this is way more complex than it should be, is there a reason we're not letting `rustup` to figure out the version, component list etc. from toolchain toml file directly? If toolchain toml file does not contain some specific components we're using, why not add them?
> 
> While this is a bit slower, it is cleaner, so changes applied. We still needed to add the toolchain manually in the sev/sgx hosted runner builds though since `rustup` was missing there.

If speed is an issue, we could alleviate it by using caching perhaps? (since we pin the nightly version, that should be fine at least in that case)

rvolosatovs (approved):

haraldh (approved):

Issue #1621, opened 2022-04-01
[Bug]: All `wasmldr` tests are broken


jarkkojs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

```
Error: error obtaining attestation key id

Caused by:
    No such file or directory (os error 2)
test check_tcp ... FAILED
Error: error obtaining attestation key id
```

### Expected Behaviour

No attestation error.

### Environment Information

-

jarkkojs:
How can I prevent `wasmldr` tests not to do attestation?

jarkkojs:
```
❯ git bisect good
d83730eca29485f45a1a30ad1a04633dd6f8c193 is the first bad commit
commit d83730eca29485f45a1a30ad1a04633dd6f8c193
Author: Richard Zak <richard@profian.com>
Date:   Tue Mar 22 15:29:55 2022 -0400

    feat: attestation checking for wasmldr
    
    Signed-off-by: Richard Zak <richard@profian.com>
    Co-authored-by: Nathaniel McCallum <nathaniel@profian.com>

 internal/wasmldr/Cargo.lock                   | 50 +++++++++++++++++++++++++++
 internal/wasmldr/Cargo.toml                   |  1 +
 internal/wasmldr/src/loader/configured/mod.rs | 39 ++++++++++++++-------
 3 files changed, 77 insertions(+), 13 deletions(-)
```

Issue #1618, opened 2022-03-31
[Bug]: bccce095 breaks `ENARX_BACKEND=sgx cargo test` when AESM daemon is not running


jarkkojs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

All tests fail.

### Expected Behaviour

The test, which do not use attestation pass.

### Environment Information

XPS-13 2020 model laptop

haraldh:
Just remove `ENARX_BACKEND=sgx` ... the backend is not able to run.. point.

haraldh:
or just:
```
$ sudo touch /var/run/aesmd/aesm.socket
```

platten:
This looks like a non-issue. Let's discuss.

PR #1626, opened 2022-04-03
build(sgx): `force-backend` feature

jarkkojs:
Enables a clean `cargo test` run for developer without full attestation
stack setup.

Closes: #1621
Closes: #1618


jarkkojs:
@haraldh This would really help me to not use CI for active development (which I've been doing pro-actively since everything collapsed).

jarkkojs:
I think a better name for this would be `disable-attestation` as that is what this effectively does.

jarkkojs:
Closing, as part of #1608.

PR #1631, opened 2022-04-05
build: add rusty-tags.vi to .gitignore

jarkkojs:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/
 .idea/
 .vscode/
+**/rusty-tags.vi


rvolosatovs (commented):
I think this is something that should go in your global `git` config (`core.excludesfile`), but given that we also have `.idea` and `.vscode` here (which IMO also do not belong), I guess that's fine?

jarkkojs:
> I think this is something that should go in your global `git` config (`core.excludesfile`), but given that we also have `.idea` and `.vscode` here (which IMO also do not belong), I guess that's fine?

I don't think this causes anyone any harm.

rvolosatovs:
> > I think this is something that should go in your global `git` config (`core.excludesfile`), but given that we also have `.idea` and `.vscode` here (which IMO also do not belong), I guess that's fine?
> 
> I don't think this causes anyone any harm.

Sure, that's why I don't mind merging, this, but let's see what others think.
I don't know if we want to have maintain a list of various environment-specific files here (e.g. `.DS_Store` could be one example of a file that should probably be included here if we do that, probably `tags` for ctags-style tags, vim temporary files, maybe something for emacs? kakoune? etc. etc.)

jarkkojs:
@rvolosatovs I agree that a global setting is better because otherwise same would have to spread to all repositories, so I've decided to close this.

PR #1639, opened 2022-04-05
feat: add a Go server example

rvolosatovs:
NOTE: This will still be updated

You will have to compile a custom `tinygo` from this PR https://github.com/tinygo-org/tinygo/pull/2748
See https://tinygo.org/docs/guides/build/#with-a-self-built-llvm
Make sure you do a recursive clone of repo!
```
make llvm-source
make llvm-build
make wasi-libc
make binaryen
make
```
It will take a while.

You can use the resulting `build/tinygo` to compile the example and run in enarx with:
```
tinygo build -wasm-abi=generic -target=wasi
```

cc @rjzak


bstrie:
Asking people to build LLVM from source is pretty rough. Should we close this PR until the tinygo patch is merged?

rvolosatovs:
> Asking people to build LLVM from source is pretty rough. Should we close this PR until the tinygo patch is merged?

It should also be possible to use prebuilt LLVM, at least according to the docs https://tinygo.org/docs/guides/build/#with-system-installed-llvm

The whole process is handled by tinygo build system, so it's just
```
make llvm-source
make llvm-build
```
to clone and build from within tinygo repo checkout, so it's a really trivial process.

It may take a while though (an hour or more), if that's what you're referring to as "heavy".

Do as you wish, but I think that it's nice to provide at least some kind of documentation and a starting point for people who do want to use Go and having this here serves as an indication that we're working on it. Let's discuss during standup.

We could also provide an OCI image containing a tinygo build, for example that community could use to compile their Go code. cc @platten

rvolosatovs:
> I don't think having a go-examples top-level directory is going to scale once we have support for a dozen languages and they all want their own top-level directory. Can we just have an examples/go directory instead?

That's precisely why it's a FIXME commit, `go-examples` is not the right way forward.
IIRC the plan was that you'd create a separate repository for language-specific examples?

Issue #1638, opened 2022-04-05
[Feature]: Implement `mremap()` for KVM


rjzak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Implement `mremap()` for KVM

Issue #1637, opened 2022-04-05
[Feature]: Implement _an appropriate subset_ of `mremap()` for SGX


rjzak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Implement `mremap()` for SGX. Required at least for Ruby, .Net (#1630).

jarkkojs:
OK so this is man-page for `mremap()`: https://man7.org/linux/man-pages/man2/mremap.2.html

jarkkojs:
@rjzak Can you recall, which parameters were supplied to `mremap()`?

jarkkojs:
Of a subset won't scale to some other application but then we refine it. Dealing with these highly specialized syscalls is an evolutionary process. I.e. we need to check the exact subset that Ruby uses and work from there.

rjzak:
I don't know what the parameters were, but I do have a test .wasm file you can use for testing. https://rjzak.keybase.pub/hello-rb.wasm?dl=1

jarkkojs:
> @jarkkojs I don't know what the parameters were, but I do have a test .wasm file you can use for testing. https://rjzak.keybase.pub/hello-rb.wasm?dl=1

Thanks that is sufficient!

rjzak:
@jarkkojs if you need another example, I'm sure @Deepansharora27 can share his .Net example

jarkkojs:
@Deepansharora27 Can you also provide my pre-compiled wasm-file?

jarkkojs:
> @jarkkojs if you need another example, I'm sure @Deepansharora27 can share his .Net example

Thank you!

Deepansharora27:
@jarkkojs Here is the link to it : https://keybase.pub/deepansharora27/wasm_file.wasm

jarkkojs:
> @jarkkojs Here is the link to it : https://keybase.pub/deepansharora27/wasm_file.wasm

Thank you.

jarkkojs:
I'm investigating this next week after https://github.com/enarx/enarx/pull/1608 has been merged.

haraldh:
Hmm, still not complete, because munmap() is not working.

haraldh:
> @jarkkojs Here is the link to it : https://keybase.pub/deepansharora27/wasm_file.wasm

This one needs the `SYS_clock_getres` syscall

jarkkojs:
[...]

PR #1645, opened 2022-04-06
feat: add mremap() via sallyport update

haraldh:
simple version with copying to a new location

Fixes: #1638
Fixes: #1637

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1466,7 +1466,7 @@ checksum = "73b4b750c782965c211b42f022f59af1fbceabdd026623714f104152f1ec149f"
 [[package]]
 name = "sallyport"
 version = "0.3.0"
-source = "git+https://github.com/enarx/sallyport?rev=580480b19b1542bc3da2fdafef08f895f5bb3389#580480b19b1542bc3da2fdafef08f895f5bb3389"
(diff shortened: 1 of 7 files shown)


rvolosatovs (commented):
Rather than maintaining 2 identical implementations for 2 different shims, why not implement this within sallyport directly?

haraldh:
> Rather than maintaining 2 identical implementations for 2 different shims, why not implement this within sallyport directly (so shims can override the default implementation if required)?

This will be implemented in a specialized version in the end anyway.

haraldh:
but yeah, let's do it in sallyport

internal/shim-kvm/src/hostcall.rs
```diff
@@ -439,6 +441,80 @@ impl Handler for HostCall<'_> {
...
+                            );
+
+                            eprintln!("SC> mremap() shrink = {:?}", target_addr);
+
+                            target_addr
+                        }
+                        Ordering::Less => {
+                            // simply copy the old data to a new location
+
+                            // FIXME: find out the permissions of the old segment
```
jarkkojs, line 488: `mmledger` could be used for this. it does not have query function for this but nothing prevents to add one.

rvolosatovs (approved):

PR #1670, opened 2022-04-11
refactor(shims): convert shims to binary dependencies and workspace members

bstrie:
Part of https://github.com/enarx/enarx/issues/1445 .


haraldh (changes requested):

src/bin/shim-sgx/build.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    println!(
+        "cargo:rustc-link-arg-bin=enarx-shim-sgx=-T{}/layout.ld",
+        env!("CARGO_MANIFEST_DIR")
```
haraldh, line 6: That means, that the user must not ever move the parent directories around... right?

haraldh: once build.rs is compiled

haraldh (approved):

PR #1687, opened 2022-04-11
chore: prepare for release

npmccallum:
Two commits in preparation for release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -117,7 +117,7 @@ dependencies = [
  "cc",
  "cfg-if 1.0.0",
  "libc",
- "miniz_oxide",
+ "miniz_oxide 0.4.4",
  "object",
  "rustc-demangle",
 ]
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -1798,33 +1807,22 @@ dependencies = [
 name = "sev_attestation"
 version = "0.1.0"
 dependencies = [
- "sallyport",
+ "sallyport 0.3.0",
```
bstrie, line 1810: Looks like sev_attestation still needs updating.

bstrie:
I thought we disabled the action that checked the PR title?

Cargo.lock
npmccallum: I don't see this in my latest push.

bstrie (approved):

bstrie:
The SGX builder has been randomly failing with that error all day, re-running the failed actions tends to fix it. I told Paul a while back, not sure if he's looked into it yet.

Issue #1663, opened 2022-04-07
[Feature]: Integrate `sallyport` crate into this repository


rvolosatovs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

We should integrate https://github.com/enarx/sallyport into this repository

I would like to emphasize, that the commit history of that repository __should be preserved__, that means that all commits from original repository should be rebased onto the tree within this repository rooted at a path (TBD, I think we should have `crates` directory at top-level containing e.g. `crates/sallyport`).

PR #1696, opened 2022-04-13
chore: integrate the merged sallyport crate

haraldh:
Closes https://github.com/enarx/enarx/issues/1663


haraldh:
> @haraldh I'm inclined to merge this, if only because rebasing this will not be fun... Any objections?

@npmccallum I would manually merge this with a merge commit.

crates/sallyport/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
rvolosatovs, line 1: Do we need this still?

haraldh: but the other subcrates still have it, too

crates/sallyport/tests/integration_tests/syscall.rs
```diff
@@ -724,28 +724,24 @@ fn open() {
...
-    let dev_null_0 = dev_null();
-    let dev_null_1 = dev_null();
-    let dev_null_2 = dev_null();
-
-    let dev_null_0_fd = dev_null_0.into_raw_fd();
-    let dev_null_1_fd = dev_null_1.into_raw_fd();
-    let dev_null_2_fd = dev_null_2.into_raw_fd();
-
     run_test(2, [0xff; 16], move |i, platform, handler| {
+        let dev_null_0 = dev_null();
```
rvolosatovs, line 728: this should be `dev_null().into_raw_fd()` or otherwise we should use `as_raw_fd()` in the constructors.
The FD might get closed while being used with current implementation, I believe, since files are moved in the constructors and thus go out of scope

rvolosatovs (approved):
LGTM
```
$ diff -qr ../sallyport crates/sallyport 
Files ../sallyport/src/elf.rs and crates/sallyport/src/elf.rs differ
Files ../sallyport/tests/integration_tests/syscall.rs and crates/sallyport/tests/integration_tests/syscall.rs differ
Files ../sallyport/tests/mod.rs and crates/sallyport/tests/mod.rs differ
```

.github/workflows/test.yml
```diff
@@ -51,8 +51,16 @@ jobs:
       - run: sudo apt -o Acquire::Retries=3 update
       - run: sudo apt -o Acquire::Retries=3 install -y musl-tools
       - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
```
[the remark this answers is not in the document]
haraldh: installing `rustup` in the first place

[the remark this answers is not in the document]
haraldh: removed

.github/workflows/test.yml
```diff
@@ -66,28 +66,58 @@ jobs:
...
+        env:
+          MIRIFLAGS: "-Zmiri-disable-isolation"
+        with:
+          command: miri
+          args: test --manifest-path ${{ matrix.crate.path }}/Cargo.toml
+    strategy:
+      fail-fast: false
+      matrix:
+        crate:
+          - { name: sallyport, path: crates/sallyport }
```
rvolosatovs, line 116: We probably want to enable miri for other crates as well (eventually?)

rvolosatovs (approved):

.github/workflows/test.yml
haraldh: exactly.. but not now

PR #1714, opened 2022-04-14
refactor: move bindeps into crates directory

bstrie:
This is part of making our new workspace more uniform.


haraldh (approved):

bstrie:
Note that the coverage decrease above is merely due to the fact that moving these directories has confused coveralls.

PR #1720, opened 2022-04-14
docs: overhaul readme badges and remove cargo-readme action

bstrie:
See a preview of what the new badges look like here: https://github.com/bstrie/enarx/tree/badges#readme

* The "lint workflow" badge has been removed, as this information is fairly useless to users.
* The "enarxbot workflow" badge has been removed, as no workflow by this name exists, so any status displayed by this badge was entirely imaginary.
* The "test workflow" badge has had its URLs cleaned up to reflect what is supported according to GitHub's docs.
* The "time to resolve an issue" badge and the "percentage of issues open" badge have been replaced by a "number of open issues with the 'bug' label" badge. The old badges were highly misleading as our issue tracker is also used for things like far-off wishlist features which make the old numbers useless. The new badge better reflects data that users will be interested in. The badge now links to the "bug" label on the issue tracker.
* The "maintenance" badge has been upgraded. Rather than being a hardcoded value, the badge now displays the number of commits to the repo over the past year, and now links to the repo's Pulse page. In addition, the badge text is now spelled correctly for the first time in two years. :P 
* A "crates.io" badge has been added to display the most recent release. This also links to Enarx's crates.io page.
* A "coverage" badge has been added to display our code coverage. This number is slightly misleading as our ability to test some aspects of our code is artificially limited, but perhaps we can address that in the future.

This also removes the cargo-readme action. It's unclear what value this action currently provides, as the primary purpose of cargo-readme is to make it easier to test doc comments contained in a readme, but none of our readmes contain any Rust code to test. On the other hand, supporting cargo-readme makes us clutter our top-level repo with a README.tpl file and our Cargo.toml with a vestigial `[badges]` section, and also adds to our rapidly-ballooning GitHub Actions queue on every PR. The only remaining use of the cargo-readme action is to make sure the crate-level documentation is kept in-sync with the readme, which I've resolved by making the crate-level document `include!` the readme so that it shows up in the rendered rustdoc output.


bstrie:
I am starting to suspect that coveralls might be too unreliable to gate PRs on.

haraldh:
you are right... I hopefully turned it off

haraldh:
will look into switching to codecov.io

haraldh:
can we use this: https://blog.guillaume-gomez.fr/articles/2021-08-03+Improvements+for+%23%5Bdoc%5D+attributes+in+Rust
???

bstrie:
Indeed, we absolutely can use include inside of doc comments. I was going to make that part of a future PR but I can make it part of this one if you'd like (the disabling of cargo-readme was already a last-minute addition to this PR, so it probably makes sense to do it now too).

haraldh:
yes please

haraldh:
rebase and it's using codecov.io instead

haraldh:
Here is the badge for codevoc: https://app.codecov.io/gh/enarx/enarx/settings/badge

haraldh:
[codecov](https://codecov.io/gh/enarx/enarx)

haraldh (changes requested):
You are missing the sallyport readme

bstrie:
Yes, I was also missing the exec-wasmtime readme. Both are updated now.

haraldh (approved):

bstrie:
@haraldh Since you have permissions, you will need to also change the repo settings to no longer make the cargo-readme actions required.

Issue #1563, opened 2022-03-18
Survey our complete list of dependencies


bstrie:
In the long term we will want to reduce our trusted computing base (TCB) as much as possible. One of the ways in which our TCB is currently larger than it needs to be is due to the use of a large number of third-party Rust dependencies. This doesn't necessarily mean that we need to remove all of our third-party dependencies; rather, we want to reduce our usage of dependencies to only those that are strictly necessary. Eventually we will want to perform audits of all of our dependencies, and having fewer dependencies will make that eventual audit easier.

It's worth noting that not all dependencies present the same amount of risk:

1. Any dependency owned by the Rust organization itself (e.g. the libc crate) is less of a concern. There is no avoiding the fact that the Rust organization will inevitably be a part of our TCB. The only concern here is if an attacker manages to compromise the Rust organization; we can imagine that they might find it easier to slip something in to one of the official third-party crates than to target something with much more eyes on it such as the Rust compiler itself.
2. Any dependency owned by an official Rust team member is less of a concern (with the same caveats as above). This includes crates like anyhow, owned by David Tolnay, who is a Libs Team member.
3. Any dependency that is extremely widely used, despite being a more prominent target for attackers, would also be commensurately harder to hide a backdoor in. This includes crates like rand.

One problem is that the above categories are not always cut-and-dry; for example, on crates.io the libc crate is owned not just by the rust-lang organization, but also directly by a few users, some of whom may no longer be active; this is probably merely due to history, and we should work with the Rust team to tighten up the access control here.

This also doesn't mean that we should tolerate unnecessary dependencies even if they're owned by a trustworthy source. Unnecessary dependencies contribute to unnecessarily longer compilation times, which is both an annoyance for regular developers and a roadblock to new contributors (I know of at least one volunteer contributor on a low-end computer that regularly shuts down from overheating). Furthermore, we care greatly about binary size, and a dependency that doesn't exist can't contribute to binary bloat. Instead, this just means that we should first prioritize examining dependencies that are lesser-used or from unknown authors.

As an example, here's what our list of dependencies looks like as of this writing:
<details>
<summary>List of collective dependencies in the Enarx repo (including internal crates) as of 2022-03-18 (click to expand)</summary>
<pre>
addr2line v0.17.0
adler v1.0.2
aead v0.4.3
aes-gcm v0.9.4
aes v0.7.5
aho-corasick v0.7.18
ambient-authority v0.0.1
ansi_term v0.12.1
anyhow v1.0.56
[...]

PR #1759, opened 2022-04-20
chore: opt-out of default features for all crates

bstrie:
As an experiment I mechanically added `default-features = false` to all of our dependencies in all crates in the workspace, and fixed any errors that cropped up by explicitly enabling necessary features. As a result `cargo build` builds 23 fewer crates (413 -> 390), `cargo build --release` builds 25 fewer (422 -> 397), `cargo test` builds 20 fewer (475 -> 455), and `cargo test --release` builds 22 fewer (485 -> 463). This is a step towards reducing our dependency graph (https://github.com/enarx/enarx/issues/1563), although let me know if you object to the current approach, which just throws `default-features = false` on every crate even if that crate has no features at all (though one could argue that this is useful, as it prevents us from automatically including any default features those crates may add in the future).


rvolosatovs (approved):
I wish there was a better way to do this, but I think the benefits of this approach outweigh the cons.
Did you test `cargo doc --features doc` of crates? (At least sallyport uses the `doc` feature)

Issue #1780, opened 2022-04-25
[Bug]: feature flags have no effect to `shim-sgx`


jarkkojs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

* `disable-sgx-attestation` does not apply.
* `dbg` does not lead to `debugln`'s to print out.

### Expected Behaviour

-

### Environment Information

-

jarkkojs:
blocker for SGX2 features, it's too difficult and time-consuming to debug it.

PR #1788, opened 2022-04-26
fix: propagate features to bindeps

haraldh:
Fixup of commit 5639b589079563299211d23b919554e69a5becf6, which removed
all default features.

Fixes: #1780

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,9 @@ backend-sev = ["backend-kvm", "ureq"]
 backend-sgx = ["sgx", "ureq", "x86_64"]
 
 # non-default features
-gdb = ["gdbstub"]
-dbg = []
-disable-sgx-attestation = []
+gdb = ["gdbstub", "enarx-shim-kvm/gdb", "enarx-shim-sgx/gdb"]
+dbg = [ "enarx-shim-kvm/dbg", "enarx-shim-sgx/dbg" ]
+disable-sgx-attestation = ["enarx-shim-sgx/disable-sgx-attestation"]
 
 [dependencies]
 anyhow = { version = "1.0.56", features = ["std"], default-features = false }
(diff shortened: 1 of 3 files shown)


haraldh:
@bstrie too bad making the bindeps shims optional does not work somehow.

bstrie:
Making bindeps optional is tracked by https://github.com/rust-lang/cargo/issues/10526 .

Cargo.toml
```diff
@@ -18,9 +18,9 @@ backend-sev = ["backend-kvm", "ureq"]
 backend-sgx = ["sgx", "ureq", "x86_64"]
 
 # non-default features
-gdb = ["gdbstub"]
-dbg = []
-disable-sgx-attestation = []
+gdb = ["gdbstub", "enarx-shim-kvm/gdb", "enarx-shim-sgx/dbg"]
```
bstrie, line 21: Can this be `gdb = ["dbg", "gdbstub"]`?

bstrie: Wait, is this correct? Why is KVM enabling "gdb" whereas SGX is enabling "dbg"? The names of these flags just need to be changed, they're far too easy to confuse.

jarkkojs:
I'll test this soon. The change itself looks legit.

Cargo.toml
haraldh: oops.. typo.. good catch :)

jarkkojs (dismissed):
`disable-sgx-attestation` and `dbg` work for `shim-sgx`

Cargo.toml
haraldh: yeah, first we had `dbg` ... and then `gdb` was added... feel free to rename either of them and the documentation

bstrie (approved):

PR #1794, opened 2022-04-27
feat: add direct backend

haraldh:
This backend calls `exec-wasmtime` directly via a library call, without using any elf binary and shim.
    
```
❯ enarx run --backend direct hello_wasi_snapshot1.wasm
Hello, world!
```


bstrie:
@npmccallum Can you elaborate on how this makes the static binary problem more acute?

Cargo.toml
```diff
@@ -12,10 +12,11 @@ categories = ["os", "os::linux-apis", "network-programming", "hardware-support"]
 exclude = [".github/"]
 
 [features]
-default = ["backend-kvm", "backend-sev", "backend-sgx"]
+default = ["backend-kvm", "backend-sev", "backend-sgx", "backend-direct"]
```
bstrie, line 15: Do all of these really need to be on by default? The whole reason that we're splitting out the exec into two crates is to avoid unnecessarily building wasmtime twice, so why bother if we're going to build it twice anyway?

src/backend/direct.rs
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::workldr::Workldr;
+
+use std::sync::{Arc, RwLock};
+
+use anyhow::{bail, Result};
+
+pub struct NilWorkldr;
```
bstrie, line 9: Let's be consistent about naming and rename `Nil` to `Direct`. Or maybe we should call this entire backend the "Wasm" backend and use that instead?

src/workldr/exec_wasmtime/mod.rs
```diff
@@ -25,7 +25,7 @@ pub(crate) mod test {
         let wasmldr = Box::new(Wasmldr);
         assert_eq!(
             wasmldr.exec(),
-            include_bytes!(env!("CARGO_BIN_FILE_ENARX_EXEC_WASMTIME"))
+            include_bytes!(env!("CARGO_BIN_FILE_ENARX_EXEC_WASMTIME_BIN"))
```
bstrie, line 28: It occurs to me that these tests are meaningless, I cannot imagine any problem that they would actually detect. But maybe that's for a different PR to fix.

bstrie (commented):
So AIUI the reason to split out exec-wasmtime into lib and bin crates is so that we can make the lib optional and avoid building wasmtime twice, once for the host and once for musl, since wasmtime takes a while to build. @npmccallum This isn't a new bug in bindeps, because it's expected behavior; bindeps allows you to have both an artifact and a normal dependency via the `lib = true` key, but there's no facility to make `lib = true` optional, and it's unclear how you'd even expose that. So this is bindeps working as intended, and is not relevant to the known `optional = true` bug ( https://github.com/rust-lang/cargo/issues/10526 ).

Instead, the ultimate fix for this would be multidep ( https://github.com/rust-lang/cargo/pull/10061 ), which would allow us to have a single exec-wasmtime crate and rely on it twice from within Cargo.toml, once as an optional direct dependency and once as a bindep. Please leave a FIXME somewhere mentioning that we can undo this split once multidep arrives.

Cargo.toml
haraldh: in our CI jobs we turn off the default features. What backend is on by default should be discussed separately in a broader group.

src/backend/direct.rs
haraldh: But it is an empty Exec

haraldh: Workldr might be renamed to Exec and this to NilExec or NullExec

haraldh: ah..  name bikeshedding 😀

haraldh:
Shouldn't we open an issue which tracks the external multidep and a second one depending on this to fix our monorepo?

bstrie:
@haraldh https://github.com/enarx/enarx/issues/1798 and https://github.com/enarx/enarx/issues/1799. Can you add the first as a FIXME comment?

haraldh:
> @haraldh #1798 and #1799. Can you add the first as a FIXME comment?

@bstrie done

PR #1801, opened 2022-04-28
ci(actions): assorted GH workflow related fixes

platten:
- Fix `workflow_dispatch` to enable on-demand workflow execution
- Address bug caused by `examples/tcp_server` not excluded from workspace
- Enable tests for executing `enarx` using `nil` backend on workflows

### DIFF
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -1,6 +1,5 @@
 on:
   workflow_dispatch:
-    branches: [ main ]
   schedule:
     - cron: '15 3 * * 4'
 concurrency:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,4 +97,4 @@ opt-level = 3
 opt-level = 3
 
 [workspace]
-members = ["crates/*"]
+members = ["crates/*", "examples/*"]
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -98,3 +98,5 @@ opt-level = 3
 
 [workspace]
 members = ["crates/*"]
+exclude = ["examples/tcp_server"]
```
haraldh, line 101: why?

.github/workflows/cargo-update.yml
```diff
@@ -1,6 +1,5 @@
 on:
   workflow_dispatch:
-    branches: [ main ]
```
haraldh, line 3: on every branch? Why not add branches explicitly?

haraldh (changes requested):

Cargo.toml
platten: I was getting failures in CI. We either exclude it or include it, but not defining it has triggered intermittent issues.

.github/workflows/cargo-update.yml
platten: To prevent issues like this: https://github.com/platten/enarx/actions/runs/2241459166

Cargo.toml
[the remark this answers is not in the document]
platten: I thought we wanted to keep things small, that's why I excluded it. Including it works though! https://github.com/platten/enarx/runs/6229206531?check_suite_focus=true

PR #1802, opened 2022-04-29
feat: rename `workldr` to `exec`

haraldh:
and make picking the `exec` more generic.


src/exec/mod.rs
```diff
@@ -31,13 +38,53 @@ use once_cell::sync::Lazy;
...
 /// Basically, this is a generic view of exec_wasmtime.
-pub trait Workldr: Sync + Send {
+pub trait Exec: Sync + Send {
     /// The name of the Workldr
     fn name(&self) -> &'static str;
 
     /// The builtin Workldr binary (e.g. exec_wasmtime)
     fn exec(&self) -> &'static [u8];
+
+    fn with_backend(&self, backend: &dyn Backend) -> bool;
```
rvolosatovs, line 48: Should we add a comment documenting what it does?

rvolosatovs (approved):
LGTM, but some documentation regarding what `pick` does (picking a suitable workldr given a backend) and what `with_backend` does would help

src/exec/mod.rs
haraldh: done

rvolosatovs (approved):

PR #1812, opened 2022-05-03
refactor(shim-sgx): reorder handler functions alphabetically

jarkkojs:
I spent a lot of time thinking where to place a new function to
`Handler`. Reorder function alphabetically so that this time is
used instead for something productive.

Closes https://github.com/enarx/enarx/issues/1813

### DIFF
--- a/crates/shim-sgx/src/handler/mod.rs
+++ b/crates/shim-sgx/src/handler/mod.rs
@@ -126,35 +126,6 @@ impl<'a> Write for Handler<'a> {
 }
 
 impl guest::Handler for Handler<'_> {
-    fn sally(&mut self) -> sallyport::Result<()> {
-        // prevent earlier writes from being moved beyond this point
-        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::Release);
-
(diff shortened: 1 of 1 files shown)


rvolosatovs (commented):
I'm not sure this improves the organization - the methods above are dependencies of everything below.

For example, see the `sallyport` crate `Handler` trait, for each kind of calls, there are "sections" defined https://github.com/enarx/enarx/blob/45486e4a84e7457682730d1584c891bd9881fd3d/crates/sallyport/src/guest/handler.rs#L85
https://github.com/enarx/enarx/blob/45486e4a84e7457682730d1584c891bd9881fd3d/crates/sallyport/src/guest/handler.rs#L996

I'll let others (@haraldh ?) decide

jarkkojs:
I'll move this to draft but **do** want to continue on this (probably post-release).

The reason is that I need to take a few days away from enarx to look into some issues in the kernel side of SGX. Turning to draft.

jarkkojs:
I'll close this PR and keep the issue open. I'll include the change as part of a PR instead.

PR #1823, opened 2022-05-04
feat(rust_syscall_tests): use the `rcrt1::x86_64_linux_startup!` macro

haraldh:
and convert `main` return value with the `Termination` trait to an
`exit()` value.

Also add some convenience io macros like `dbg!()`, `println!()` and
`eprintln()!`.

Also `eprintln!()` the `panic()` `PanicInfo`.

### DIFF
--- a/tests/rust_syscall_tests/src/io.rs
+++ b/tests/rust_syscall_tests/src/io.rs
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Traits, helpers, and type definitions for core I/O functionality.
+
+use core::fmt;
+
+struct Stdio<const FD: i32>;
+
+impl<const FD: i32> fmt::Write for Stdio<FD> {
+    #[inline(always)]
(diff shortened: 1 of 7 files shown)


tests/rust_syscall_tests/src/io.rs
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Traits, helpers, and type definitions for core I/O functionality.
+
+use core::fmt;
+
+struct StdIO<const FD: i32>;
```
rvolosatovs, line 7: The casing of this type is really painful. Not sure it's an improvement, but what about just `Stdio`?

haraldh: oops, sorry.. changed..

rvolosatovs (approved):

haraldh:
@rvolosatovs rebased due to conflicts

rvolosatovs (approved):

PR #1839, opened 2022-05-06
refactor(tests): reorganize integration tests

rvolosatovs:
Reorganize the integration tests to have a clear, organized and maintainable structure.


tests/integration.rs
```diff
@@ -1,6 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #![allow(dead_code)]
+#![cfg(not(feature = "gdb"))]
+
+#[cfg(all(not(miri), not(feature = "backend-nil")))]
+mod exec;
+#[cfg(all(not(miri), not(feature = "backend-nil")))]
```
haraldh, line 8: veto

haraldh (changes requested):

tests/integration.rs
haraldh: this removes testing if exec and syscalls if all features are enabled

haraldh: make that `not(target_os = "linux")` maybe?

bstrie: Elaborate?

haraldh: sorry this one:
```
#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```

bstrie: > this removes testing of exec and syscalls if all features are enabled

These tests are already disabled if all features are enabled, because of the not(feature = "gdb")) above. More concerning is that this is disabled under the default feature profile.

tests/integration.rs
```diff
@@ -2,6 +2,23 @@
 
 #![allow(dead_code)]
 
+#[cfg(all(
+    not(miri),
+    not(feature = "gdb"),
```
bstrie, line 7: Just out of curiosity, why have we always disabled our tests under gdb?

bstrie:
I'm prepared to approve as soon as it's no longer marked as a draft.

tests/integration.rs
rvolosatovs: Just tried disabling that, and they all hang with the feature enabled. Probably it's waiting for connection to be established?

rvolosatovs:
> I'm prepared to approve as soon as it's no longer marked as a draft.

It still does not work for `backend-nil` case: https://github.com/enarx/enarx/runs/6325754567?check_suite_focus=true

bstrie:
Nil coverage test is failing because of this:
```
cargo llvm-cov --no-default-features --features=backend-nil --test integration --bin enarx --target x86_64-unknown-linux-gnu --manifest-path ./Cargo.toml --lcov --output-path lcov.info
```
The current cfgs only disable these tests when not on linux and x64, which we are here, so either we need to make the cfgs more precise or reintroduce `--test integration wasm`.

bstrie:
> And there does not seem to be a way to select the tests to run in cargo-llvm-cov (except for the --test)

Ah, quite possible.

haraldh (approved):

tests/integration.rs
haraldh: it's waiting for gdb to connect..

PR #1861, opened 2022-05-09
test: enforce maximum shim size in release tests

### LINKED ISSUE #1609: [Infrastructure]: Have CI gate PRs on binary size

### Code of Conduct

### Description

We care about the binary size of Enarx, the shims, and wasmldr. But right now binary size regressions are only detected when someone manually inspects the binaries. CI should aid us in detecting size regressions. I'm unclear on the best way to achieve this, but one strategy is the following:

1. For each PR, do `cargo build` on each of the following: enarx, shim-kvm, shim-sev, wasmldr.
2. Then, build all of the above tools without the PR, i.e. using the current main branch.
3. If the new binary size is more than X% larger than its prior version, then cause the CI job to fail.

Deciding the proper size for X% will be interesting. Increasing binary size isn't always a bad thing, and some PRs will be fully expected to increase the binary size, so approvers will be expected to understand when they should override the CI and merge a PR anyway. So let's aim for making X% quite low, but still high enough that it will not trigger on the random minor perturbations that result from the feeding the optimizer slightly different code.

Ideally the job would report the exact percentage increase when it fails. Ideally I'd also like a bot to leave a comment in the PR itself so that people don't have to go digging for this information.

bstrie:
Add new unit tests to the Enarx crate to guard against regressions in binary size for the shims.

Tests are only run in release mode, and will fail if a shim exceeds 500 KB in size.

This also removes a useless test from exec-wasmtime.

Closes #1609.

### DIFF
--- a/src/backend/kvm/mod.rs
+++ b/src/backend/kvm/mod.rs
@@ -108,3 +108,22 @@ impl crate::backend::Backend for Backend {
         Ok(Vec::new())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    #[cfg_attr(debug_assertions, ignore)]
+    fn shim_kvm_binary_size() {
+        use crate::backend::Backend;
+
+        let max_shim_size = 500_000;
+        let shim = super::Backend.shim();
+        if shim.len() > max_shim_size {
+            panic!(
+                "shim size should be less than {} bytes, but is {} bytes",
+                max_shim_size,
+                shim.len()
+            );
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


bstrie:
The decrease in coverage is expected from deleting the pointless `is_builtin` test.

haraldh:
Did you test this with strip=false?

bstrie:
Why would we need to test with strip=false?

haraldh:
> Why would we need to test with strip=false?

ah, release tests... sorry

haraldh (approved):

PR #1862, opened 2022-05-09
fix: link unknown-none binaries (except shim-kvm) with LLD

bstrie:
By specifying the rustflags specifically for shim-kvm, we can avoid having to use the Cargo config, which is generally better as it means that someone using (for example) `cargo install` won't have to preface the command with something grotesque like `CARGO_TARGET_X86_64_UNKNOWN_NONE_RUSTFLAGS="-C linker=gcc"`.

This does mean using a new unstable Cargo feature, but that's acceptable for the moment as bindeps won't be stabilizing any time soon, and in the long term we hope to not have to manually specify the linker for shim-kvm anyway.


bstrie:
We'll investigate and reopen this in the future.

PR #1874, opened 2022-05-11
build(nix): various improvements

rvolosatovs:
- Enable `nix run .`, i.e. users can now do `nix run github:enarx/enarx info` to build and test `enarx` compiled for the host target directly from `main` and run on success `enarx info` (or any other subcommand) with no other dependency than `nix` on any x86_64 Linux distribution supported by `nix`
- Refactor `enarx-docker` to provide an `enarx` binary in `PATH`
- Split the nix CI job to benefit from concurrent builds and get clearer output
- Filter the source to improve caching efficiency
- Test, whether the binary is indeed static as part of the build
- Execute the static binary on Ubuntu

For testing Docker image (will be added to CI):
```
docker run --rm $(docker load < ./result | sed 's/Loaded image(s): \(.*\)/\1/') enarx info
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,5 +150,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2.4.0
     - uses: cachix/install-nix-action@v16
+    - run: nix build -L '.#enarx-static'
     - run: nix run -L '.#enarx-static' info
+    - run: ./result/bin/enarx info
     - run: nix build -L '.#enarx-docker'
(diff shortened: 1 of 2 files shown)


bstrie:
When do the nix-develop, nix-static, and nix-dynamic jobs run? Is it on every PR, or on every tag, or is it manually triggered?

rvolosatovs:
> > they will be fast once we have #1875
> 
> How far off is that? In the meantime, how much time do they take?

Longest is, by some reason, the dynamic build at the moment at 11-12m on average. The static and develop builds are around half of that.

haraldh (approved):

Issue #1876, opened 2022-05-11
[Feature]: Mac support for `nix`


rvolosatovs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Add support for Mac in `nix` flake

platten:
@rvolosatovs with #1881 and #1906 , is this still an issue?

rvolosatovs:
Closed, because the binaries appear portable (i.e. work on actual real Mac systems)
If something breaks, we need to reopen this issue and such breakages should be caught by CI (I do not know if sufficient release QA process is in place), not reported by our users.

PR #1881, opened 2022-05-12
ci: add release and nix workflows

rvolosatovs:
Refs https://github.com/enarx/enarx/issues/1876 https://github.com/enarx/enarx/issues/1888

Building arm64 binaries on CI in QEMU is too slow, therefore we should add support for cross-compilation to the flake in #1888. I have a draft for that already.
Mac binaries are still non-portable, that should be fixed in #1876.

To build a static `linux-aarch64` package via QEMU: 
```
nix build '.#packages.aarch64-linux.enarx-static'
```


.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+jobs:
+  build:
+    strategy:
+      matrix:
+        platform:
+        - os: ubuntu-20.04
+          target: x86_64-unknown-linux-musl
+        - os: macos-11
+          target: x86_64-apple-darwin
+        # TODO: provide aarch64 builds
```
rjzak, line 26: How about macOS [fat binary](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) with the `lipo` tool?

rvolosatovs: Is it reproducible?

bstrie:
Hm, the nix CI actions appear to be stuck.

rvolosatovs:
> Hm, the nix CI actions appear to be stuck.

Yes, but it does not look like a workflow issue:
```
Waiting for a runner to pick up this job... 
```

I'll try to re-push the HEAD

rvolosatovs:
@bstrie removed MacOS from test matrix, by some reason MacOS `nix` workflows would never start, all green now.

haraldh (approved):

PR #1890, opened 2022-05-17
feat: add Windows support with the `nil` backend

haraldh:
Also support `test --workspace` on MacOS and Windows.

### DIFF
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -13,12 +13,10 @@ mod loader;
 use loader::Loader;
 use url::Url;
 
-use std::io::Read;
-use std::mem::forget;
+#[cfg(unix)]
 use std::os::unix::io::{FromRawFd, RawFd};
-use std::os::unix::net::UnixStream;
 
-use anyhow::Context;
+#[cfg(unix)]
 use serde::{Deserialize, Serialize};
 
 /// Name of package entrypoint file
@@ -28,6 +26,7 @@ pub const PACKAGE_ENTRYPOINT: &str = "main.wasm";
 pub const PACKAGE_CONFIG: &str = "Enarx.toml";
 
 /// Package to execute
+#[cfg(unix)]
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields, tag = "t", content = "c")]
 pub enum Package {
@@ -43,9 +42,26 @@ pub enum Package {
     },
 }
 
+/// Package to execute
+#[cfg(windows)]
+#[derive(Debug)]
+pub enum Package {
+    /// Remote URL to fetch package from
+    Remote(Url),
+
+    /// Local package
+    Local {
+        /// Open WASM module file
+        wasm: std::fs::File,
+        /// Optional open config file
+        conf: Option<std::fs::File>,
+    },
+}
+
 /// The Arguments
 // NOTE: `repr(C)` is required, otherwise `toml` serialization fails with `values must be emitted before tables`
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug)]
+#[cfg_attr(unix, derive(Deserialize, Serialize))]
 #[repr(C)]
 pub struct Args {
     /// Package
@@ -53,17 +69,7 @@ pub struct Args {
 }
 
 /// Execute
-pub fn execute() -> anyhow::Result<()> {
-    // This is the FD of a Unix socket on which the host will send the TOML-encoded execution arguments
-    // and shutdown the write half of it immediately after.
-    // TODO: Use the write half of the socket to write logs/errors to the host
-    let mut host = unsafe { UnixStream::from_raw_fd(3) };
-
-    let mut args = String::new();
-    host.read_to_string(&mut args)
-        .context("failed to read arguments")?;
-    let args = toml::from_str::<Args>(&args).context("failed to decode arguments")?;
-
+pub fn execute_with_args(args: Args) -> anyhow::Result<()> {
     // Step through the state machine.
     let configured = Loader::from(args);
     let requested = configured.next()?;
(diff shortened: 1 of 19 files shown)


crates/exec-wasmtime/src/lib.rs
```diff
@@ -43,17 +42,53 @@ pub enum Package {
...
+    let attested = requested.next()?;
+    let compiled = attested.next()?;
+    let connected = compiled.next()?;
+    let completed = connected.next()?;
+    drop(completed);
+    Ok(())
+}
+
+#[cfg(unix)]
 /// Execute
```
rvolosatovs, line 85: Let's document that this assumes a Unix socket on fd 3 and reads and parses args from there.

src/main.rs
```diff
@@ -167,23 +189,36 @@ fn main() -> Result<()> {
...
+                    vec![wasm].into_iter().chain(conf),
+                ))
+            };
+
+            #[cfg(windows)]
+            let get_pkg = || {
+                let (wasm, conf) = open_package(module, wasmcfgfile)?;
+                Ok((
+                    Package::Local { wasm, conf },
+                    empty(), // vec![wasm].into_iter().chain(conf)
```
rvolosatovs, line 210: What's the commented-out code for?

src/main.rs
```diff
@@ -238,7 +275,18 @@ fn main() -> Result<()> {
...
                         ))
-                    })?
+                    };
+
+                    #[cfg(windows)]
+                    let get_pkg = || {
+                        let (wasm, conf) = open_package(wasm, conf)?;
+                        Ok((
+                            Package::Local { wasm, conf },
+                            empty(), // vec![wasm].into_iter().chain(conf)
```
rvolosatovs, line 285: see above

src/main.rs
```diff
@@ -238,7 +275,18 @@ fn main() -> Result<()> {
                             },
                             vec![wasm].into_iter().chain(conf),
```
rvolosatovs, line 276: Couldn't we do the config switch on this line instead to reduce duplication?

rvolosatovs (commented):
Why don't we use something like https://docs.microsoft.com/en-us/windows/win32/ipc/named-pipes instead, or better yet an `AF_UNIX` socket (I looked through Rust stdlib docs, but it seems that Windows support here is not straight-forward).
See also https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

src/main.rs
haraldh: done

crates/exec-wasmtime/src/lib.rs
haraldh: done

crates/exec-wasmtime/src/lib.rs
```diff
@@ -43,17 +42,53 @@ pub enum Package {
...
+#[derive(Debug)]
+pub enum Package {
+    /// Remote URL to fetch package from
+    Remote(Url),
+
+    /// Local package
+    Local {
+        /// Open WASM module file Handle
+        wasm: std::fs::File,
+        /// Optional open config file descriptor
```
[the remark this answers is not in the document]
haraldh: done

src/main.rs
haraldh: removed it .. see below

haraldh:
Although I created the uds_windows crate, I think windows is lacking socketpair and abstract unix sockets. While we could use temp files and pass the filename, I like the direct File approach more

bstrie (commented):
Looks pretty good, although in general I'm worried that we're growing a lot of conditional-compilation profiles that are going to be hard to keep track of.

haraldh:
Although I created the [uds_windows crate](https://crates.io/crates/uds_windows), I like the direct File approach more.

Because windows is lacking socketpair and abstract unix sockets, we would have to use named pipes, which would mean different structs anyway.

And still we would have to serialize things though serde_json... which would mean to transform `Handle` -> `(void *)` to `usize` and then back again, which is more ugly than what I propose here I guess.

crates/exec-wasmtime/src/lib.rs
```diff
@@ -13,12 +13,10 @@ mod loader;
...
 use url::Url;
 
-use std::io::Read;
-use std::mem::forget;
+#[cfg(unix)]
 use std::os::unix::io::{FromRawFd, RawFd};
-use std::os::unix::net::UnixStream;
 
-use anyhow::Context;
+#[cfg(unix)]
```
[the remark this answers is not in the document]
haraldh:
```
53   | /         /// Open WASM module file
54   | |         wasm: std::fs::File,
     | |___________________________^ the trait `Deserialize<'_>` is not implemented for `std::fs::File`
```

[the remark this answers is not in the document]
haraldh: > And while we're at it, did you want to make serde a conditional dependency in Cargo.toml?

done ... [8c58a1d](https://github.com/enarx/enarx/pull/1890/commits/8c58a1d967ff192c352195de12fa5253696120c7)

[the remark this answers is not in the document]
haraldh: oops, that didn't work out :) Of course it is still used.... removed again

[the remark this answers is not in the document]
haraldh: see https://github.com/enarx/enarx/runs/6503206014?check_suite_focus=true

haraldh:
> Although I created the [uds_windows crate](https://crates.io/crates/uds_windows), I like the direct File approach more.
> 
> Because windows is lacking socketpair ....

@rvolosatovs @bstrie 

I stand corrected... someone with the name "Harald Hoyer" [added some undocumented method `pair()`](https://github.com/haraldh/rust_uds_windows/commit/1d4073d3f237400a7509721c6b81162a216e06a7) on Dec 12 2018 ... **cough**

https://docs.rs/uds_windows/1.0.2/uds_windows/struct.UnixStream.html#method.pair

Let me revise my strategy and maybe revert this PR partially... :) `RawSocket` is `u64`, so we could merge `RawFd` and `RawSocket` in a u64.

Issue #1894, opened 2022-05-19
Improve code coverage


bstrie:
As of this writing our overall code coverage sits at 58%: https://app.codecov.io/gh/enarx/enarx . It would be nice to get that somewhere north of 90%. People have mentioned that there are some difficulties related to our ability to instrument the shims which is dragging down our percentage, so if something that we're doing is truly untestable then we should try to find a way to exclude the untestable portions from the coverage results, because otherwise it's easy to miss when we're not testing something that we should be testing.

haraldh:
|                   |        |
|-------------------|-------:|
| src               | 82.08% |
| exec-wasmtime | 57.94% |
| sallyport     | 84.16% |
| shim-kvm      |  4.56% |
| shim-sgx      | 14.13% |

haraldh:
With #1899 I marked untestable regions... other functions/methods could be tested with mocked objects in `#[test]` functions.

haraldh:
@jarkkojs could mark those regions as well for the `shim-sgx`.

jarkkojs:
> @jarkkojs could mark those regions as well for the `shim-sgx`.

`#![cfg_attr(coverage, no_coverage)]` to the beginning of the file?

haraldh:
no, per function.   see https://github.com/enarx/enarx/pull/1901

bstrie:
@jarkkojs keeping in mind that we only want to use that approach on functions that *cannot* be tested, and not just functions that we don't currently test.

PR #1899, opened 2022-05-19
chore(shim-kvm): mark coverage regions to exclude

haraldh:
Since `#[no_coverage]` is unstable, it is recommended to use it together with `cfg(coverage)` or `cfg(coverage_nightly)` set by `cargo-llvm-cov`.

See https://github.com/taiki-e/cargo-llvm-cov/#exclude-function-from-coverage

Related: https://github.com/enarx/enarx/issues/1894

### DIFF
--- a/crates/shim-kvm/src/debug.rs
+++ b/crates/shim-kvm/src/debug.rs
@@ -2,6 +2,8 @@
 
 //! Debug functions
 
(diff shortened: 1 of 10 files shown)


haraldh:
https://github.com/enarx/enarx/pull/1901

Issue #1916, opened 2022-05-24
[Security]: `shim-sgx` doesn't validate pointer to the sallyport block


Freax13:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

The pointer to the sallyport block passed in from the host is never validated: https://github.com/enarx/enarx/blob/83751fccb04e23912fc2d8f82dfdc4adf432d608/crates/shim-sgx/src/main.rs#L223

A malicious host could pass in a pointer that points to the enclave's memory, so that when the shim writes to the sallyport block it corrupts its own memory.

### Expected Behaviour

The sgx shim should validate that using the passed in pointer doesn't cause memory corruption or disclose any memory when used.

### Environment Information

```
Linux nuc 5.16.0-sgx4 #1 SMP PREEMPT Thu Apr 14 13:55:05 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
enarx 0.5.0
Enarx version 0.5.0
System Info: Linux 5.16.0-sgx4 #1 SMP PREEMPT Thu Apr 14 13:55:05 UTC 2022 x86_64
✔ Backend: sgx
  ✔ Driver: /dev/sgx_enclave
  ✔ AESM Daemon Socket: /var/run/aesmd/aesm.socket
  ✔ CPU: Intel(R) Pentium(R) Silver J5040 CPU @ 2.00GHz | GenuineIntel
  ✔  SGX Support
  ✔   Version 1
  ✔   Version 2
  ✔   FLC Support
  ✔   Max Size (32-bit): 2 GiB
  ✔   Max Size (64-bit): 64 GiB
  ✔   MiscSelect: EXINFO
  ✔   Features: DEBUG | MODE64BIT | PROVISIONING_KEY | EINIT_KEY
  ✔   Xfrm: X87 | SSE | BNDREG | BNDCSR
  ✔   EPC Size: 94 MiB
✗ Backend: sev
  ✗ Driver: /dev/sev
  ✗  SEV-SNP is enabled in host kernel
  ✗  /dev/sev is readable by user
  ✗  /dev/sev is writable by user
  ✔ Driver: /dev/kvm
  ✗  API Version
  ✔ MEMLOCK rlimit allows for: ~12 keeps (soft limit = 67108864 bytes, hard limit = 67108864 bytes)
[...]

PR #1918, opened 2022-05-24
fix(shim-sgx): check that sallyport block is outside shim

jarkkojs:
Reported-by: : Tom Dohrmann <erbse.13@gmx.de>


bstrie:
cc @Freax13 , do you see any problems with this fix for #1916?

crates/shim-sgx/src/main.rs
```diff
@@ -249,15 +250,26 @@ pub unsafe extern "sysv64" fn _start() -> ! {
 }
 
 unsafe extern "C" fn main(
-    port: &mut [usize; BLOCK_SIZE / core::mem::size_of::<usize>()],
+    block: &mut [usize; BLOCK_SIZE / core::mem::size_of::<usize>()],
     ssas: &mut [StateSaveArea; 3],
     cssa: usize,
 ) {
     ssas[cssa].extra[0] = 1; // Enable exceptions
+                             //
```
bstrie, line 258: Unnecessary line comment.

crates/shim-sgx/src/main.rs
```diff
@@ -249,15 +250,26 @@ pub unsafe extern "sysv64" fn _start() -> ! {
...
+    let block_start = block.as_ptr() as usize;
+    let block_end = block_start + BLOCK_SIZE;
+    let shim_start = shim_address();
+    let shim_end = shim_start + ENCL_SIZE;
+
+    if (block_start >= shim_start && block_start < shim_end)
```
bstrie, line 264: Let's leave a comment here explaining why it's important for security to verify the block like this.

crates/shim-sgx/src/main.rs
jarkkojs: +1

crates/shim-sgx/src/main.rs
jarkkojs: +1

PR #1928, opened 2022-05-25
build(release): extend release workflow to build Windows MSI installer

platten:


platten:
@bstrie @nickvidal we might want to look at including some branding for the installer (icons, colors, etc). While that all goes to waste if you're using `winget`, there might be an audience for interactive GUI installs.

bstrie:
Icons are easy enough to make, but I'd leave that to a future PR. File an issue for it.

bstrie:
Do these files need to live in a `wix` top-level directory? I'm already frustrated at how cluttered the repo is for random things like Nix, as it makes for bad contributor UX to have so many random things lying around that almost no contributors will need to care about.

platten:
> Do these files need to live in a `wix` top-level directory? I'm already frustrated at how cluttered the repo is for random things like Nix, as it makes for bad contributor UX to have so many random things lying around that almost no contributors will need to care about.

@bstrie We will need a place for release assets and putting the WiX-specific assets somewhere. Thoughts around putting the artifacts in `release/installers/windows/*`?

bstrie:
@platten A `/release` directory sounds fine for now, and maybe we'll be able to shift more things into it in the future. I just like want the repo to be organized and self-explanatory.

platten:
> @platten A `/release` directory sounds fine for now, and maybe we'll be able to shift more things into it in the future. I just like want the repo to be organized and self-explanatory.

@bstrie Addressed. Please take another look.

bstrie (approved):

PR #1931, opened 2022-05-26
build(arm64): add ARM64 build release target

platten:


platten:
This PR is on hold until we figure out how to have Nix not "break" on a self-hosted GH runner with each subsequent run. Would appreciate @rvolosatovs insight when he gets back.

PR #1935, opened 2022-05-31
build(release): add post v0.5.1 release improvements

platten:
### Improvements:
- Add the ability to do a "FAST" release using just  a single test image using the "FAST" environment variable
- Added the ability to skip tests using the "SKIP_TESTS" environment variable
- Ensure that the repository is not deleted when "DRY_RUN" is set to false

### Fixes:
- Changed commit message to be compliant with conventional commits
- Added a pause between crate publish runs to avoid race condition
- Remove the non-functioning DR_PUBLISH block (since there are interdependencies between crates, this will never work).

### DIFF
--- a/helper/release_enarx.sh
+++ b/helper/release_enarx.sh
@@ -18,7 +18,8 @@ readonly DEPS=(crt0stack flagset iocuddle lset mmarinus mmledger nbytes noted pr
 
 # Constants for release testing container images
 readonly IMAGE_PREFIX="registry.gitlab.com/enarx/misc-testing/"
-readonly IMAGES=(ubuntu debian centos7 centos8 fedora)
+readonly IMAGES_THOROUGH=(ubuntu debian centos7 centos8 fedora)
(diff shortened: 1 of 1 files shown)


helper/release_enarx.sh
```diff
@@ -343,10 +339,20 @@ on_exit() {
 }
 trap on_exit EXIT
 
+
+if [[ "$FAST" == "true" ]]; then
+    echo -e "Warning: FAST environment variable set, will only test with single image."
+    IMAGES=( "${IMAGES_FAST[@]}" )
+else
+    shout "THOROUGH environment variable set, will test with all images. This will take up to 90 minutes for the entire process"
```
bstrie, line 347: Better to say "FAST environment variable not set", since this might encourage someone to try setting a THOROUGH environment variable which doesn't get checked at all.

platten: Fixed

bstrie (dismissed):

bstrie (approved):

PR #1939, opened 2022-06-01
build(toolchain): bump Rust to 2022-05-28

bstrie:
Also remove the rust-src component, which is no longer necessary now that build-std is no longer being used.


bstrie:
Closing for now, there are some nightly clippy regressions that are waiting on https://github.com/rust-lang/rust-clippy/pull/8869 to land and we're not upgrading the toolchain for any urgent reason.

Issue #1459, opened 2022-02-25
[Infrastructure]: Failed to create SGX enclave and caused by Input/output error


rwwwwww21:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Description

Hi ~
I ran the helloworld.wasm in my enarx and got the error:
```
Error: Failed to create SGX enclave

Caused by:
    Input/output error (os error 5)
```

My machine is pc and only support SGX1 without FLC, so I applied the [patch](https://github.com/haosanzi/inclavare-containers/commit/f6a272443d93971784d8148476155a9f6eb08ddc) to my 5.13 linux kernel and appeared the `sgx_enclave` driver and ran the [Intel SampleEnclave](https://github.com/intel/linux-sgx/tree/master/SampleCode/SampleEnclave) successfully. But when I ran the enarx, encountered the following error, here is the `strace` information:
```
execve("./target/debug/enarx", ["./target/debug/enarx", "run", "../rww/enarx-test/hello-world/ta"...], 0x7ffca85f9380 /* 70 vars */) = 0
brk(NULL)                               = 0x561dbfaf8000
arch_prctl(0x3001 /* ARCH_??? */, 0x7ffc61af9680) = -1 EINVAL (Invalid argument)
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=77724, ...}) = 0
mmap(NULL, 77724, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f8a570c7000
close(3)                                = 0
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libssl.so.1.1", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0p\367\1\0\0\0\0\0"..., 832) = 832
fstat(3, {st_mode=S_IFREG|0644, st_size=598104, ...}) = 0
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f8a570c5000
[...]

PR #1974, opened 2022-06-23
fix(sgx): revert the use of AVX2 and AVX512

jarkkojs:
Fixes: 55caa95159c9 ("feat(shim-sgx): enable AVX2 and AVX512 usage in the enclave")
Closes: #1459


jarkkojs:
@platten ping

npmccallum (approved):

Issue #1953, opened 2022-06-07
CLI API for Drawbridge integration


bstrie:
# Acceptance Criteria:
- [x] Draft user markdown user documentation
- [ ] Develop subcommand
- [ ] Integration / E2E testing

# Proposed API

Adding Drawbridge integration to the CLI will also involve changes to the existing CLI API. Here is a proposed overview of what the final CLI will look like:

```
enarx
├── run [--backend <BACKEND>] [--config <CONFIG>] <PACKAGE>
├── platform
│   ├── info [--format <FORMAT>]
│   ├── sgx
│   │   └── register
│   └── snp
│       ├── info [--file]
│       └── update [--path <PATH>]
├── user
│   ├── info [--format <FORMAT>] <USER_ID>
│   ├── login [--oidc-domain <URL>] [--oidc-client-id <ID>] [--insecure]
│   ├── logout
│   └── register <USER_ID>
├── repo
│   ├── info [--format <FORMAT>] <REPO_ID>
│   ├── register --public|--private <REPO_ID>
│   ├── search [--format] [--host <HOST>] <QUERY>
│   ├── token
│   │   ├── info [--format <FORMAT>] <REPO_ID>
│   │   ├── generate <REPO_ID> <REPO_TOKEN_NAME>
│   │   └── revoke <REPO_ID> <REPO_TOKEN_NAME>
│   └── yank [--undo] <REPO_ID>
├── package
│   ├── info [--format <FORMAT>] <PACKAGE_ID>
│   ├── fetch <PACKAGE_ID>
│   ├── publish <PACKAGE_ID> <FILES>
│   └── yank [--undo] <PACKAGE_ID>
├── tree
│   ├── info [--format <FORMAT>] [--host <HOST>] <DIGEST>
│   ├── fetch [--host <HOST>] <DIGEST>
│   └── digest <FILES>
└── unstable
    └── exec [--backend <BACKEND>] <BINARY>
```

Changes to the existing CLI API:

* `enarx info`, `enarx sgx`, and `enarx snp` are moved into the new `enarx cpu` subcommand, becoming `enarx cpu info`, `enarx cpu sgx`, and `enarx cpu snp`.
* `enarx snp vcek show` becomes `enarx cpu snp info`, and `enarx snp vcek show-file` becomes `enarx cpu snp info --file`.
[...]

PR #1976, opened 2022-06-24
feat(cli): restructure CLI into new hierarchy

bstrie:
Implements the CLI restructuring proposed in https://github.com/enarx/enarx/issues/1953 . This PR doesn't add any new functionality, it only moves existing commands to new places in the CLI and adds stub implementations of the new commands that have yet to be implemented (these stub implementations are hidden from the help output so that users do not accidentally attempt to use them; commands will be individually un-hidden as they are implemented). This also makes individual command handling more regular by moving functionality out of main.rs and into dedicated files for each subcommand, which is important for keeping things manageable since we're planning to add a lot of new commands.

To reiterate, don't be dismayed by the size of this PR, as most of the code here is code that already exists and is just being moved into new locations for better organization (and the rest of the new code is largely provisional for now), so it probably only needs to be lightly reviewed.

(Note also that this does not merge `enarx run` and `enarx deploy` as proposed in #1953; I figure that's a large enough change that it's better left for later.)


src/backend/sev/mod.rs
```diff
@@ -24,7 +24,7 @@ use sallyport::item::{self, Item};
 mod builder;
 mod cpuid_page;
 mod data;
-mod snp;
+pub mod snp;
```
rvolosatovs, line 27: See https://enarx.dev/docs/Contributing/Coding-Style#use-declarations

rvolosatovs, line 27: see https://enarx.dev/docs/Contributing/Coding-Style#use-declarations

rvolosatovs:
I will do a proper review once this is done!

bstrie:
From this point forward I'll push any changes as new commits, to make it easier to review. I'll squash those once this is ready for approval.

src/backend/sev/mod.rs
bstrie: Done.

rvolosatovs (approved):
Went through the PR on a call

PR #1993, opened 2022-06-27
fix(deps): update linked_list_allocator requirement from 0.9.1 to 0.10.1

haraldh:
This requires switching from `RwLock` to `Mutex`, because the inner Allocator is not `Sync`.

Supercedes: https://github.com/enarx/enarx/pull/1992

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1911,9 +1911,9 @@ checksum = "33a33a362ce288760ec6a508b94caaec573ae7d3bbbd91b87aa0bad4456839db"
 
 [[package]]
 name = "linked_list_allocator"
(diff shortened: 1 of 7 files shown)


haraldh:
```
panicked at 'Freed node aliases existing hole! Bad free?', /home/harald/.cargo/registry/src/github.com-1ecc6299db9ec823/linked_list_allocator-0.10.0/src/hole.rs:439:17
```

Seems like the new allocator version is buggy, because I can't see any double free.

haraldh:
Upstream bug and fix: https://github.com/rust-osdev/linked-list-allocator/pull/63

PR #2002, opened 2022-06-28
feat(cli): user login CLI

bstrie:
Not ready for review, waiting on keyring updates.


rvolosatovs (changes requested):
We should use device code flow here, see https://github.com/rvolosatovs/oidc-login

bstrie:
Hm, cargo-deny is revealing that there's a CVE in a deep transitive dependency of the crate that I'm pulling in for keyring storage. I'll need to see whether the maintainers are responsive and think about the best way to address this.

.github/workflows/nix.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       with:
         name: enarx
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-    - run: nix run -L . info
+    - run: nix run -L . cpu info
```
rvolosatovs, line 45: Does this belong in this PR/commit? What's the reasoning here?

rvolosatovs (dismissed):

.github/workflows/nix.yml
bstrie: I saw that this test was failing due to the lack of this change, and rather than file a whole new PR for this I just stuffed it in here. Do you want me to extract it into its own PR?

bstrie: I've made it into its own commit.

src/cli/user/login.rs
```diff
@@ -0,0 +1,65 @@
...
+use oauth2::basic::BasicClient;
+use oauth2::devicecode::StandardDeviceAuthorizationResponse;
+use oauth2::ureq::http_client;
+use oauth2::url::Url;
+use oauth2::{AuthType, AuthUrl, ClientId, DeviceAuthorizationUrl, Scope, TokenResponse, TokenUrl};
+
+/// Log in to an Enarx package host and save credentials locally.
+#[derive(Args, Debug)]
+pub struct Options {
+    #[clap(long, default_value = "https://dev-l20dl4-w.us.auth0.com/")]
```
rvolosatovs, line 14:
```suggestion
    #[clap(long, default_value = "https://auth.profian.com/")]
```

src/cli/user/login.rs
```diff
@@ -0,0 +1,65 @@
...
+    domain: Url,
+    #[clap(long, default_value = "mhIPNbaEGEvkgqhznnWUb6vFwnfFy0dM")]
```
rvolosatovs, line 16:
```suggestion
    #[clap(long, default_value = "4NuaJxkQv8EZBeJKE56R57gKJbxrTLG2")]
```

bstrie:
Pushed updated default domain and client ID.

bstrie:
The necessary dependencies have had their new versions published and I've updated the lockfile to use them. This is ready for review.

rvolosatovs (approved):

rvolosatovs (approved):

PR #2008, opened 2022-06-29
feat(exec-wasmtime): add `addr` to `listen` in Enarx.toml

haraldh:
This adds the `addr` field to the `listen` entry in Enarx.toml, in case
somebody wants to restrict the listen address to a specific IP,
rather than the `::` IPV6 any address, which binds to all interfaces.

### DIFF
--- a/crates/exec-wasmtime/src/config.rs
+++ b/crates/exec-wasmtime/src/config.rs
@@ -6,8 +6,12 @@ use std::{collections::HashMap, ops::Deref};
 use serde::{de::Error as _, Deserialize, Deserializer};
 use url::Url;
 
-fn default_port() -> u16 {
+const fn default_port() -> u16 {
     443
+}
+
+fn default_addr() -> String {
+    "::".into()
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -105,6 +109,9 @@ pub enum File {
 
         #[serde(default)]
         prot: Protocol,
+
+        #[serde(default = "default_addr")]
+        addr: String,
     },
 
     #[serde(rename = "connect")]
@@ -188,14 +195,15 @@ mod test {
                 File::Listen {
                     name: "X".into(),
                     port: 9000,
-                    prot: Protocol::Tcp
+                    prot: Protocol::Tcp,
+                    addr: default_addr()
                 },
                 File::Stdout { name: None },
                 File::Null { name: None },
                 File::Stderr { name: None },
                 File::Connect {
                     name: None,
-                    port: 443,
+                    port: default_port(),
                     prot: Protocol::Tls,
                     host: "google.com".into(),
                 },
--- a/crates/exec-wasmtime/src/loader/compiled/mod.rs
+++ b/crates/exec-wasmtime/src/loader/compiled/mod.rs
@@ -8,8 +8,6 @@ use null::Null;
 use super::{Compiled, Connected, Loader};
 use crate::config::{File, Protocol};
 
-use std::net::Ipv4Addr;
-
 use anyhow::Result;
 use cap_std::net::{TcpListener, TcpStream};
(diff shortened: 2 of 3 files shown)


tests/wasm/mod.rs
```diff
@@ -243,6 +243,7 @@ kind = "listen"
 prot = "tcp"
 port = {cport}
 name = "LISTEN"
+addr = "0.0.0.0"
```
haraldh, line 246: The windows CI does not like IPv6

crates/exec-wasmtime/src/config.rs
```diff
@@ -105,6 +109,9 @@ pub enum File {
 
         #[serde(default)]
         prot: Protocol,
+
+        #[serde(default = "default_addr")]
```
rvolosatovs, line 113:
```suggestion
        #[serde(default_value = "::")]
```

crates/exec-wasmtime/src/config.rs
```diff
@@ -6,10 +6,14 @@ use std::{collections::HashMap, ops::Deref};
 use serde::{de::Error as _, Deserialize, Deserializer};
 use url::Url;
 
-fn default_port() -> u16 {
+const fn default_port() -> u16 {
```
rvolosatovs, line 9: I think we should just remove it and instead use `default_value` on L107 https://github.com/enarx/enarx/pull/2008/files#diff-1dffd47a20f4acb91d39396f05a61ec4b063f041570242e918b3b8ebb5631428R107

crates/exec-wasmtime/src/config.rs
```diff
@@ -188,7 +195,8 @@ mod test {
                 File::Listen {
                     name: "X".into(),
                     port: 9000,
-                    prot: Protocol::Tcp
+                    prot: Protocol::Tcp,
+                    addr: "::".to_string()
```
rvolosatovs, line 199: Shouldn't this be using the helper function if we keep it?

crates/exec-wasmtime/src/loader/compiled/mod.rs
```diff
@@ -47,13 +45,15 @@ impl Loader<Compiled> {
...
+                File::Listen {
+                    addr, port, prot, ..
+                } => {
                     let caps = FileCaps::FILESTAT_GET
                         | FileCaps::FDSTAT_SET_FLAGS
                         | FileCaps::POLL_READWRITE
                         | FileCaps::READ;
 
-                    let tcp = std::net::TcpListener::bind((Ipv4Addr::UNSPECIFIED, *port))?;
+                    let tcp = std::net::TcpListener::bind((&**addr, *port))?;
```
rvolosatovs, line 56: Is there a way we can avoid the `&**`? Perhaps a `ref addr` on L49 could help at least a bit? Is the purpose of this to make a bitwise copy? (if so, shouldn't we use a method?)

crates/exec-wasmtime/src/config.rs
rvolosatovs: My fault, confused with `clap` https://github.com/profianinc/steward/blob/ed234c346821c8eab541ee6a39d45d5198beb883/src/main.rs#L67-L70

crates/exec-wasmtime/src/loader/compiled/mod.rs
haraldh: I can use `addr.as_str()`

crates/exec-wasmtime/src/config.rs
haraldh: why?  ... changed it to `"::1"`

haraldh: yes

rvolosatovs (approved):

PR #2010, opened 2022-06-29
build: use new Cargo sparse registry support

bstrie:
The registry is the list of crate versions from crates.io, which Cargo uses to resolve dependency versions. As originally implemented, this registry is stored as a Git repo, which is cloned on first use and pulled for updates. However, as crates.io has grown the size of the registry has become a bottleneck, and Cargo is preparing to address this by supporting "sparse registries", which involves only pulling dependency information for crates that are actually being used for the compilation. This promises to be somewhat faster for regular development and much faster for CI builds.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -11,3 +11,4 @@ rustflags = ["-C", "target-feature=+crt-static"]
 
 [unstable]
 bindeps = true
+sparse-registry = true


bstrie:
@haraldh for some reason the automerge job was skipped. Can you merge this manually?

PR #2020, opened 2022-07-04
feat: flesh out fundamental drawbridge CLI

bstrie:


src/cli/package/info.rs
```diff
@@ -1,13 +1,29 @@
...
 impl Options {
     pub fn execute(self) -> anyhow::Result<()> {
-        unimplemented!()
+        let cl = client(&self.spec.host, &self.insecure_auth_token_file)
+            .context("Failed to build client")?;
+        let tag = cl.tag(&self.spec.ctx);
+        let _tag_entry = tag
+            .get()
+            .context("Failed to retrieve package information")?;
+        // TODO: make this actually useful
```
rvolosatovs, line 24: Why not just print the JSON directly and add text representation later?
JSON can be used for scripting and `serde_json` should take care of the pretty-printing.

src/cli/repo/info.rs
```diff
@@ -1,13 +1,33 @@
...
+        let record = repo
+            .get()
+            .context("Failed to retrieve repository information")?;
+        let visibility = if record.public { "public" } else { "private" };
+        println!("Visibility: {visibility}");
+        let tags = repo.tags().context("Failed to retrieve repository tags")?;
+        println!("Tags:");
+        for tag in tags {
+            println!("\t{tag}");
+        }
```
rvolosatovs, line 30: It's nice to have text output "for humans", but IMO it's more important to output something that can by used for scripting (pretty-printed JSON)
How hard would it be to add `--format text|json` parameter to these commands? We don't necessarily have to implement everything already and some inconsistency should be fine for now.

src/cli/user/info.rs
```diff
@@ -1,13 +1,28 @@
...
 impl Options {
     pub fn execute(self) -> anyhow::Result<()> {
-        unimplemented!()
+        let cl = client(&self.spec.host, &self.insecure_auth_token_file)
+            .context("Failed to build client")?;
+        let user = cl.user(&self.spec.ctx);
+        let record = user
+            .get()
+            .with_context(|| format!("Failed to get record for user: {}", self.spec.ctx.name))?;
+        // TODO: make this useful
```
rvolosatovs, line 24: JSON would be more useful here and easier to maintain going forward

src/cli/user/login.rs
```diff
@@ -58,7 +66,7 @@ impl Options {
 
         // TODO: graceful timeout, so that users are not forced to Ctrl+C if the server errors
         let secret = res.access_token().secret();
-        keyring::Entry::new("drawbridge", "enarx")
+        keyring::Entry::new("enarx", "drawbridge")
```
rvolosatovs, line 69:
```suggestion
        keyring::Entry::new("enarx", oidc_domain)
```
Assuming arbitrary strings are allowed here

src/cli/user/register.rs
```diff
@@ -0,0 +1,60 @@
...
+use drawbridge_client::types::UserRecord;
+use drawbridge_client::Url;
+use oauth2::AccessToken;
+use openidconnect::core::{CoreClient, CoreProviderMetadata, CoreUserInfoClaims};
+use openidconnect::ureq::http_client;
+use openidconnect::{ClientId, IssuerUrl};
+
+/// Register a new user account with a package host.
+#[derive(Args, Debug)]
+pub struct Options {
```
rvolosatovs, line 16: We should really have `$XDG_CONFIG_HOME/enarx/conf.toml`, where `oidc_domain` and `oidc_client_id` would be configured.

src/drawbridge.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let url = format!("https://{host}");
+
+    let mut cl = Client::builder(
+        url.parse()
+            .with_context(|| format!("Failed to parse URL: {url}"))?,
+    );
+
+    // For local development and integration testing
+    if host.starts_with("localhost:") {
```
rvolosatovs, line 88: I don't think this belongs in production code

tests/client/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+        enarx(format!($($arg)+))
+    )
+}
+
+#[async_std::test]
+async fn full() {
+    run(|oidc_addr, db_addr| {
+        // test for failure when looking up a user that does not exist
+        let cmd = cmd!(
+            "enarx user info --insecure-auth-token-file tests/client/token {db_addr}+testuser"
```
rvolosatovs, line 19: Why `+` and not just `{db_addr}/testuser`? That aligns better with existing tooling, e.g. `docker` (same for all other occurences)

src/cli/package/info.rs
```diff
@@ -1,13 +1,29 @@
...
+
+use anyhow::Context;
 use clap::Args;
 
 /// Retrieve information about a published package.
 #[derive(Args, Debug)]
-pub struct Options {}
+pub struct Options {
+    #[clap(long)]
+    insecure_auth_token_file: Option<String>,
```
rvolosatovs, line 12: So on `login` the token is stored in the keychain, but to actually use it, users have to manually retrieve it from the keychain and write it to a file?
I think this should be an environment variable and just a plaintext token there is fine IMO. (e.g. `ENARX_AUTH_TOKEN`) It would be very painful to specify this parameter for every `enarx` invocation

Cargo.toml
```diff
@@ -26,13 +26,17 @@ bitflags = { version = "1.2", default-features = false }
 clap = { version = "3.1", features = ["env", "derive", "std"], default-features = false }
 colorful = { version = "0.2", default-features = false }
 dirs = { version = "4.0", default-features = false }
+drawbridge-client = { git = "https://github.com/profianinc/drawbridge", rev = "d8e0befbd8da6b9f357c396273996e6141f7569f", default-features = false }
```
rvolosatovs, line 29: We should release Drawbridge crates, I'll take care of that

rvolosatovs (commented):
Good start!

src/drawbridge.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use std::fs::read_to_string;
+use std::str::FromStr;
+
+use anyhow::Context;
+use drawbridge_client::types::{RepositoryContext, TagContext, UserContext};
+use drawbridge_client::Client;
+use rustls::{Certificate, RootCertStore};
+
+const DEFAULT_HOST: &str = "example.com";
```
rvolosatovs, line 11: What is this for and why would we want that in production code?

src/drawbridge.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (host, ctx) = parse_slug(s).with_context(|| format!("Failed to parse slug: {s}"))?;
+        Ok(Self { host, ctx })
+    }
+}
+
+fn parse_slug<T: FromStr>(s: &str) -> Result<(String, T), <T as FromStr>::Err> {
+    let (host, name) = match s.split_once('+') {
+        Some((host, name)) => (host, name),
+        None => (DEFAULT_HOST, s),
```
rvolosatovs, line 61: This should default to `drawbridge.profian.com`

src/cli/package/publish.rs
```diff
@@ -1,13 +1,31 @@
...
+    path: String,
+}
 
 impl Options {
     pub fn execute(self) -> anyhow::Result<()> {
-        unimplemented!()
+        let cl = client(&self.spec.host, &self.insecure_auth_token_file)
+            .context("Failed to build client")?;
+        let tag = cl.tag(&self.spec.ctx);
+        // why does create_from_path_unsigned return a bool? what is this useful for?
```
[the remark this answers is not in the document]
bstrie: So a return value of Ok(false) means that the user has uploaded a new tag with the exact same contents as an old tag, whereas Err means they've attempted to upload a tag that already exists?

src/cli/user/register.rs
bstrie: I've filed https://github.com/enarx/enarx/issues/2021 . We can pursue this for a future release.

src/drawbridge.rs
bstrie: What solution do you propose? Integration tests compile the binary in normal release mode, there's no conditional compilation flags to employ.

tests/client/mod.rs
bstrie: I also prefer that, I just didn't feel like duplicating all the slug parsing code on the Enarx side, since that's required to support this. I can do that.

src/cli/package/info.rs
bstrie: Sure, although environment variables are quite a bit more insecure than even normal files are, and also now the insecure behavior is entirely implicit. Maybe we just deal with the pain for now and prioritize getting config files up?

Cargo.toml
bstrie: Sure, although note that they all need explicit licenses in their Cargo.tomls.

src/drawbridge.rs
bstrie: This is just the default drawbridge, which we certainly want to support, but I don't the actual address of our production drawbridge instance.

src/cli/package/publish.rs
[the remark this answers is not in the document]
rvolosatovs: Correct

src/cli/package/info.rs
rvolosatovs: Environment variables are still the only thing e.g. a Github Action CI job would have.
Sure, it a job could write the value of the env var to a file and then use it by path everywhere, but that's terrible UX

tests/client/package/tEst-file..__.foo.42
```diff
@@ -0,0 +1 @@
+invalidext
```
rvolosatovs, line 1: Use https://github.com/enarx/enarx/tree/main/examples/tcp_server directly instead

tests/client/certs/ca.conf
```diff
@@ -0,0 +1,16 @@
+[req]
```
rvolosatovs, line 1: These belong in `testdata` for consistency with other repos

src/drawbridge.rs
bstrie: I've addressed this by making it a CLI parameter that can be set via environment variable.

Issue #2017, opened 2022-06-30
[Feature]: Extract Enarx.toml handling into separate crate


platten:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

It would be helpful to have Enarx.toml parsing/handling & documentation to be packaged as a sub-crate.

### Acceptance Criteria

- Extracted code
- Published subcrate crate
- Updated documentation
- Updated Cargo.toml to use Enarx.toml subcrate

haraldh:
first version: https://github.com/haraldh/enarx_config

@rvolosatovs @bstrie @npmccallum @rjzak @definitelynobody if you want to have a look, before it gets imported?

PR #2022, opened 2022-07-05
feat: add enarx-config crate

haraldh:
Fixes: https://github.com/enarx/enarx/issues/2017


crates/enarx-config/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
rvolosatovs, line 1: Do we need a separate LICENSE in the subcrate or can we reuse the one from the root of the workspace?
@bstrie you mentioned we'd need explicit licenses in https://github.com/enarx/enarx/pull/2020#discussion_r913161527, does this mean we also need to have LICENSE files in each subcrate?

crates/enarx-config/src/lib.rs
```diff
@@ -1,5 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
-//! The Enarx Configuration file format
+
+//! Configuration for a WASI application in an Enarx Keep
```
rvolosatovs, line 3: What was wrong with the original comment?
Do we need to call workloads "WASI applications"? Why not just Enarx workload or Enarx Keep configuration?

crates/enarx-config/src/lib.rs
```diff
@@ -15,6 +22,9 @@ fn default_addr() -> String {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+/// Name of a file descriptor
```
rvolosatovs, line 25: Named assigned to a file descriptor

crates/enarx-config/src/lib.rs
```diff
@@ -15,6 +22,9 @@ fn default_addr() -> String {
...
+///
+/// This is used to export a list of file descriptor names in the `FD_NAMES` environment variable.
```
rvolosatovs, line 27: This lacks context on what `FD_NAMES` actually looks like

crates/enarx-config/src/lib.rs
```diff
@@ -45,30 +55,52 @@ impl<'de> Deserialize<'de> for FileName {
         let name = String::deserialize(deserializer)?;
 
         if name.contains(':') {
-            return Err(D::Error::custom("invalid name"));
+            return Err(D::Error::custom("invalid value for `name` contains ':'"));
```
rvolosatovs, line 58: It looks like `contains ':'` should be the cause for the outermost error.
Normally I'd expect the caller to do something like:
```rust
let name = v.deserialize(de).context("invalid file name")?
````

File: crates/enarx-config/src/lib.rs [hunk 6]
```diff
@@ -45,30 +55,52 @@ impl<'de> Deserialize<'de> for FileName {
...
     pub files: Vec<File>,
 
+    /// An optional Steward URL
     #[serde(default)]
-    /// Optional Steward URL
     pub steward: Option<Url>,
 }
 
-impl Default for Config {
+impl Default for EnarxConfig {
```
rvolosatovs, line 103: Why the `Enarx` prefix? We're already in `enarx-config` crate, so it could not be any other config.

crates/enarx-config/src/lib.rs
```diff
@@ -85,50 +117,78 @@ impl Default for Config {
...
+        /// protocol to use
         #[serde(default)]
         prot: Protocol,
-
-        host: String,
     },
 }
 
 impl File {
+    /// get the name for a file descriptor
```
rvolosatovs, line 191: Why lowercase?

crates/exec-wasmtime/src/loader/mod.rs
```diff
@@ -16,8 +16,8 @@ mod connected;
 mod pki;
 mod requested;
 
-use super::config::Config;
 use super::{Args, Package};
+use enarx_config::EnarxConfig;
```
rvolosatovs, line 20:
```suggestion
use enarx_config::Config;
```

Would be way nicer to use

haraldh: I knew you would say this

crates/enarx-config/src/lib.rs
haraldh: Hmm, because it's only relevant for exec-wasmtime

crates/enarx-config/src/lib.rs
haraldh: How about
```suggestion
            return Err(D::Error::custom("`name` contains ':'"));
```

crates/enarx-config/src/lib.rs
```diff
@@ -85,50 +117,78 @@ impl Default for Config {
     }
 }
 
+/// Parameters for a pre-opened file descriptor
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
 #[serde(tag = "kind")]
 pub enum File {
+    /// file descriptor to `/dev/null`
```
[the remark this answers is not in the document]
haraldh: right, right

crates/enarx-config/src/lib.rs
haraldh: :shrug:

crates/enarx-config/src/lib.rs
haraldh: fixed

crates/enarx-config/src/lib.rs
haraldh: fixed

crates/enarx-config/src/lib.rs
haraldh: fixed

crates/enarx-config/src/lib.rs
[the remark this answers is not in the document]
haraldh: fixed

crates/exec-wasmtime/src/loader/mod.rs
haraldh: fixed

npmccallum:
LGTM when @rvolosatovs is happy.

rvolosatovs (approved):

PR #2023, opened 2022-07-05
fix(tcp_server): correct target_os

haraldh:
It's `target_os="wasi"` and not `target_os="wasm32-wasi"`.


bstrie (approved):

rvolosatovs (approved):
LGTM, but shouldn't this have been picked up by CI? Why is there no job building all examples to `wasm32-wasi`?

PR #2036, opened 2022-07-07
feat: add Serialize to enarx-config

haraldh:


rvolosatovs (approved):
We probably want to guard this by a feature going forward, but LGTM for now

Issue #2021, opened 2022-07-04
Support configuration files in the CLI


bstrie:
Unanswered questions: what to name the file? "Enarx.toml" is already taken.

PR #2041, opened 2022-07-11
feat: support custom credential helpers

### LINKED ISSUE #2019: [Bug]: Support storing access token on Linux systems without GNOME keyring

### Code of Conduct

### Current Behaviour

On a Linux distribution without [GNOME Keyring](https://wiki.gnome.org/Projects/GnomeKeyring/) running on D-Bus:
```
Caused by:
    0: Platform secure storage failure: zbus error: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.secrets was not provided by any .service files
    1: zbus error: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.secrets was not provided by any .service files
    2: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.secrets was not provided by any .service files
```

### Expected Behaviour

Success with an interactive prompt to permit insecure credential storage.
We should also have an `--insecure` flag, which would store the secret in `$XDG_CONFIG_HOME/enarx/tokens/<issuer-url>` with `0600` permissions bypassing the keyring and the prompt

### Environment Information

2e1852c71599f2adc1ca7d1fe4e8f1c90cf0b0dd

### Steps To Reproduce

_No response_

rvolosatovs:
Closes #2019 

This should really be configured via CLI config #2021 

E.g. for https://github.com/gopasspw/gopass this can be used the following way:

`enarx-credential-helper-gopass`
```
#!/bin/sh
set -e
if [ "${1}" = "insert" ]; then
    exec gopass insert -f misc/enarx
elif [ "${1}" = "show" ]; then
    gopass find misc/enarx 2>/dev/null
    exec gopass show -n -o misc/enarx
else
    echo "Unknown command '${1}'"
    exit 1
fi
```

`enarx user register myuser --credential-helper ./enarx-credential-helper-gopass`

Once we add the plugin system for `enarx`, this will can also be looked up from `PATH`, i.e. users could contribute and install custom `enarx-credential-helper-*` and these would be picked up from the path.

This is a standard approach used by a variety of other tools, e.g.:
- https://github.com/docker/docker-credential-helpers
- https://git-scm.com/docs/gitcredentials

This also fixes a bug, where credentials for different providers would be stored in the same namespace (`"oidc_domain"`, which clearly should have been a variable)

### DIFF
--- a/src/cli/repo/info.rs
+++ b/src/cli/repo/info.rs
@@ -8,6 +8,7 @@ use anyhow::Context;
 use camino::Utf8PathBuf;
 use clap::Args;
 use drawbridge_client::types::{RepositoryConfig, TagName};
+use oauth2::url::Url;
 use serde::Serialize;
 
 #[derive(Serialize)]
@@ -21,6 +22,8 @@ struct RepoInfo {
 pub struct Options {
     #[clap(long, env = "ENARX_CA_BUNDLE")]
     ca_bundle: Option<Utf8PathBuf>,
+    #[clap(long, default_value = "https://auth.profian.com/")]
+    oidc_domain: Url,
     #[clap(long, env = "ENARX_INSECURE_AUTH_TOKEN")]
     insecure_auth_token: Option<String>,
     #[clap(long, env = "ENARX_CREDENTIAL_HELPER")]
@@ -32,6 +35,7 @@ impl Options {
     pub fn execute(self) -> anyhow::Result<()> {
         let cl = client(
             &self.spec.host,
+            &self.oidc_domain,
             &self.insecure_auth_token,
             &self.ca_bundle,
             &self.credential_helper,
--- a/src/cli/repo/info.rs
+++ b/src/cli/repo/info.rs
@@ -2,6 +2,8 @@
 
 use crate::drawbridge::{client, RepoSpec};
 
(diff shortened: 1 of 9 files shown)


haraldh (changes requested):
LGTM, but your initial comment should be in some markdown document.

src/cli/repo/info.rs
```diff
@@ -19,14 +22,24 @@ struct RepoInfo {
 pub struct Options {
     #[clap(long, env = "ENARX_CA_BUNDLE")]
     ca_bundle: Option<Utf8PathBuf>,
+    #[clap(long, default_value = "https://auth.profian.com/")]
+    oidc_domain: Url,
     #[clap(long, env = "ENARX_INSECURE_AUTH_TOKEN")]
     insecure_auth_token: Option<String>,
+    #[clap(long, env = "ENARX_CREDENTIAL_HELPER")]
+    credential_helper: Option<OsString>,
```
haraldh, line 30: These elements look like they could be refactored.

rvolosatovs: Yes, it should be in #2021 , which does not belong in this PR.
Adding these was a massive pain...

Issue #2045, opened 2022-07-12
[Feature]: Make SEV VCEK path configurable


rvolosatovs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

There should be a way to configure the VCEK path in #2021

PR #2049, opened 2022-07-12
feat(sev): remove vcek cache directory in `$HOME`

haraldh:
If there is no vcek certificate cached in `/var/cache/amd-sev`, fail.

There must be a system service setting up the cache file.
`enarx platform snp update` does exactly that.

This removes all confusion about which file is used and when it should be created.

Fixes: https://github.com/enarx/enarx/issues/2045


rvolosatovs (approved):
This could benefit from some refactoring now that the logic changed, but we can always address that later

PR #2054, opened 2022-07-14
chore(release): release v0.6.1

platten:
chore(release): release v0.6.1

PR #2058, opened 2022-07-14
revert: disable sparse registry to address cargo publish issues

platten:

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -11,4 +11,4 @@ rustflags = ["-C", "target-feature=+crt-static"]
 
 [unstable]
 bindeps = true
-sparse-registry = true
+# sparse-registry = true


.cargo/config
```diff
@@ -11,4 +11,4 @@ rustflags = ["-C", "target-feature=+crt-static"]
 
 [unstable]
 bindeps = true
-sparse-registry = true
+# sparse-registry = true
```
rvolosatovs, line 14: As already mentioned in https://github.com/enarx/enarx/pull/2054#discussion_r921169839, please document the reasoning for this change in the commit.

The line itself should be either removed completely, or commented-out with an issue reference to reenable it in the future

rvolosatovs (changes requested):

PR #2061, opened 2022-07-15
ci: fix toolchain

rvolosatovs:
Alternative soluton to #2060 , which should prevent such situations from occuring in the future by relying on all packages locked in a lock file and cached via `nix`.
That also means that we can manage our rust toolchain version and components in one standard location (`rust-toolchain.toml`)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -14,6 +14,10 @@ jobs:
   test:
     name: ${{ matrix.crate.name }} coverage
     runs-on: ${{ matrix.crate.host }}
+    container:
+      image: nixos/nix
+      volumes:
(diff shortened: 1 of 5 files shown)


rvolosatovs:
This would not work in existing self-hosted runners out-of-the-box. It would require one of the following changes:
1. Getting `docker` (aliased to podman potentially) available on self-hosted runners
2. Getting `nix` installed by default on self-hosted runners.
3. Set up `USER` env variable on self-hosted runners, which may let us use the existing `install-nix-action`, refs https://github.com/cachix/install-nix-action/issues/98
4. Don't know if `podman` is available already, but we could also run containers via `podman` directly https://github.com/actions/runner/issues/505

PR #2082, opened 2022-07-21
feat(sgx): add ENARX_TEST_SGX_KEY_FILE env variable

haraldh:
By specifying a precomputed PEM private RSA key file, a lot of CI time generating RSA keys can be saved:

```
❯ export ENARX_TEST_SGX_KEY_FILE=$(pwd)/tests/sgx-test.key
❯ time cargo test --test integration
[…]
real	0m46.452s
user	0m33.920s
sys	0m11.816s

❯ unset ENARX_TEST_SGX_KEY_FILE
❯ time cargo test --test integration
[…]
real	2m39.912s
user	2m27.618s
sys	0m11.786s
```

Removed the need to manually set `ENARX_TEST_SGX_KEY_FILE` after this test.
`cargo test` does this automatically now.

Fixes: https://github.com/enarx/enarx/issues/2083


bstrie (commented):
Do we also need to update the CI YAML files to make use of this?

bstrie:
Ah, never mind, I see that it should happen automatically.

PR #2106, opened 2022-08-01
feat(shim-kvm): update aes-gcm requirement from 0.9 to 0.10.1

haraldh:
Requires
```
[target.x86_64-unknown-none]
rustflags = [
    "-C", "linker=gcc",
    "--cfg", "polyval_force_soft",
    "--cfg", "aes_force_soft",
]
```

otherwise it doesn't compile for the target and the build errors out with:

```
LLVM ERROR: Do not know how to split the result of this operator!
```

### DIFF
--- a/crates/shim-kvm/src/snp/ghcb.rs
+++ b/crates/shim-kvm/src/snp/ghcb.rs
@@ -15,9 +15,7 @@ use core::arch::asm;
 use core::mem::size_of;
 use core::ptr;
 
-use aes_gcm::AeadInPlace;
-use aes_gcm::NewAead;
-use aes_gcm::{Aes256Gcm, Key, Nonce, Tag};
+use aes_gcm::{AeadInPlace, Aes256Gcm, KeyInit, Nonce, Tag};
 use const_default::ConstDefault;
 use sallyport::libc::EINVAL;
 use spinning::Lazy;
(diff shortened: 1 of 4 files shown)


crates/shim-kvm/src/snp/ghcb.rs
```diff
@@ -668,8 +666,7 @@ impl GhcbExtHandle {
 
         let vmpck0 = SECRETS.get_vmpck0();
 
-        let key = Key::from_slice(&vmpck0);
-        let cipher = Aes256Gcm::new(key);
+        let cipher = Aes256Gcm::new_from_slice(&vmpck0).unwrap();
```
bstrie, line 669: Why an unwrap rather than handling the error?

crates/shim-kvm/src/snp/ghcb.rs
```diff
@@ -730,8 +727,7 @@ impl GhcbExtHandle {
         }
 
         let vmpck0 = SECRETS.get_vmpck0();
-        let key = Key::from_slice(&vmpck0);
-        let cipher = Aes256Gcm::new(key);
+        let cipher = Aes256Gcm::new_from_slice(&vmpck0).unwrap();
```
bstrie, line 730: Same here.

crates/shim-kvm/src/snp/ghcb.rs
haraldh: If this does not work, we are in deep trouble, because we can't talk to the firmware with the key, the firmware gave us.
Better panic and bail out. There is no nice way out.

crates/shim-kvm/src/snp/ghcb.rs
haraldh: Same here.

bstrie (approved):

PR #2113, opened 2022-08-03
feat(shim-sgx): add `deny(clippy::integer_arithmetic)`

haraldh:
This should prevent some integer_arithmetic bug exploits.


Freax13:
Just FYI there's also a profile option [`overflow-checks`](https://doc.rust-lang.org/cargo/reference/profiles.html#overflow-checks) that can be used to enable panics on overflow even in release mode. This would also affect any dependencies.

haraldh:
Superseded by https://github.com/enarx/enarx/pull/2114

PR #2116, opened 2022-08-04
wasi-crypto support

rjzak:
`wasi-crypto` support.
Needs changed Wasmtime, which then pulls in the latest `wasi-crypto` changes.
Blocked by https://github.com/bytecodealliance/wasmtime/pull/4612


bstrie:
Marking this as a draft for now so that it doesn't get merged, since we can't release with a git dependency.

dpal:
@rjzak What do we need to do here? @npmccallum Should this be in?

rjzak:
Closing since this used an older version of Wasmtime, and discussions need to be had about using Wasi-Crypto in Enarx, and how to get Steward certs into Wasi-Crypto for application use.

PR #2146, opened 2022-08-30
fix(shim-kvm): correctly set XCR0 according to cpuid

haraldh:
and patch the osxsave feature flag for cpuid.

Directly modifying the CPUID page did not have any effect (as if it was read-only).

### DIFF
--- a/crates/shim-kvm/src/snp/cpuid_page.rs
+++ b/crates/shim-kvm/src/snp/cpuid_page.rs
@@ -73,6 +73,12 @@ impl CpuidPage {
     const fn space_size() -> usize {
         4096 - size_of::<CpuidPageEntry>()
     }
+
+    /// Get all entries
+    #[inline]
+    pub fn get_functions(&self) -> &[CpuidFunctionEntry] {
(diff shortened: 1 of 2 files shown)


crates/shim-kvm/src/snp/cpuid_page.rs
```diff
@@ -146,12 +144,19 @@ pub fn cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
...
-                    Some(CpuidResult {
+                    let mut res = CpuidResult {
                         eax: e.eax,
                         ebx: e.ebx,
                         ecx: e.ecx,
                         edx: e.edx,
-                    })
+                    };
+
+                    if leaf == 1 && sub_leaf == 0 && (res.ecx & (1u32 << 26)) != 0 {
```
bstrie, line 154: Is whatever this operation is doing properly documented by the existing function docstring? If not, can we add documentation explaining what this is doing?

haraldh: added docs and reshuffled things a bit.

bstrie (approved):

Issue #2102, opened 2022-07-29
[Feature]: Extend the `sign` cli command


haraldh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Depends on https://github.com/enarx/enarx/pull/2101

### Future CLI

`-o, --out FILE` is optional and will write the output to a file, otherwise standard output is used.

#### SGX key creation

```console
$ enarx sign sgx create --out sgx.key 
```

#### SEV key creation

```console
$ enarx sign sev create --author --out sev-author.key
$ enarx sign sev create --out sev-id.key
$ enarx sign sev sign --sev-author-key sev-author.key --sev-id-key sev-id.key --out sev-id-key-signature.blob
```

`sev-id-key-signature.blob` contains the signature `id_auth.id_key_sig` and the public key of the sev-author.key `id_auth.author_key`,
so the `sev-author.key` does not have to be used in the final signing of the enarx executable. 

#### Enarx executable signing

```console
$ enarx sign --sgx-key sgx.key --sev-id-key sev-id.key --sev-id-key-signature sev-id-key-signature.blob --out sig.json 
```

### SEV key digest

```console
$ enarx sign sev digest --key sev-author.key --out sev-author-key-digest.json
$ enarx sign sev digest --key sev-id.key     --out sev-id-key-digest.json
```

haraldh:
@dxfg you might want to assign yourself

dxfg:
@haraldh I'm not sure I have permissions to assign myself to this

haraldh:
@dxfg now that you answered, I was able to do so :)

PR #2151, opened 2022-08-31
feat: extend the enarx CLI with a key command and change the sign command

haraldh:
Fixes: https://github.com/enarx/enarx/issues/2102

Extend the enarx CLI with a `key` command and change the `sign` command.

`-o, --out FILE` is optional and will write the output to a file, otherwise standard output is used.

```console
$ enarx key sgx create --out sgx.key
```

```console
$ enarx key sev create --out sev-author.key
$ enarx key sev create --out sev-id.key
$ enarx key sev sign --sev-author-key sev-author.key --sev-id-key sev-id.key --out sev-id-key-signature.blob
```

`sev-id-key-signature.blob` contains the signature `id_auth.id_key_sig` and the public key of the sev-author.key `id_auth.author_key`,
so the `sev-author.key` does not have to be used in the final signing of the enarx executable.

```console
$ enarx sign --sgx-key sgx.key --sev-id-key sev-id.key --sev-id-key-signature sev-id-key-signature.blob --out sig.json
```

```console
$ enarx key sev digest --key sev-author.key --out sev-author-key-digest.json
$ enarx key sev digest --key sev-id.key     --out sev-id-key-digest.json
```

Co-authored-by: Harald Hoyer <harald@profian.com>
Signed-off-by: Sriram Tolety <dxfg@535e9e96d61d>


src/cli/key/sev/create.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[clap(long)]
+    out: Option<Utf8PathBuf>,
+}
+
+impl Options {
+    pub fn execute(self) -> anyhow::Result<()> {
+        let rng = rand::thread_rng();
+        let signing_key = SigningKey::random(rng);
+
+        // somehow SigningKey does not implement EncodePrivateKey
```
bstrie, line 29: Is this an issue with the p384 crate? Has it been reported?

src/cli/key/sev/sign.rs
```diff
@@ -0,0 +1,131 @@
...
+use p384::ecdsa::SigningKey;
+use p384::elliptic_curve::sec1::Coordinates;
+use p384::pkcs8::DecodePrivateKey;
+use p384::EncodedPoint;
+
+/// Generate SEV Signature for provided SEV keys and write to file.
+#[derive(Args, Debug)]
+pub struct Options {
+    /// SEV P-384 private key in PEM form
+    #[clap(long, required = true)]
```
bstrie, line 23: Is `required = true` doing anything here? Isn't it implied that all the arguments here required since they're not `Option`s?

src/cli/key/sev/sign.rs
```diff
@@ -0,0 +1,131 @@
...
+    id_key: Utf8PathBuf,
+
+    /// File path to write signature
+    #[clap(long, required = true)]
+    out: Utf8PathBuf,
+}
+
+pub fn sign_id_sev_key(
```
bstrie, line 35: Is this something that should live in the sev crate?

src/cli/key/sev/sign.rs
```diff
@@ -0,0 +1,131 @@
...
+        Coordinates::Uncompressed { x, y } => (x.to_vec(), y.to_vec()),
+        _ => bail!("Invalid verifying key"),
+    };
+    // The r and s values have to be in little-endian order
+    r.reverse();
+    s.reverse();
+    // and are zero extended to the size of the components in the IdAuth struct.
+    id_auth.id_key.component.r[..r.len()].copy_from_slice(&r);
+    id_auth.id_key.component.s[..s.len()].copy_from_slice(&s);
+    // Sign the SEV signing key with the SEV author key. This will be removed in the future.
```
bstrie, line 56: Is this comment still relevant, or is it a leftover from when this function was copied over here?

src/cli/key/sev/digest.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// SEV P-384 private key in PEM form
+    #[clap(value_name = "SEV KEY")]
+    key: Utf8PathBuf,
+
+    /// File path to write digest
+    #[clap(long)]
+    out: Option<Utf8PathBuf>,
+}
+
+pub fn sev_key_digest(sev_key: &SigningKey) -> anyhow::Result<Vec<u8>> {
```
bstrie, line 32: Does this need to be pub?

tests/wasm/mod.rs
```diff
@@ -91,18 +36,23 @@ pub fn enarx_run<'a>(
...
         |cmd| {
             cmd.args(vec![
-                "sign",
-                "--sgx-key",
-                "tests/data/sgx-test.key",
-                "--sev-key",
-                "tests/data/sev-id.key",
-                "--sev-author-key",
-                "tests/data/sev-author.key",
+                OsStr::new("sign"),
```
bstrie, line 47: Why do all these need to be `OsStr`s now?

bstrie:
As far as UX goes, since we're already stuffing multiple things into the SEV key blob, is there a reason that signing SEV like this:
```
enarx sign --sev-id-key sev-id.key --sev-id-key-signature sev-id-key-signature.blob
```
couldn't become this:
```
enarx sign --sev-key sev-id-key-signature.blob
```
which is to say, would it make sense to also stuff the full SEV ID key into the blob too?

If so, then maybe we could also further simplify this:
```
enarx key sev create --out sev-author.key
enarx key sev create --out sev-id.key
enarx key sev sign --author-key sev-author.key --id-key sev-id.key --out sev-id-key-signature.blob
```
into just:
```
enarx key sev crate --out sev-id-key-signature.blob
```
However, it's also possible that I don't fully understand how all these keys are intended to be used.

src/cli/key/sev/create.rs
haraldh: filed https://github.com/RustCrypto/signatures/issues/519

src/cli/key/sev/sign.rs
haraldh: We don't use any `sev` crate... did you mean `sev` submodule?

src/cli/key/sev/sign.rs
haraldh: no, it's not.. thank you!

src/cli/key/sev/digest.rs
haraldh: err... no.. right

src/cli/key/sev/sign.rs
haraldh: leftover .. thanks!

haraldh:
@bstrie

Of course we could do this, but it would make `digest` more complicated.

Regarding "UX". This is not something a developer would have to use on a daily basis.
- The `key * create` and `key sev sign` are done only once, until the keys are breached.
- The `sign` command is only used once per release.

Also the private `key` files produced may be replaced by HW crypto, where we don't have access to the private keys, therefore we should not store private keys anywhere else.

src/cli/key/sev/sign.rs
bstrie: I thought we were using https://crates.io/crates/sev , but maybe that's for something different.

tests/wasm/mod.rs
bstrie: The current approach is fine, I was just curious.

bstrie (approved):

Issue #2153, opened 2022-08-31
[Feature]:  Release dpkg-packaged version of enarx


platten:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Extend release workflow to produce enarx `dpkg`/`deb` artifact and include it as part of released assets.

### Acceptance Criteria

- Create the RPM spec for enarx
- Extend the workflow
- build the RPM using the release workflow
- test the proper installation and functionality of enarx installed using the `dpkg` package

Issue #2152, opened 2022-08-31
[Feature]: Release rpm packaged version of enarx


platten:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Extend release workflow to produce enarx RPM artifact and include it as part of released assets.

### Acceptance Criteria

- Create the RPM spec for enarx
- Extend the workflow
- build the RPM using the release workflow
- test the proper installation and functionality of enarx installed using the rpm package

PR #2154, opened 2022-09-01
feat: add rpm and debian packages to release

puiterwijk:
Fixes: #2152
Fixes: #2153

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -190,11 +190,76 @@ jobs:
 
   # TODO: redevelop SBOM generation per issue #1954
 
-  release:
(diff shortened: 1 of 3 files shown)


bstrie (changes requested):
I see that this adds a new top-level `contrib` directory. Instead, can we use the existing top-level `release` directory?

rjzak:
Shouldn't `contrib/enarx.dpkg` be renamed to `contrib/enarx_control`, since it's a control file, and not a package itself?

puiterwijk:
Regarding both: fixed it, moved the files to release/linux/{rpm/enarx.spec, deb/control}.

bstrie (approved):

PR #2156, opened 2022-09-02
ci(release): add keep signatures to released artifacts

platten:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -124,6 +124,41 @@ jobs:
           name: enarx-x86_64-windows-msi
           path: target\wix\enarx-x86_64-windows.msi
 
+  sign-x86_64:
+    needs: [ build-x86_64 ]
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: enarx-x86_64-unknown-linux-musl
+    - name: Generate the enarx keep signature (no fork)
+      if: ${{ !github.event.pull_request.head.repo.fork }}
+      env:
+        SEV_ID_KEY: ${{ secrets.SEV_ID_KEY }}
+        SEV_ID_KEY_SIGNATURE_BLOB: ${{ secrets.SEV_ID_KEY_SIGNATURE_BLOB }}
+        SGX_KEY: ${{ secrets.SGX_KEY }}
+      run: |
+        base64 --decode <<<${SEV_ID_KEY} | gunzip > sev-id.key
+        base64 --decode <<<${SEV_ID_KEY_SIGNATURE_BLOB} | gunzip > sev-id-key-signature.blob
+        base64 --decode <<<${SGX_KEY} | gunzip > sgx.key
+        chmod +x enarx-x86_64-unknown-linux-musl
+        ./enarx-x86_64-unknown-linux-musl sign --sgx-key sgx.key --sev-id-key sev-id.key --sev-id-key-signature sev-id-key-signature.blob --out enarx-x86_64-unknown-linux-musl.sig
+        rm -f sev-id.key sev-id-key-signature.blob sgx.key
+    - name: Generate the enarx keep signature (fork)
+      if: ${{ github.event.pull_request.head.repo.fork }}
+      run: |
+        chmod +x enarx-x86_64-unknown-linux-musl
+        ./enarx-x86_64-unknown-linux-musl key sgx create --out sgx.key
+        ./enarx-x86_64-unknown-linux-musl key sev create --out sev-author.key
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -240,6 +276,15 @@ jobs:
         cp "enarx-${{ matrix.architecture.build }}-unknown-linux-musl" dpkg/usr/bin/enarx
         chmod +x dpkg/usr/bin/enarx
         sudo chown root:root dpkg/usr/bin/enarx
+    - if: ${{ matrix.architecture.build == 'x86_64' }}
+      run: |
+        mkdir -p dpkg/usr/lib/enarx
+        cp "enarx-x86_64-unknown-linux-musl.sig" dpkg/usr/lib/enarx/enarx.sig
+        chmod 444 dpkg/usr/lib/enarx/enarx.sig
+        sudo chown root:root dpkg/usr/lib/enarx/enarx.sig
+    - run: |
```
puiterwijk, line 285: I don't think you want to do `dpkg -i` here yet, neither the `enarx --version`. Those should be run in the next step.

.github/workflows/release.yml
```diff
@@ -124,6 +124,31 @@ jobs:
...
+        name: enarx-x86_64-unknown-linux-musl
+    - name: Generate the enarx keep signature
+      env:
+        SEV_ID_KEY: ${{ secrets.SEV_ID_KEY }}
+        SEV_ID_KEY_SIGNATURE_BLOB: ${{ secrets.SEV_ID_KEY_SIGNATURE_BLOB }}
+        SGX_KEY: ${{ secrets.SGX_KEY }}
+      run: |
+        echo -e "$SEV_ID_KEY" | base64 --decode | gunzip > sev-id.key
+        echo -e "$SEV_ID_KEY_SIGNATURE_BLOB" | base64 --decode | gunzip > sev-id-key-signature.blob
+        echo -e "$SGX_KEY" | base64 --decode | gunzip > sgx.key
```
haraldh, line 142: Why `echo -e`?? Did you mean `echo -n`??

haraldh (changes requested):

.github/workflows/release.yml
platten: fixed.

.github/workflows/release.yml
```diff
@@ -124,6 +124,31 @@ jobs:
...
+        name: enarx-x86_64-unknown-linux-musl
+    - name: Generate the enarx keep signature
+      env:
+        SEV_ID_KEY: ${{ secrets.SEV_ID_KEY }}
+        SEV_ID_KEY_SIGNATURE_BLOB: ${{ secrets.SEV_ID_KEY_SIGNATURE_BLOB }}
+        SGX_KEY: ${{ secrets.SGX_KEY }}
+      run: |
+        echo -n "$SEV_ID_KEY" | base64 --decode | gunzip > sev-id.key
+        echo -n "$SEV_ID_KEY_SIGNATURE_BLOB" | base64 --decode | gunzip > sev-id-key-signature.blob
+        echo -n "$SGX_KEY" | base64 --decode | gunzip > sgx.key
```
haraldh, line 142:
```suggestion
        base64 --decode <<<${SEV_ID_KEY} | gunzip > sev-id.key
        base64 --decode <<<${SEV_ID_KEY_SIGNATURE_BLOB} | gunzip > sev-id-key-signature.blob
        base64 --decode <<<${SGX_KEY} | gunzip > sgx.key
```

haraldh (changes requested):

.github/workflows/release.yml
platten: fixed.

.github/workflows/release.yml
```diff
@@ -124,6 +124,31 @@ jobs:
...
+  sign-x86_64:
+    needs: [ build-x86_64 ]
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: enarx-x86_64-unknown-linux-musl
+    - name: Generate the enarx keep signature
+      env:
+        SEV_ID_KEY: ${{ secrets.SEV_ID_KEY }}
```
puiterwijk, line 136: The problem here (and why CI is failing) is that these secrets aren't available for a PR created from another repository, which they also shouldn't be (since that would allow any contributor to get stuff signed by us).
Perhaps we should check whether these secrets are unavailable, and if they are (like they will be in e.g. PR initiated CI), generate and use a new key during this process?

platten: The secrets are not available when the workflow is executed from a fork, so I put in a workaround where new keys would be generated for that run. Situation should be addressed.

haraldh (approved):

Issue #2170, opened 2022-09-07
[Bug]: nil/kvm keeps cannot communicate to each other over TLS


rvolosatovs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Current Behaviour

Due to the certificates being self-signed in nil and kvm keeps, connection to them from another keep with always fail, due to lack of a common root of trust.

### Expected Behaviour

That should just work, e.g. for development and testing purposes.

Possible approaches:
1. Add `--insecure` flag for development builds, which would allow skipping the cert validation
2. Embed a CA cert in Enarx binary, which would be used to issue the certs for development keeps and could be exported to use with e.g. `curl` or `openssl` for testing
3. Allow Steward to issue certs for KVM and nil backends. The backend would then have to be included in the certificate and e.g. Drawbridge would need to check what backend is currently used before granting access. That means that only Steward-enabled keeps can communicate between each other.

### Environment Information

enarx 0.6.3 and latest main

rvolosatovs:
Assigning @puiterwijk @npmccallum @haraldh and myself to discuss possible solutions

rvolosatovs:
Assuming we have a possibility to run `N` workloads as a single "unit", what if `steward` was a unit-wide setting, where each workload in a unit would get a certificate issued by the selected steward.

In such scenario, when no steward is selected, Enarx could generate a CA cert and then issue workload certs using that cert.

This way workloads within a unit can communicate securely

PR #2174, opened 2022-09-07
fix: outgoing TLS connections, `localhost` handling

### LINKED ISSUE #2134: [Bug]: TLS "connect" socket I/O occasionally succeeds (and usually fails)

### Code of Conduct

### Current Behaviour

Occasional
```
Caused by:
    Resource temporarily unavailable (os error 6)
Error: failed to execute default function
```

When using

```toml
[[files]]
name = "api.github.com"
kind = "connect"
prot = "tls"
port = 443
host = "api.github.com"
```

### Expected Behaviour

Consistent success

### Environment Information

enarx 0.6.3

### Steps To Reproduce

_No response_

rvolosatovs:
Closes #2134 
Unit testing outgoing TLS requires #2170 

~Blocked by #2204~

Original implementation tested with https://github.com/enarx/codex/pull/53


npmccallum:
@rvolosatovs It seems like we need a test case for this in order to test both the blocking and non-blocking cases.

rvolosatovs:
> @rvolosatovs It seems like we need a test case for this in order to test both the blocking and non-blocking cases.

done, although that required quite a few changes in `exec-wasmtime` and implementing #2170 along the way

rvolosatovs:
I extracted a few commits from this PR into separate PRs
- #2189
- #2190 
- #2191 
- #2192 
- #2193 
- #2194 
- #2195

rvolosatovs:
now also https://github.com/enarx/enarx/pull/2196

rvolosatovs:
Interesting Windows CI failure:
```

failures:

---- wasm::listen_tcp stdout ----
waiting for workload to start...
thread '<unnamed>' panicked at 'failed to connect to TCP socket: Os { code: 10061, kind: ConnectionRefused, message: "No connection could be made because the target machine actively refused it." }', tests\wasm\mod.rs:437:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'wasm::listen_tcp' panicked at 'process timed out', tests\integration.rs:106:28

---- wasm::listen_tls stdout ----
waiting for workload to start...
thread '<unnamed>' panicked at 'failed to connect to TCP socket: Os { code: 10061, kind: ConnectionRefused, message: "No connection could be made because the target machine actively refused it." }', tests\wasm\mod.rs:519:18
thread 'wasm::listen_tls' panicked at 'process timed out', tests\integration.rs:106:28


failures:
    wasm::listen_tcp
    wasm::listen_tls
```

will do some digging

crates/exec-wasmtime/src/runtime/net/tls.rs
```diff
@@ -130,6 +131,7 @@ impl Stream {
         let mut stream = Self {
             tcp,
             tls,
+            nonblocking: false, // this is only valid under assumption that this executable has opened the socket
```
haraldh, line 134: you can query the current state of blocking with libc::fcntl(fd, libc::F_GETFL...) and check for O_NONBLOCK.

crates/exec-wasmtime/src/runtime/net/tls.rs
```diff
@@ -270,7 +278,11 @@ impl WasiFile for Listener {
             .context("could not create new TLS connection")
             .map(Connection::Server)?;
 
-        let mut stream = Stream { tcp, tls };
+        let mut stream = Stream {
+            tcp,
+            tls,
+            nonblocking: false,
```
haraldh, line 284: you can query the current state of blocking with libc::fcntl(fd, libc::F_GETFL...) and check for O_NONBLOCK.

testdata/tls/ca.conf
```diff
@@ -0,0 +1,16 @@
+[req]
```
haraldh, line 1: please use the already existing `tests/data` rather than `testdata`

tests/crates/enarx_wasm_tests/src/bin/connect.rs
```diff
@@ -0,0 +1,32 @@
...
+fn main() -> anyhow::Result<()> {
+    let fd_count: usize = env::var("FD_COUNT")
+        .context("failed to lookup `FD_COUNT`")?
+        .parse()
+        .context("failed to parse `FD_COUNT`")?;
+    ensure!(
+        fd_count == 4, // STDIN, STDOUT, STDERR and the socket connected to the endpoint
+        "unexpected amount of file descriptors received"
+    );
+    ensure!(
```
haraldh, line 26: out of curiosity, why `ensure!` and not `assert!` ??

tests/wasm/mod.rs
```diff
@@ -15,7 +15,6 @@ use process_control::Output;
 use rustls::client::{ServerCertVerified, ServerCertVerifier};
 use rustls::version::TLS13;
 use rustls::Certificate;
-use serial_test::serial;
```
haraldh, line 12: Why? Doesn't this then overload on SEV-SNP, which was the reason why we introduced this.

crates/exec-wasmtime/src/lib.rs
```diff
@@ -105,14 +105,14 @@ mod test {
...
-        let mut file = NamedTempFile::new().context("failed to create module file")?;
+        let mut file = tempfile().context("failed to create module file")?;
         file.write(wasm).context("failed to write module to file")?;
         file.rewind().context("failed to rewind file")?;
         Runtime::execute(Package::Local {
             #[cfg(unix)]
-            wasm: (&file).as_raw_fd(),
+            wasm: file.as_raw_fd(),
             #[cfg(windows)]
-            wasm: file.into_file(),
```
haraldh, line 115: well, I guess, you could have use `NamedTempFile`, if you kept the result of `file.into_file()` in scope as long as the file is needed.
```rust
{ // scope start
   let wasm_file = file.into_file();
   Runtime::execute(Package::Local /* ... */)
} // scope end, wasm_file gets deleted by `drop`
```

haraldh (changes requested):

Cargo.toml
```diff
@@ -84,6 +84,7 @@ protobuf-codegen-pure = { version = "2.27", default-features = false }
 
 [dev-dependencies]
 process_control = { version = "3.3", default-features = false }
+rustls = { version = "0.20.6", default-features = false, features = ["dangerous_configuration"] }
```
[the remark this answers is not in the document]
rvolosatovs: This is only used for testing, specifically to allow self-signed certs

crates/exec-wasmtime/src/runtime/net/tls.rs
rvolosatovs: Right, but we're opening the socket ourselves, why do a syscall?

testdata/tls/ca.conf
rvolosatovs: We already use `testdata` in https://github.com/profianinc/steward and https://github.com/profianinc/drawbridge. It would really want to have these all in one consistent location.

I'll move current files to `tests/data` and file an issue about it

tests/crates/enarx_wasm_tests/src/bin/connect.rs
rvolosatovs: Panics are much more difficult to debug, than errors

tests/wasm/mod.rs
rvolosatovs: So far it looks like it works fine.

Without this, tests consistently time out and take forever locally.
Tests are only useful if they are simple to run and run fast.

Is there a config I can use, which would only enable this on SEV?

crates/exec-wasmtime/src/lib.rs
rvolosatovs: Yes, but we don't need `NamedTempFile` and `tempfile()` is the preferred option.

> When choosing between the temporary file variants, prefer `tempfile` unless you either need to know the file’s path or to be able to persist it.

https://docs.rs/tempfile/latest/tempfile/index.html#design

crates/exec-wasmtime/src/runtime/net/tls.rs
rvolosatovs: Yes, but we don't need to do that

crates/exec-wasmtime/src/runtime/net/tls.rs
haraldh: ok

crates/exec-wasmtime/src/runtime/net/tls.rs
rvolosatovs: We're unsetting any flags below anyway (as suggested by `man 2 accept`):
> On Linux, the new socket returned by accept() does not inherit file status flags such as O_NONBLOCK and O_ASYNC from the listening socket.  This behavior differs from the canonical BSD sockets implementation.  Portable programs should not rely on inheritance or noninheritance of file status flags and always explicitly set all required flags on the socket returned from accept().

tests/wasm/mod.rs
haraldh: ok, fine for now... we can reintroduce it, if it causes unexpected failures.

testdata/tls/ca.conf
rvolosatovs: Well, since we are nesting these in different subcrates, maybe it's fine as is

haraldh (approved):

PR #2178, opened 2022-09-09
fix(shim-kvm): use SnpReportRequest struct

haraldh:
Use the SnpReportRequest struct to request attestation reports from the SEV-SNP PSP firmware.

Previously the get_attestation call failed spuriously, if there were non-zero bytes following the user data in the uninialized memory, which then causes the PSP firmware to return `report.status = 0x16`.


rvolosatovs (approved):
Not familiar with internals, but implementation LGTM

Issue #2160, opened 2022-09-05
[Feature]: Threading for SGX


haraldh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

SSIA

### Acceptance Criteria

[ ] working [`std::thread::spawn()`](https://doc.rust-lang.org/std/thread/fn.spawn.html#)
[ ] working [std](https://doc.rust-lang.org/std/index.html)::[sync](https://doc.rust-lang.org/std/sync/index.html)::[Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html#)

### Suggestions for a technical implementation

* [ ] Host: start additional threads on demand (`enarxcall::clone`)
* [ ] Keep: dequeue thread queue
* [ ] `enarxcall::sleep` and `enarxcall::wake`

Issue #2159, opened 2022-09-05
[Feature]: Threading for Enarx


haraldh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

SSIA

### Acceptance Criteria

* [ ] Threading for SGX
* [ ] Threading for KVM
* [ ] Threading for SEV-SNP

PR #2179, opened 2022-09-13
feat(sallyport): add support for threading and synchronization

haraldh:
This patch series adds sallyport support for threading and synchronization.

Enarxcalls:
- park
- unpark
- spawn

Syscalls:
- clone
- futex

Related:
- https://github.com/enarx/enarx/issues/2159
- https://github.com/enarx/enarx/issues/2160


crates/sallyport/src/guest/handler.rs
```diff
@@ -254,6 +267,54 @@ pub trait Handler {
...
+                    absolute: true,
+                });
+
+                while uaddr.load(Ordering::Relaxed) == val {
+                    let _ = self.park(timeout.as_ref());
+                }
+                Ok(0)
+            }
+            FUTEX_WAKE => {
+                // TODO: return the number of woken threads
```
rvolosatovs, line 310: Can we file an issue about it, such that we don't forget about this?

crates/sallyport/src/guest/handler.rs
```diff
@@ -254,6 +267,54 @@ pub trait Handler {
...
+    fn futex(
+        &mut self,
+        uaddr: &mut AtomicU32,
+        futex_op: c_int,
+        val: u32,
+        timespec: Option<&timespec>,
+        _uaddr2: Option<&mut AtomicU32>,
+        val3: u32,
+    ) -> Result<c_long> {
+        let futex_op = futex_op & !FUTEX_PRIVATE_FLAG;
```
rvolosatovs, line 280: Why silently drop the flag and not return `ENOTSUP` instead?
At least there should be a comment explaining why this handled this way.

crates/sallyport/src/guest/handler.rs
haraldh: added a comment

bstrie (approved):

rvolosatovs (approved):

PR #2184, opened 2022-09-15
fix(release): fix rpm artifact names

platten:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -255,12 +255,12 @@ jobs:
     # Test the just-built RPM
     - if: ${{ matrix.architecture == 'x86_64' }}
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -303,10 +303,11 @@ jobs:
       run: |
         sudo dpkg -i dpkg.deb
         /usr/bin/enarx --version
+    - run: mv dpkg.deb "enarx_${{ steps.version.outputs.version }}_${{ matrix.architecture.debarch }}.deb"
```
rvolosatovs, line 306: Why? This seems to just be extra step, which obfuscates the implementation. If you wanted to rename this, you should rename the `dpkg` directory instead containing the RPM source

rvolosatovs (changes requested):
What's the motivation for this change? Have you tested this?
From what I understood from discussion with Patrick, the `-1` part is *not* optional

npmccallum:
We don't want to remove the relver from the package name. Closing.

PR #2196, opened 2022-09-16
refactor(exec-wasmtime): minor improvements

rvolosatovs:
A few commits extracted from https://github.com/enarx/enarx/pull/2174 to simplify review

### DIFF
--- a/crates/exec-wasmtime/src/loader/compiled/mod.rs
+++ b/crates/exec-wasmtime/src/loader/compiled/mod.rs
@@ -20,15 +20,15 @@ impl Loader<Compiled> {
         let ctx = ctx.data_mut();
 
         // Set up environment variables.
-        for (k, v) in self.0.config.env.iter() {
-            ctx.push_env(k, v)?;
+        for (k, v) in self.0.config.env {
+            ctx.push_env(&k, &v)?;
         }
 
         // Set up the arguments.
         ctx.push_arg("main.wasm")
             .context("failed to push argv[0]")?;
-        for arg in self.0.config.args.iter() {
-            ctx.push_arg(arg).context("failed to push argument")?;
+        for arg in self.0.config.args {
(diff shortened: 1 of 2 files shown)


crates/exec-wasmtime/src/loader/compiled/mod.rs
```diff
@@ -20,15 +20,15 @@ impl Loader<Compiled> {
         let ctx = ctx.data_mut();
 
         // Set up environment variables.
-        for (k, v) in self.0.config.env.iter() {
```
bstrie, line 23: Is this change just intended to get rid of the explicit `.iter()`, or is the intent to change this from by-ref iteration to by-val? If it's the former, you can do `for (k, v) in &self.0.config.env {` and avoid taking references inside the loop.

crates/exec-wasmtime/src/loader/compiled/mod.rs
```diff
@@ -20,15 +20,15 @@ impl Loader<Compiled> {
...
-            ctx.push_env(k, v)?;
+        for (k, v) in self.0.config.env {
+            ctx.push_env(&k, &v)?;
         }
 
         // Set up the arguments.
         ctx.push_arg("main.wasm")
             .context("failed to push argv[0]")?;
-        for arg in self.0.config.args.iter() {
```
bstrie, line 30: Same.

crates/exec-wasmtime/src/loader/compiled/mod.rs
rvolosatovs: The intention was to move it.

crates/exec-wasmtime/src/loader/compiled/mod.rs
rvolosatovs: Same intention, `self.0.config.args` is already processed, repeated usage of it should result in compile-time error.

haraldh (approved):

PR #2197, opened 2022-09-17
feat(config): introduce file types

rvolosatovs:
This lets us provide more meaningful file variant defaults and account for e.g. TCP and TLS socket differences

Extracted from #2174 

### DIFF
--- a/crates/enarx-config/src/lib.rs
+++ b/crates/enarx-config/src/lib.rs
@@ -56,7 +56,11 @@ kind = "stderr"
 # port = 23456
 "#;
 
-const fn default_port() -> u16 {
+const fn default_tcp_port() -> u16 {
+    80
+}
+
+const fn default_tls_port() -> u16 {
     443
 }
 
@@ -149,9 +153,9 @@ pub struct Config {
 impl Default for Config {
     fn default() -> Self {
         let files = vec![
-            File::Stdin { name: None },
-            File::Stdout { name: None },
-            File::Stderr { name: None },
+            File::Stdin(Default::default()),
+            File::Stdout(Default::default()),
+            File::Stderr(Default::default()),
         ];
 
         Self {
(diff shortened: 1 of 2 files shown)


crates/enarx-config/src/lib.rs
```diff
@@ -206,65 +213,89 @@ pub enum File {
...
-            Self::Listen { name, .. } => name,
-            Self::Connect { name, host, .. } => name.as_deref().unwrap_or(host),
+            Self::Null(NullFile { name }) => name.as_deref().unwrap_or("null"),
+            Self::Stdin(StdioFile { name }) => name.as_deref().unwrap_or("stdin"),
+            Self::Stdout(StdioFile { name }) => name.as_deref().unwrap_or("stdout"),
+            Self::Stderr(StdioFile { name }) => name.as_deref().unwrap_or("stderr"),
+            Self::Listen(ListenFile::Tls { name, .. }) => name,
+            Self::Listen(ListenFile::Tcp { name, .. }) => name,
+            Self::Connect(ConnectFile::Tls { name, host, .. }) => name.as_deref().unwrap_or(host),
+            Self::Connect(ConnectFile::Tcp { name, host, .. }) => name.as_deref().unwrap_or(host),
```
haraldh, line 294: Shouldn't the name also contain the port? Otherwise we might end up with multiple duplicate names.

haraldh (changes requested):

crates/enarx-config/src/lib.rs
rvolosatovs: This method will be removed in next iteration and that's exactly how it worked before, there are no changes in behavior here.

rvolosatovs: Files will be a hashmap, where name will be required and duplicates won't be allowed

haraldh (approved):

PR #2205, opened 2022-09-20
chore: add test crate lock files

rvolosatovs:


bstrie:
The reason these aren't in the workspace is so that `cargo test --workspace` succeeds.

bstrie (approved):

PR #2206, opened 2022-09-20
fix(cli): handle cannot-be-a-base URLs in `enarx deploy`

bstrie:
`enarx deploy` allows the user to pass either a slug or a URL, and differentiates between the two based on whether the input successfully parses as a URL. However, slugs allow custom hosts to be optionally specified, along with specific ports, in which the slug looks like `host:1234/user/repo:tag`. The problem is that this successfully parses as a URL, specifically a special kind of URL known as a "cannot be a base" URL. Currently, if a slug like this is used, then `enarx deploy` will later panic because what was intended to be the host will instead be interpreted as a URL scheme. Instead, this patch causes the input to be interpreted as a slug if it is parsed as a "cannot be a base" URL.


rvolosatovs (changes requested):
I think that the closure just obfuscates the behavior.

How about we just do something like this instead?

```diff
diff --git a/src/cli/deploy.rs b/src/cli/deploy.rs
index c310ab0e..68be8df4 100644
--- a/src/cli/deploy.rs
+++ b/src/cli/deploy.rs
@@ -72,18 +72,25 @@ impl Options {
             Signatures::load(signatures)?
         };
 
-        let package: Url = package.parse().or_else(|_| {
-            use drawbridge_client::API_VERSION;
-
+        let package = match package
+            .parse()
             // TODO: Check the error
-            let (host, user, repo, tag) = parse_tag(&package)
-                .with_context(|| format!("failed to parse `{package}` as a Drawbridge slug"))?;
-            format!("https://{host}/api/v{API_VERSION}/{user}/{repo}/_tag/{tag}")
-                .parse()
-                .with_context(|| {
-                    format!("failed to construct a URL from Drawbridge slug `{package}`")
-                })
-        })?;
+            .ok()
+            .filter(|url: &Url| !url.cannot_be_a_base())
+        {
+            None => {
+                use drawbridge_client::API_VERSION;
+
+                let (host, user, repo, tag) = parse_tag(&package)
+                    .with_context(|| format!("failed to parse `{package}` as a Drawbridge slug"))?;
+                format!("https://{host}/api/v{API_VERSION}/{user}/{repo}/_tag/{tag}")
+                    .parse()
+                    .with_context(|| {
+                        format!("failed to construct a URL from Drawbridge slug `{package}`")
+                    })?
+            }
+            Some(url) => url,
+        };
 
         let code = match package.scheme() {
             "file" => {
```

Also, this lacks a test case

src/cli/deploy.rs
```diff
@@ -13,6 +13,7 @@ use crate::backend::Signatures;
 use anyhow::{anyhow, bail, Context};
 use camino::Utf8PathBuf;
 use clap::Args;
+use drawbridge_client::API_VERSION;
```
rvolosatovs, line 16: I don't think this belongs here, it's not clear what "API_VERSION" is just by itself.

rvolosatovs (approved):

bstrie:
Updated. A test case will be added in the future as part of integration testing the `enarx deploy` CLI.

rvolosatovs:
Would still be nice to see a test case covering this case

PR #2207, opened 2022-09-20
test: replace echo server with hello world in integration test

bstrie:
Eventually we want to test `enarx deploy` as part of the integration test, and in that context the interactivity of the echo server will be a pain to deal with. Since all that we care about for the client test is that a package can be uploaded and deployed, regardless of what the program actually does, replace it with a hello world example.


rvolosatovs:
> Eventually we want to test `enarx deploy` as part of the integration test,

We already do https://github.com/enarx/enarx/commit/7642f5ce0fb1aafb5ab86b11e8b4747e38cb971a

rvolosatovs (approved):

bstrie:
> Eventually we want to test enarx deploy as part of the integration test,

Specifically I'm referring to the CLI integration test.

PR #2220, opened 2022-09-22
feat(shim-sgx): add an extra stack for CSSA > 0

haraldh:
rather than reusing the stack of CSSA == 0, get a fresh one, because that rsp is not under the control of the shim.


crates/shim-sgx/src/main.rs
```diff
@@ -206,24 +206,32 @@ pub unsafe extern "sysv64" fn _start() -> ! {
         "cmp    r11,    0                   ",  // If exceptions aren't enabled yet...
         "je     2b                          ",  // ... loop forever.
 
-        // Find stack pointer for CSSA > 0
+        // Find stack pointer for CSSA == 1
+        "cmp    rax,    1                   ",  // If CSSA > 0
+        "jne    3f                          ",  // ... jump to the next section
+        "mov    r10,    rcx                 ",  // r10 = stack pointer CSSA == 0
+        "sub    r10,    0x200000            ",  // r10 = stack pointer - 2MB
```
rvolosatovs, line 213: Out of curiosity, why is this 2MB? Is that a convention? Is that an arbitrary value we chose (if so, shouldn't that be a constant defined somewhere?)

haraldh: It's currently defined in the linker layout file. Let me turn it into a constant.

rvolosatovs (approved):

PR #2222, opened 2022-09-22
feat(sgx): use CSSA level 0 EEXIT to return from `exit()` syscall

haraldh:
This lets us reuse the SGX TCS after it's returned to the pool.
It also gets rid of global arrays for thread bookkeeping.
    
sallyport:
- don't be `attacked()` by return of `exit()`
  it is correctly handled by the shim
    
shim-sgx:
- add Task Control Block (Tcb)
- load_registers() with exit_status
- return with exit_status in r8, especially CSSA=0
- remove MAX_THREADS restriction from `clone()`
- use the state in Tcb to return from `exit()` syscall
  This will EEXIT CSSA=0 to the backend ready for re-use.
    
backend/sgx:
- requeue thread on CSSA=0 EEXIT
- panic on `exit()` syscall
- use exit status in r8 of CSSA=0 EEXIT

Related: https://github.com/enarx/enarx/issues/2160

### DIFF
--- a/crates/sallyport/src/guest/handler.rs
+++ b/crates/sallyport/src/guest/handler.rs
@@ -810,7 +810,7 @@ pub trait Handler {
             (SYS_eventfd2, [initval, flags, ..]) => self
                 .eventfd2(initval as _, flags as _)
                 .map(|ret| [ret as _, 0]),
-            (SYS_exit, [status, ..]) => self.exit(status as _).map(|_| self.attacked()),
+            (SYS_exit, [status, ..]) => self.exit(status as _).map(|_| [0, 0]),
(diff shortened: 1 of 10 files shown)


npmccallum (commented):
@haraldh Looks okay to me. However, I don't think we need to allocate a full page for the thread control block. There should be ample free space in the unused section of the SSA. This would save one page per thread.

rvolosatovs (approved):
Looks good, but I can't comment on the low-level assembly bits

haraldh:
> @haraldh Looks okay to me. However, I don't think we need to allocate a full page for the thread control block. There should be ample free space in the unused section of the SSA. This would save one page per thread.

@npmccallum  there are only 824 bytes, and the Tcb needs 2624

PR #2228, opened 2022-09-22
feat(exec): encode workload ID as subject alt name in CSR

bstrie:


bstrie:
Converting to draft while we figure out exactly what information to transmit via the SANs.

npmccallum:
@bstrie This needs to be updated as we discussed today. There are three SANs:

1. the fully-qualified slug
2. the workload hash
3. the signature on the workload

We should definitely push the first two over the line ASAP.

PR #2236, opened 2022-09-23
feat: send CSR array to Steward

rjzak:
Sends the CSR as an array to Steward to be compatible with https://github.com/profianinc/steward/pull/55.

It seems the `exec-wasmtime` already takes the response into `Vec<Certificate>`.

Signed-off-by: Richard Zak <richard@profian.com>

### DIFF
--- a/crates/exec-wasmtime/src/runtime/identity/mod.rs
+++ b/crates/exec-wasmtime/src/runtime/identity/mod.rs
@@ -59,7 +59,7 @@ fn csr(pki: &PrivateKeyInfo<'_>, exts: Vec<Extension<'_>>) -> anyhow::Result<Vec
         signature: BitStringRef::from_bytes(sig.as_ref())?,
     };
 
-    Ok(req.to_vec()?)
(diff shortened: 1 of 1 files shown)


crates/exec-wasmtime/src/runtime/identity/mod.rs
```diff
@@ -59,7 +59,7 @@ fn csr(pki: &PrivateKeyInfo<'_>, exts: Vec<Extension<'_>>) -> anyhow::Result<Vec
         signature: BitStringRef::from_bytes(sig.as_ref())?,
     };
 
-    Ok(req.to_vec()?)
```
bstrie, line 62: Isn't this already a vec from the to_vec call?

rvolosatovs (changes requested):
What is this trying to achieve? I still only see one `CertReq` being sent
Compatibility issues should be covered by https://github.com/profianinc/steward/pull/58

rjzak:
I've put this on hold in favour of that PR for Steward, so I'll close this one.

Issue #1727, opened 2022-01-05
`SYS_clock_gettime` smuggling


rvolosatovs:

haraldh:
Remember https://github.com/enarx/enarx/pull/2239

rvolosatovs:
> Remember #2239

@haraldh see https://github.com/enarx/enarx/pull/2239#issuecomment-1256348974 :)

PR #2239, opened 2022-09-23
fix(shim-sgx): don't output dbg for SYS_clock_gettime

haraldh:
it's just too much


rvolosatovs (approved):
This should be reverted as part of #1727

rvolosatovs:
> This should be reverted as part of #1727

haraldh:
> > This should be reverted as part of #1727

This is why I made https://github.com/enarx/enarx/issues/1727#issuecomment-1256366094

PR #2249, opened 2022-09-29
test: check output during CLI integration tests

bstrie:
This also:

* makes the OIDC mock server return 401 on unauthorized, rather than panicking, so we can test the authorization failure mode
* makes the mock OIDC server use our hardcoded OIDC client key, rather than a dummy token
* allows setting an option to `enarx user register` via an env var
* improves the integration test helper, by making it easier to use and read, and making the error output more useful and readable

### DIFF
--- a/tests/client/util.rs
+++ b/tests/client/util.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::process::Command;
-use std::str::from_utf8;
 
 use async_std::net::{Ipv4Addr, TcpListener};
 use async_std::task::{spawn, JoinHandle};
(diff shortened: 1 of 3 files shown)


tests/client/util.rs
```diff
@@ -165,7 +180,7 @@ async fn init_drawbridge(oidc_addr: String) -> (u16, Sender<()>, JoinHandle<()>)
             OidcConfig {
                 label: "test-label".into(),
                 issuer: oidc_addr.parse().unwrap(),
-                client_id: "test-client-id".into(),
+                client_id: "4NuaJxkQv8EZBeJKE56R57gKJbxrTLG2".into(),
```
haraldh, line 183: this chunk does not match the PR or patch description

tests/client/util.rs
```diff
@@ -116,16 +131,16 @@ async fn init_oidc() -> (String, Sender<()>, JoinHandle<()>) {
...
+                                if auth.as_str().split_once(' ') != Some(("Bearer", "test-token")) {
+                                    Ok(Response::new(StatusCode::Unauthorized))
+                                } else {
+                                    json_response(&CoreUserInfoClaims::new(
+                                        StandardClaims::new(SubjectIdentifier::new(
+                                            "test|subject".into(),
+                                        )),
+                                        EmptyAdditionalClaims {},
+                                    ))
+                                }
```
haraldh, line 143: this chunk does not match the PR or patch description

haraldh (changes requested):

tests/client/util.rs
bstrie: Updated the description. This makes the mock respect our hardcoded OIDC client key; see src/cli/user/register.rs .

tests/client/util.rs
bstrie: I mentioned it in passing in the description, I've updated it to be more thorough.

bstrie:
Pushed a minor fix to address a test failure: on some machines the directory traversal would happen differently, causing the expected output of some should-fail tests to slightly differ from the received output based on which files exist/were visited first, so I've added a new directory under tests/client/data rather than hackily using the workspace itself.

haraldh:
> One of the tests is testing JSON output, which was causing test failures to due the nondeterministic order of array elements. I've pushed a fix for that, moving that test to earlier in the script to a point where there's only one element in the returned array, making the output deterministic.

Seems like we could use a JSON normalizer for comparison.

bstrie:
> Seems like we could use a JSON normalizer for comparison.

We may need that in the future, as we begin more aggressively testing CLI output. But for now it should be fine.

haraldh (approved):

PR #2254, opened 2022-09-30
test: parse JSON in CLI tests and refactor

bstrie:
By comparing equality via JSON rather than via string representation, we guard against spurious failures that might result due to property order being unspecified in JSON objects.

This also includes a few other refactorings and cleanups to the CLI test runner.


rjzak (approved):
Looks good to me

haraldh (approved):

Issue #2167, opened 2022-09-06
Deployment signing


bstrie:

bstrie:
The first step here is to support signing packages when they're published to drawbridge. To make it convenient we intend to automatically generate a keypair if one doesn't already exist and store it locally in the store that we already use for the oauth credentials.

PR #2260, opened 2022-10-03
feat: `enarx user key` subcommand for generating signing keys

bstrie:
Part of #2167 .


rvolosatovs (changes requested):
Doesn't look like this PR builds for all architectures

bstrie:
Converting to draft while I hunt down a musl bug.

PR #2268, opened 2022-10-07
feat: remove naked functions

haraldh:
and replace them with `global_asm!`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1286,6 +1286,7 @@ dependencies = [
  "lset",
  "nbytes",
  "noted",
+ "paste",
  "primordial",
  "rcrt1",
  "sallyport",
--- a/crates/shim-kvm/Cargo.toml
+++ b/crates/shim-kvm/Cargo.toml
@@ -28,6 +28,7 @@ sallyport = { version = "0.6.4", path = "../sallyport", default-features = false
 spinning = { version = "0.1", default-features = false }
 x86_64 = { version = "0.14.9", features = ["instructions", "inline_asm"], default-features = false }
 xsave = { version = "2.0.2", default-features = false }
+paste = { version = "1.0", default-features = false }
 
 # optional dependencies
 gdbstub = { version = "0.5.0", optional = true, default-features = false }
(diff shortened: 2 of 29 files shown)


bstrie:
For the benefit of future readers, this PR is an experiment to see whether or not it is onerous to use the (stable) `global_asm!` macro as a replacement for the (unstable) `#[naked]` function attribute. Since we are involved in the design, implementation, and potential stabilization of Rust's naked function support, we're gathering information to see whether or not it's worth continuing to pursue stablization for naked functions, and what the practical downsides would be of using `global_asm!` instead.

The *theoretical* downsides of using `global_asm!` rather than naked functions [as acknowledged by Amanieu](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/.E2.9C.94.20When.20can.20inline-asm.20return.20from.20inside.20the.20assembly.3F/near/290564028) are as follows:

> * you don't have to deal with name mangling.
> * you don't need to insert the target-specific function prefix/suffix directives.
> * it works with generic functions (but I haven't heard of anyone actually using that).

In the process of converting our naked functions to `global_asm!`, Harald had the following feedback:

> So, one nice argument for naked_functions is, if you have prolog/epilog assembler parts, especially with a macro, where you can easily make a wrapper without constructing new idents.
> See https://github.com/enarx/enarx/pull/2268/files#diff-3a740b7f0e9fca578d209e016036a034ed177269051ac1b7ea3207121f736ba0
> With global_asm! I have to use the paste! crate to create a new identifier, which should be private, but now can clash, because of no namespacing.
> This: https://github.com/enarx/enarx/pull/2268/files#diff-3a740b7f0e9fca578d209e016036a034ed177269051ac1b7ea3207121f736ba0R131
> `[<__ $name _inner>]`
> doable, but ugly
> same here:
https://github.com/enarx/enarx/pull/2268/files#diff-d4049c0229f5bac756f2e12260fb347194ec8e83fe42515d5656b22d9a889576

We may consider accepting this PR even if only to avoid having one more unstable feature in the codebase, although even if we do accept this PR we may still decide to continue pursuing the stabilization of naked functions depending on the result of our discussion.

bstrie:
Marking as a draft to make sure it doesn't get merged before Nathaniel has a chance to look at it.

bstrie:
Closing in favor of the rebased https://github.com/enarx/enarx/pull/2350.

PR #2269, opened 2022-10-07
build: use crane

rvolosatovs:
- Git dependencies are now supported out-of-the-box
- The build is greatly simplified
- Caching performance should be improved
- `cargo clippy`, `cargo fmt` and `cargo nextest` are now run in release mode on `nix flake check`
- All binaries except for Windows are now cross-compiled via nix on Github default runners

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,24 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use once_cell::sync::Lazy;
+use std::env::var;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+/// Whether Enarx compilation target is x86_64 Linux or not.
+// NOTE: This may or may not correspond to `target_os` and `target_arch`, since `build.rs` is
+// compiled for the host triple and not the cargo compilation target
+static IS_X86_64_LINUX: Lazy<bool> = Lazy::new(|| {
+    var("CARGO_CFG_TARGET_OS").expect("missing CARGO_CFG_TARGET_OS") == "linux"
+        && var("CARGO_CFG_TARGET_ARCH").expect("missing CARGO_CFG_TARGET_ARCH") == "x86_64"
+});
+
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 use std::os::unix::fs::FileTypeExt;
 
-#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 fn generate_protos() {
-    use std::path::{Path, PathBuf};
-
     fn create(path: &Path) {
-        match std::fs::create_dir(path) {
+        match fs::create_dir(path) {
             Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {}
             Err(e) => {
                 eprintln!("Can't create {:#?} : {:#?}", path, e);
@@ -18,7 +28,7 @@ fn generate_protos() {
         }
     }
 
-    let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
+    let out_dir = PathBuf::from(var("OUT_DIR").unwrap());
     let out_dir_proto = out_dir.join("protos");
     create(&out_dir_proto);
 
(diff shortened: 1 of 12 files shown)


rust-toolchain.toml
```diff
@@ -1,5 +1,10 @@
 [toolchain]
 channel = "nightly-2022-09-15"
-targets = ["x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-unknown-none", "wasm32-wasi"]
+targets = [
+    "aarch64-unknown-linux-musl",
```
bstrie, line 4: Why wasn't this necessary before this PR?

crates/exec-wasmtime/src/bin/enarx-exec-wasmtime.rs
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 #![cfg(all(target_arch = "x86_64", target_os = "linux"))]
-#![doc = include_str!("../README.md")]
+#![doc = include_str!("../../README.md")]
```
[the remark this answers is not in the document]
rvolosatovs: It's about the `bin` directory, not about the `main`. All other crates contain `bin` directory

rust-toolchain.toml
rvolosatovs: Because it was specified within `flake.nix`, this file was not constructing a build toolchain, only the development one

crates/exec-wasmtime/src/bin/enarx-exec-wasmtime.rs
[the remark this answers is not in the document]
bstrie: Which crates contain a bin directory? None of the crates in the workspace do.

rvolosatovs:
I think I know how to improve dependency caching, small change incoming, want to test it first

build.rs
```diff
@@ -39,15 +49,17 @@ fn main() {
...
+        println!("cargo:rustc-cfg=enarx_with_shim");
+    }
 
-    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
-    generate_protos();
+    if *IS_X86_64_LINUX {
+        generate_protos();
+    }
 
     #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```
bstrie, line 60: You don't want to change this cfg as well?

build.rs
```diff
@@ -57,20 +69,17 @@ fn main() {
...
-                .file_type()
-                .is_socket()
+            && Path::new(AESM_SOCKET).exists()
+            && fs::metadata(AESM_SOCKET).unwrap().file_type().is_socket()
         {
             println!("cargo:rustc-cfg=host_can_test_attestation");
         }
     }
 
     #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```
bstrie, line 79: Or this one?

build.rs
rvolosatovs: No, because this is only valid when x86_64 linux is the host

build.rs
rvolosatovs: No, because this is only valid when x86_64 linux is the host

build.rs
```diff
@@ -39,15 +49,17 @@ fn main() {
     // FIXME: this exists to work around https://github.com/rust-lang/cargo/issues/10527
     println!("cargo:rerun-if-changed=crates/");
 
-    #[cfg(all(target_os = "linux", target_arch = "x86_64",))]
-    println!("cargo:rustc-cfg=enarx_with_shim");
+    if *IS_X86_64_LINUX {
```
bstrie, line 52: Is a `Lazy` necessary here? Furthermore, are two separate `if`s necessary here? How about a single `if` with:
```rust
if "linux" == var("CARGO_CFG_TARGET_OS").expect("missing CARGO_CFG_TARGET_OS")
    && "x86_64" == var("CARGO_CFG_TARGET_ARCH").expect("missing CARGO_CFG_TARGET_ARCH") {
```

flake.lock
```diff
@@ -1,23 +1,29 @@
...
       },
       "locked": {
-        "lastModified": 1663570974,
-        "narHash": "sha256-ncUdRdY70VdJIX6Mi+820xeD7FutADd3NbQR0BKkFYA=",
-        "owner": "nix-community",
-        "repo": "fenix",
-        "rev": "02093d3aca186135da78b76ac28ec58031391076",
+        "lastModified": 1665490127,
+        "narHash": "sha256-ZQdaPEgnR8AZGSjQ5dUYz8+D+EsKG4GymmDenBBwT30=",
+        "owner": "rvolosatovs",
```
bstrie, line 18: Why use your fork rather than upstream? Can we leave a comment with a link to an issue regarding using upstream instead? And if we must use a fork in the long-term, should we fork it into the Enarx organization?

build.rs
rvolosatovs: That's how existing code was organized and I did not change that. This organization also makes sense, because these are two distinctively different features we're enabling the way I see it.

IMO the "single `if`" approach is terrible for readability

flake.lock
rvolosatovs: It will go away in a few days, I just need to get back to the PR referenced in the comment in the flake

I can create an issue and reference in a follow-up PR

rvolosatovs: https://github.com/enarx/enarx/pull/2269/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R7

bstrie (approved):

PR #2275, opened 2022-10-11
refactor: remove `IS_X86_64_LINUX` variable

rvolosatovs:
Define much more useful `target_os` and `target_arch` variables at runtime which are also consistent with Rust `cfg` attributes


rvolosatovs:
The CI of course, as always, fails with

> Run sudo apt -o Acquire::Retries=3 install -y musl-tools curl
>
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
>
> E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem. 

Why don't we use nix for this again?

bstrie:
LGTM, let me know when CI is working again so I can approve it.

bstrie (approved):

PR #2276, opened 2022-10-11
fix: support all platforms in tests

### LINKED ISSUE #2274: [Bug]: `cargo test` does not work out-of-the-box on all platforms

### Code of Conduct

### Current Behaviour

```
 enarx-nextest>         FAIL [   0.006s] enarx::integration exec::cpuid
enarx-nextest> --- STDOUT:              enarx::integration exec::cpuid ---
enarx-nextest> running 1 test
enarx-nextest> test exec::cpuid ... FAILED
enarx-nextest> failures:
enarx-nextest> failures:
enarx-nextest>     exec::cpuid
enarx-nextest> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 44 filtered out; finished in 0.00s
enarx-nextest> --- STDERR:              enarx::integration exec::cpuid ---
enarx-nextest> Error: The nil backend cannot be called with an executable!
enarx-nextest> thread 'exec::cpuid' panicked at 'assertion failed: `(left == right)`
enarx-nextest>   left: `1`,
enarx-nextest>  right: `0`: Expected exit status differs.', tests/integration.rs:225:5
enarx-nextest> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
enarx-nextest>    Canceling due to test failure: 1 tests still running
enarx-nextest>         FAIL [   0.014s] enarx::integration exec::echo
enarx-nextest> --- STDOUT:              enarx::integration exec::echo ---
enarx-nextest> running 1 test
enarx-nextest> test exec::echo ... FAILED
enarx-nextest> failures:
enarx-nextest> failures:
enarx-nextest>     exec::echo
enarx-nextest> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 44 filtered out; finished in 0.01s
enarx-nextest> --- STDERR:              enarx::integration exec::echo ---
enarx-nextest> Error: The nil backend cannot be called with an executable!
enarx-nextest> thread '<unnamed>' panicked at 'failed to write stdin to child: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }', tests/integration.rs:92:18
enarx-nextest> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
enarx-nextest> thread 'exec::echo' panicked at 'failed to provide input for process: Any { .. }', tests/integration.rs:109:22
enarx-nextest>         PASS [   0.304s]   enarx::bin/enarx backend::sev::snp::vcek::tests::test_write
enarx-nextest>      Summary [   0.331s] 29/61 tests run: 27 passed, 2 failed, 1 skipped
enarx-nextest> error: test run failed
```

### Expected Behaviour

`cargo test` working always on all targets

### Environment Information

This is `nil` backend build for x86_64 linux within Nix sandbox

### Steps To Reproduce

_No response_

rvolosatovs:
Closes #2274 

When building for x86_64 Linux, shims should be built in the binary, but
that does not mean that the host can (or should) actually use them.

Introduce integration test helpers, which check `ENARX_BACKEND` within
the environment among other things at runtime

This means that `ENARX_BACKEND=nil cargo test` now succeeds even on systems
with e.g. SGX available

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -7,14 +7,15 @@ extern crate core;
 #[cfg(not(windows))]
 mod client;
 
-#[cfg(enarx_with_shim)]
+#[cfg(any(host_can_test_kvm, host_can_test_sev, host_can_test_sgx))]
 mod exec;
 
-#[cfg(enarx_with_shim)]
+#[cfg(any(host_can_test_kvm, host_can_test_sev, host_can_test_sgx))]
 mod syscall;
 
 mod wasm;
 
+use std::env::{var, VarError};
 use std::ffi::{OsStr, OsString};
 use std::io;
 use std::io::{BufReader, Read, Write};
@@ -22,6 +23,7 @@ use std::path::PathBuf;
 use std::process::{Command, Stdio};
 use std::time;
 
+use once_cell::sync::Lazy;
 use process_control::{ChildExt, Control, Output};
 use tempfile::tempdir;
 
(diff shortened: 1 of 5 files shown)


tests/integration.rs
```diff
@@ -32,6 +33,54 @@ pub const TEST_BINS_OUT: &str = "bin";
...
+                .is_some()
+        })
+        .unwrap_or_default()
+}
+
+/// Returns `true` if SGX backend should be used by Enarx given the environment
+pub fn is_sgx() -> bool {
+    selected_backend(["sgx"])
+        || cfg!(host_can_test_sgx)
+            && !selected_backend(BACKENDS.iter().filter(|&&backend| backend == "sgx"))
```
bstrie, line 55: I don't quite understand the goal with this. `BACKENDS` is const, so we know that this filter always produces an iterator equivalent to `["sgx"]`. So the condition inside the function is saying `selected_backend(["sgx"]) || (foo && !selected_backend(["sgx"]))`. When `selected_backend(["sgx"])` is true, the RHS of the `||` is irrelevant. And when it's false, the `!selected_backend(["sgx"])` is guaranteed to be true, so it has no point. But why would we want to `&&` the `cfg` with the negation of this function in the first place? What is this expressing?

rvolosatovs: Sorry, it's late - that should have been negated.

This is selecting the backend that should the user would expect to be picked by Enarx.
- If `ENARX_BACKEND = "sgx"`, then enarx with use SGX backend or fail trying
- If `/dev/sgx` device is present and no `ENARX_BACKEND` other than `sgx` is selected, then Enarx should pick SGX

rvolosatovs: Although writing this down and looking at this, I think this can be simplified by simply checking if `ENARX_BACKEND` is set at all in the last step, let me do that

rvolosatovs: Original itention was to use this as `selected_backend(["sgx", "sev"])` to assert TEE, for example, but I learned that it would not be as simple and went with functions instead. The previous shape of this helper does not make much sense anymore

rvolosatovs: The new version should make much more sense

rvolosatovs: In fact, I've found an even cleaner solution, see updated version

tests/integration.rs
```diff
@@ -32,6 +34,50 @@ pub const TEST_BINS_OUT: &str = "bin";
...
+        None => cfg!(host_can_test_sev) && !is_sgx(),
+    }
+}
+
+/// Returns `true` if KVM backend should be used by Enarx given the environment
+pub fn is_kvm() -> bool {
+    match ENARX_BACKEND.as_deref() {
+        Some("kvm") => true,
+        Some(..) => false,
+        None => cfg!(host_can_test_kvm) && !is_sgx() && !is_sev(),
```
bstrie, line 68: Since this match arm means that every other `is_foo` function will take the `None` branch, and since `is_sev`'s `None` branch calls `!is_sgx`, I believe the call to `!is_sgx()` is redundant work on this line.

tests/integration.rs
```diff
@@ -32,6 +34,50 @@ pub const TEST_BINS_OUT: &str = "bin";
...
+    }
+}
+
+/// Returns `true` if nil backend should be used by Enarx given the environment
+pub fn is_nil() -> bool {
+    match ENARX_BACKEND.as_deref() {
+        Some("nil") => true,
+        Some(..) => false,
+        None => !is_sgx() && !is_sev() && !is_kvm(),
```
bstrie, line 77: Likewise I believe that this condition can be simplified to just `!is_kvm()`.

rvolosatovs: It can, but I don't think it should.

bstrie (approved):

PR #2279, opened 2022-10-12
build: switch to `nixify`

rvolosatovs:
- Rely on https://github.com/rvolosatovs/nixify for builds
- Update workflows
- Remove redundant steps
- Test OCI


flake.lock
```diff
@@ -58,13 +58,54 @@
...
+        ],
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-overlay": "rust-overlay"
+      },
+      "locked": {
+        "lastModified": 1665662838,
+        "narHash": "sha256-v/O5BaaHc1X0DgxqN9MmQ/96RLL2G0/Gr7Teiy6wWto=",
+        "owner": "rvolosatovs",
```
bstrie, line 76: How hard will it be/how long will it take to get off of your custom fork and begin using upstream?

rvolosatovs: This is done now, btw, just need to update the dependency.
I'd rather do that in a follow-up

haraldh (approved):

PR #2282, opened 2022-10-18
ci: automate flake update

rvolosatovs:
Run `nix flake update` once and automate it to run every night an 00.00

The PR will be filed similar to https://github.com/rvolosatovs/nixify/pull/22

If it's not merged until the action runs again, it should update again and update the PR. (https://github.com/peter-evans/create-pull-request/issues/1265#issuecomment-1264784900)

This is just the first step, most notably these PRs do not trigger `push` and `pull_request` workflows until manually closed and reopened. There are better workarounds https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

I think the right thing to do is to use @enarxbot for this, but I lack access to set it up
Maybe @puiterwijk can assist here with some solution, which would let us open PR from within this repo (not a fork) to benefit from caching, while also have a way to automatically merge once CI runs and passes. See above for a GitHub App option.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,27 @@
+name: Update
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  nix-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: nixbuild/nix-quick-install-action@v18
+        with:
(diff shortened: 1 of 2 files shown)


bstrie:
So we're going to have a bot submitting a PR for this every single day? Is that sort of frequency necessary? Why not, say, once a week, or once a month, or once per release?

bstrie (approved):

PR #2245, opened 2022-09-27
feat!: rework file management via VFS

rvolosatovs:
Refs #2150 https://github.com/enarx/enarx/issues/2141
Closes #2149 

This is still a work-in-progress draft, but it does work, so exposing it for others to be able to contribute or build on top of this (specifically for devfs, memfs, and RNG access. I will get back to this on Monday.

You will find a minimal shell added in this PR. Note, that `ls` on absolute paths does not work yet and you cannot `cd ..`. You can `cd` to an absolute path, though.

For example, with `Enarx.toml`:
```toml
[stdin]
kind = "host"

[stdout]
kind = "host"

[stderr]
kind = "host"

[listen.8080]
prot = "tls"

[connect."google.com"]
prot = "tls"
```

```
$ cargo run -- run --wasmcfgfile Enarx.toml tests/crates/enarx_wasm_tests/target/wasm32-wasi/debug/shell.wasm                 

/ $ ls
con lis
/ $ ls lis
8080
/ $ ls con
google.com
/ $ cd con
/con $ cd /
/ $ ls
con lis
/ $ cd lis 
/lis $ ls
8080
/lis $ cd /con
/con $ ls
google.com
/con $ 
```

Unfortunately, I was not able to make `cat` and `echo` work on streams yet, but that's in-progress.

This should be enough to start with implementations of other FS'es

PR #2308, opened 2022-11-09
feat(cli): support passing `enarx run` configuration via CLI

### LINKED ISSUE #2281: Specify Enarx.toml via CLI
The goal is to be able to specify and/or override the values in Enarx.toml from within `enarx run`. UI to be determined. And if we have all the code necessary to manipulate Enarx.toml from the CLI, we may as well also expose it from `enarx config` as well.

bstrie:
`enarx deploy` isn't modified because the plan is to merge `enarx deploy` into `enarx run` in the near future.

Closes #2281.

### DIFF
--- a/crates/exec-wasmtime/src/workload.rs
+++ b/crates/exec-wasmtime/src/workload.rs
@@ -42,21 +42,15 @@ pub enum Package {
     Remote(Url),
 
     /// Local package
-    #[cfg(unix)]
     Local {
         /// Open WASM module file descriptor
+        #[cfg(unix)]
         wasm: std::os::unix::prelude::RawFd,
-        /// Optional open config file descriptor
-        conf: Option<std::os::unix::prelude::RawFd>,
-    },
-
-    /// Local package
-    #[cfg(windows)]
-    Local {
         /// Open WASM module file
+        #[cfg(windows)]
         wasm: std::fs::File,
-        /// Optional open config file
-        conf: Option<std::fs::File>,
+        /// Enarx configuration
+        config: Config,
     },
 }
 
@@ -107,10 +101,7 @@ fn get_package(root: Entity<'_, impl Scope, scope::Node>, dir: TreeDirectory) ->
         *PACKAGE_CONFIG,
     );
     let config = toml::from_slice(&config).context("failed to parse config")?;
-    Ok(Workload {
-        webasm,
-        config: Some(config),
-    })
+    Ok(Workload { webasm, config })
 }
 
 /// Acquired workload
@@ -119,7 +110,7 @@ pub struct Workload {
     pub webasm: Vec<u8>,
 
     /// Enarx keep configuration
-    pub config: Option<Config>,
+    pub config: Config,
 }
 
 impl TryFrom<Package> for Workload {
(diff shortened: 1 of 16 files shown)


src/cli/deploy.rs
```diff
@@ -113,19 +113,23 @@ impl Options {
                     )
                 };
 
+                let config = match conf {
+                    Some(path) => toml::from_str(&read_to_string(path)?)?,
+                    None => Config::default(),
+                };
```
rvolosatovs, line 119: `match` is redundant here

crates/exec-wasmtime/src/workload.rs
```diff
@@ -42,21 +42,15 @@ pub enum Package {
...
-    /// Local package
-    #[cfg(windows)]
-    Local {
         /// Open WASM module file
+        #[cfg(windows)]
         wasm: std::fs::File,
-        /// Optional open config file
-        conf: Option<std::fs::File>,
+        /// Enarx configuration
+        config: Config,
```
rvolosatovs, line 53: Why rename this field?
Also note that @npmccallum likes variables/fields to be the same character length

rvolosatovs (commented):
I would expect to see a `merge` implemented on two configs here and internally it would probably use a `HashMap` or e.g. https://docs.rs/toml/latest/toml/value/type.Table.html

tests/client/mod.rs
```diff
@@ -206,7 +205,59 @@ async fn full() {
...
+        {example_dir}/main.wasm"#,
+        text:
+        "Hello, world!"
+    );
+
+    cmd!(
+        succeed: // when configuring via file and overriding via CLI
+        r#"enarx run
+        --wasmcfgfile {example_dir}/Enarx.toml
+        --with-files '[{{kind = "stdin"}}, {{kind = "stdin"}}, {{kind = "stdin"}}]'
```
rvolosatovs, line 254: How about we don't do that yet and wait until file management is reworked in #2245, which would allow this to be implemented in much cleaner manner?

bstrie: > I'm really not a fan of passing TOML on command-line

We've already determined that we need to pass config arguments on the command line somehow, and this seems like the least bad option. The alternatives are either using JSON, which can't be copy-pasted to/from the config file and is even more verbose, or having some custom syntax which has the same problem as well as the downside of forcing the user to learn it.

bstrie:
> I would expect to see a merge implemented on two configs here and internally it would probably use a HashMap or e.g. https://docs.rs/toml/latest/toml/value/type.Table.html

I proposed a scheme like this originally and it was rejected in favor of having a second Config with optional fields.

src/cli/deploy.rs
bstrie: As opposed to using `map`? Being able to leverage `?` here seems preferable.

dpal:
Let us wait for the VFS work to be complete and the Network policy.
We need to think about the problem more.

dpal:
What do we hash when we do the attestation? Do we normalize things?
File an issue for toml normalization.

bstrie:
Closing this PR for now. Additional discussion can be done in #2281. I've filed https://github.com/enarx/enarx/issues/2368 to track the issue of normalizing the TOML file.

PR #2311, opened 2022-11-09
feat(drawbridge): specify OIDC scopes and audience

puiterwijk:
Specify OpenID Connect audience and scopes, so the tokens are properly restricted to the environment/API they are intended for.


puiterwijk:
The server side is in https://github.com/profianinc/drawbridge/pull/312

puiterwijk:
As a note: this patch will get new tokens that are accepted by both the current and the patched Drawbridge server code, so this can go in without the server side needing to be live.

bstrie (approved):

PR #2354, opened 2022-11-17
feat(drawbridge): add client secrets login

puiterwijk:
This adds logging in with client credential grant for machine-2-machine interactions, like bots for uploading data to a Drawbridge instance.

### DIFF
--- a/src/cli/user/login.rs
+++ b/src/cli/user/login.rs
@@ -1,12 +1,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::drawbridge::login;
+use crate::drawbridge::{login, LoginContext, LoginCredentials};
 
 use std::ffi::OsString;
 
-use clap::Args;
+use anyhow::Context;
+use clap::{Args, ValueEnum};
 use oauth2::url::Url;
 
+#[derive(Debug, Clone, ValueEnum)]
+enum LoginMethod {
+    ClientCredentials,
+    DeviceAuthorization,
+}
 /// Log in to an Enarx package host and save credentials locally.
 #[derive(Args, Debug)]
 pub struct Options {
@@ -14,6 +20,9 @@ pub struct Options {
     oidc_domain: Url,
     #[clap(long, default_value = "4NuaJxkQv8EZBeJKE56R57gKJbxrTLG2")]
     oidc_client_id: String,
+    #[clap(long)]
+    #[arg(value_enum)]
+    login_method: LoginMethod,
     #[clap(long, env = "ENARX_CREDENTIAL_HELPER")]
     credential_helper: Option<OsString>,
     #[clap(long, default_value = "store.profian.com")]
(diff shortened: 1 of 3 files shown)


src/cli/user/login.rs
```diff
@@ -14,6 +14,8 @@ pub struct Options {
     oidc_domain: Url,
     #[clap(long, default_value = "4NuaJxkQv8EZBeJKE56R57gKJbxrTLG2")]
     oidc_client_id: String,
+    #[clap(long)]
+    oidc_client_secret: Option<String>,
```
rvolosatovs, line 18: Do we really want to expose this on the command-line?
Especially if this is to be used by CI, using an env variable for this is probably the lesser of all evils

src/drawbridge.rs
```diff
@@ -193,49 +195,53 @@ pub fn login(
...
-    .set_device_authorization_url(dev_auth_url);
+    .set_device_authorization_url(dev_auth_url))
+}
 
-    let details: StandardDeviceAuthorizationResponse = client
-        .exchange_device_code()
-        .context("Failed to construct device authorization request")?
-        .add_scope(Scope::new("openid".into()))
-        .add_scope(Scope::new("profile".into()))
+pub fn login_client_credentials(
```
rvolosatovs, line 208: I think there should only be one `login` function, which should simply be extended with additional parameter (`oidc_client_secret`)

That would reduce duplication and simplify maintenance and would prevent errors, like the one made in this PR already missing e.g. this case https://github.com/enarx/enarx/blob/09bce3fd923bbbb52aac185c5858e377b949f043/src/cli/user/register.rs#L56-L61

Since the list of parameters is getting long, we probably want an `AuthContext`, `OidcContext` or something of that sort be taken as an agument

rvolosatovs (changes requested):

src/drawbridge.rs
```diff
@@ -267,11 +273,54 @@ pub fn login(
             .set_password(secret)
             .context("Failed to save user credentials")?;
     }
+
```
rvolosatovs, line 276: Please avoid making unrelated diffs like this

src/drawbridge.rs
puiterwijk: Well, the thing is that there's basically no overlap between this and the current `login` function: the only line of code they share is the `let client = get_oauth2_client(...)`.
So then I'd basically turn both `login` (maybe rename to `login_device_flow`) and `login_client_credentials` into hidden functions, and just have the public `login` function call either of those.
Is that what you'd prefer?

src/drawbridge.rs
puiterwijk: You mean the extra blank line, or the refactoring to a generalized function?

src/cli/user/login.rs
puiterwijk: I personally think having the behaviour of a command radically change to where it does something pretty different (exchange client credentials for access token, instead of print a device key, waiting for that, and then exchanging that for an access token) because of the existence (or non-existence) of a "magical" environment variable is kinda irky.
At least as a CLI flag, it's very explicit.

But if you really prefer an env var, sure.

rvolosatovs: We should not encourage users to pass secrets on the command line.
This secret belongs in a config file (refs #2021 I suppose).
In CI, writing a config file is cumbersome and an env var is far more common and simpler to use.

Regarding the different behavior - I agree, so it seems like we need another CLI flag, which would enable `m2m` login. It should probably be an enum and default to something like `--login device`, `--login client` being the other option for now.

src/drawbridge.rs
rvolosatovs: > Well, the thing is that there's basically no overlap between this and the current `login` function: the only line of code they share is the `let client = get_oauth2_client(...)`.

That's not true, there's also
```
        .add_scope(Scope::new("manage:drawbridge_users".into()))
        .add_scope(Scope::new("manage:drawbridge_repositories".into()))
        .add_scope(Scope::new("manage:drawbridge_tags".into()))
        .add_extra_param("audience", format!("https://{}/", host))
```
block, for example, which I am much more concerned about, since it's much more likely to get out-of-sync as time goes on and result in a painful-to-debug bugs

Additionally, the `login` and `login_client_credentials` function distinction does not make sense to me.

As a follow-up on https://github.com/enarx/enarx/pull/2354/files#r1026469814
Let's add an enum:
```
enum AuthContext{
   Device{ // fields relevant to device flow },
   Client{ // fields relevant to client cred flow },
}
```
And pass that to `login`

Login can then match on the enum and act accordingly. There's no need to split it into further helper functions

puiterwijk: So, the problem with merging the scopes and params between the two is that both methods (`oauth2::Client::exchange_client_credentials` and `oauth2::Client::exchange_device_code`) return totally different structs (`oauth2::ClientCredentialsTokenRequest` and `oauth2::DeviceAuthorizationRequest` respectively), which happen to share those methods. So it's not really possible to extract that into common code.
However, I can use the `add_scopes` for both, and just use a statically set list of scopes, which at least forces those two to be in sync.

puiterwijk: I made that change right now.

PR #2355, opened 2022-11-17
chore(drawbridge): add user-agent to OIDC requests

puiterwijk:
Add a User-Agent to outgoing OpenID Connect requests, so we can track the source of authentication requests.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/src/drawbridge.rs
+++ b/src/drawbridge.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::borrow::Borrow;
+use std::convert::TryInto;
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::{stderr, Write};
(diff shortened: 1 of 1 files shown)


src/drawbridge.rs
```diff
@@ -179,6 +180,16 @@ pub fn client(
...
 }
 
+fn http_client(mut req: oauth2::HttpRequest) -> Result<oauth2::HttpResponse, oauth2::ureq::Error> {
+    req.headers.insert(
+        "User-Agent",
+        format!("enarx-cli/{}", env!("CARGO_PKG_VERSION"))
+            .try_into()
+            .unwrap(),
+    );
+    ureq_http_client(req)
```
rvolosatovs, line 190: Why not just `ureq::http_client` or `oauth2::ureq::http_client` to avoid having to rename the import?

puiterwijk: Don't know, that was just personal feeling. I fixed it now, thanks!

rvolosatovs (approved):

PR #2356, opened 2022-11-18
feat(drawbridge): add client secrets login

rvolosatovs:
This is based on top of and closes #2354 

Original description:
>This adds logging in with client credential grant for machine-2-machine interactions, like bots for uploading data to a Drawbridge instance.

Additionally, this adds an `oidc-flow` flag, which defaults to `device` and fixes few missed cases in #2354 (e.g. `enarx user register`)

- Support specifying a `client-secret-id` on `enarx user register`
- Reorganize code for clarity
- Remove redundant clones and borrows

### DIFF
--- a/src/cli/package/info.rs
+++ b/src/cli/package/info.rs
@@ -26,11 +26,11 @@ pub struct Options {
 impl Options {
     pub fn execute(self) -> anyhow::Result<()> {
         let cl = client(
-            &self.spec.host,
-            &self.oidc_domain,
-            &self.insecure_auth_token,
-            &self.ca_bundle,
-            &self.credential_helper,
+            self.spec.host,
+            self.oidc_domain,
+            self.insecure_auth_token,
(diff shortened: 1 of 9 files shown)


puiterwijk (approved):
This looks decent to me.

bstrie (changes requested):
How difficult would it be to add integration tests for the new CLI options? Would the OIDC mock need to be extended?

puiterwijk:
@bstrie The OIDC mock doesn't actually exercise the OIDC login flows: it starts with a pre-existing authentication token, and just skips over all of the code touched by this PR.

rvolosatovs:
> How difficult would it be to add integration tests for the new CLI options? Would the OIDC mock need to be extended?

Since you're in charge of CLI, please either file an issue about it and add these tests or add them here.
This is really just for internal use (at least for now)

bstrie (approved):

PR #2357, opened 2022-11-18
fix(exec-wasmtime): prevent workload FD from being dropped

### LINKED ISSUE #2352: Remove ManuallyDrop from wasmtime test runner
https://github.com/enarx/enarx/blob/19ea426405a79c664806d5804a1a62149f5b8a40/crates/exec-wasmtime/src/lib.rs#L107-L123

There has to be a cleaner solution for this!

Introduced by #2348

rvolosatovs:
Closes #2352 

~Here's a minimized example of what's happening https://github.com/rvolosatovs/fd-leak~

Refs https://github.com/rust-lang/rust/issues/104734

### DIFF
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -64,7 +64,7 @@ mod test {
 
     use std::io::{Seek, Write};
     #[cfg(unix)]
-    use std::os::unix::prelude::AsRawFd;
+    use std::os::unix::io::IntoRawFd;
 
     use anyhow::Context;
     use tempfile::tempfile;
@@ -109,17 +109,11 @@ mod test {
         file.write(wasm).context("failed to write module to file")?;
         file.rewind().context("failed to rewind file")?;
         #[cfg(unix)]
-        let file = std::mem::ManuallyDrop::new(file);
-        let res = Runtime::execute(Package::Local {
-            #[cfg(unix)]
-            wasm: file.as_raw_fd(),
-            #[cfg(windows)]
+        let file = file.into_raw_fd();
+        Runtime::execute(Package::Local {
             wasm: file,
             conf: None,
-        });
-        #[cfg(unix)]
-        let _ = std::mem::ManuallyDrop::into_inner(file);
-        res
+        })
     }
 
     #[test]


haraldh (dismissed):

rvolosatovs:
I will rework this with `into_raw_fd`

bstrie (approved):

haraldh:
fn [into_raw_fd](https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html#tymethod.into_raw_fd)(self) -> [RawFd](https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html)

Consumes this object, returning the raw underlying file descriptor.

This function is typically used to transfer ownership of the underlying file descriptor to the caller. When used in this way, callers are then the unique owners of the file descriptor and must close it once it’s no longer needed.

However, transferring ownership is not strictly required. Use a [`Into<OwnedFd>::into`](https://doc.rust-lang.org/std/convert/trait.Into.html#tymethod.into) implementation for an API which strictly transfers ownership.

haraldh:
Why don't we use `OwnedFd` then?

rvolosatovs:
> Why don't we use `OwnedFd` then?

Because we cannot pass an `OwnedFd` directly over the binary boundary.
Only way is to convert `OwnedFd` into a `RawFd` via [`IntoRawFd`](https://doc.rust-lang.org/std/os/unix/io/struct.OwnedFd.html#impl-IntoRawFd-for-OwnedFd)

We also cannot use `OwnedFd` in the tests, while passing `RawFd` from the host by using something like `impl Into<File>` in the function signature, because `File` does not implement `From<RawFd>` (it'd be a violation of Rust security primitives). Only [`FromRawFd`](https://doc.rust-lang.org/std/fs/struct.File.html#impl-FromRawFd-for-File)

So the best we could do is to refactor the implementation to use a generic [`IntoRawFd`](https://doc.rust-lang.org/std/os/unix/io/trait.IntoRawFd.html) in the struct https://github.com/enarx/enarx/blob/3302901ff2e639864d826f3ac7608ee2bed0b129/crates/exec-wasmtime/src/workload.rs#L45-L52, which could be a marginal improvement, since we could then directly pass the file in the test. The host serde implementation would still need to use `IntoRawFd::into_raw_fd`.
It would not change anything in actual production code flow, but would make the implementation more complex.
If you want I can make that change. I don't think we'd really gain anything that way, except that we'd not need to call `IntoRawFd::into_raw_fd` on the test workload `File` anymore (and instead would do it in the `<TryFrom<Package> for Workload>::try_from` https://github.com/enarx/enarx/blob/3302901ff2e639864d826f3ac7608ee2bed0b129/crates/exec-wasmtime/src/workload.rs#L196-L224

Issue #2367, opened 2022-11-29
[Feature]: Networking policy


rvolosatovs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

`enarx` network policy should be configurable, which is especially relevant for runtime networking streams and sockets.
When nothing is specified, `enarx` should assume safe defaults. This is a sane default in production setting, but does not work well for testing and development.
We already have #1548 for some limited TLS configurability
We also have https://github.com/enarx/enarx/issues/2170 to specify peer certificates to expect.
We need to do more than that:
- Eventually, Enarx will have multiple certs available, there should be a way to specify which certificate to use for which connection
- There should be a way to set trust anchors and define assertions on peer certificates for incoming and outgoing connections
- There should be a way to say that all connections on ports in range `ABC-XYZ` should use TCP (TLS is the default)

dpal:
We need a design proposal for the config.

dpal:
The policy should support certificate pinning (remote cert must have a specific subject line).

PR #2369, opened 2022-11-29
feat(exec-wasmtime): outgoing TLS connections during runtime

Trundle:
This PR adds a host function `tls_client_connect` to `exec-wasmtime`. It allows to open TLS client connections during runtime inside a Keep. Currently, the function always expects a client certificate. This could be changed, e.g. to default to Enarx's certificate created on startup if a key/cert of length 0 is given.

[With some minor modifications of `hyper`](https://github.com/hyperium/hyper/compare/0.14.x...yaxitech:hyper:wasi), it's possible to create a custom connector and then use hyper as HTTP client in a Rust guest.

The tests check that a connection to `enarx.dev` is possible. This assumes that online availability is given during test execution.

I understand that this is a bit of a hack and not a long-term solution. @npmccallum encouraged me to open a PR nevertheless.

Signed-off-by: Andreas Stührk <andy@yaxi.tech>


rvolosatovs (changes requested):
Thanks for contribution @Trundle, but this feature will be available shortly via VFS (#2245 )

You will be able to open TLS connections by opening a file in `/net/con/example.com`, to connect to `example.com` (exact paths/API still TBD)

`Enarx.toml` will contain a networking policy defining the behavior #2367 

Would that fulfill your use case? Please reach out to me on rocket chat if you want to get a draft to test

rvolosatovs:
@Trundle https://github.com/enarx/enarx/pull/2245 is now ready to be tested. See integration tests for examples

Trundle:
Thank you for the update - I didn't know that VFS support was so close.

I'm not committed to having a function. For our use-case, we need a way to have (dynamic) outgoing TLS connections inside a Keep with a client certificate that is not under our control (it's issued by a 3rd party). So doing it through VFS is absolutely fine for me. Unfortunately, in its current state,  the VFS implementation seems to only allow a Steward-issued/self-signed client certificate (if I read the code correctly).

rvolosatovs:
> Thank you for the update - I didn't know that VFS support was so close.
> 
> I'm not committed to having a function. For our use-case, we need a way to have (dynamic) outgoing TLS connections inside a Keep with a client certificate that is not under our control (it's issued by a 3rd party). So doing it through VFS is absolutely fine for me. Unfortunately, in its current state, the VFS implementation seems to only allow a Steward-issued/self-signed client certificate (if I read the code correctly).

Yes, peer certificates will be part of networking policy #2170 
If that's blocking for you, then I will try my best to prioritize this work. We need to work out the design for this feature first and we would greatly appreciate any feedback on it. I'll ping you once I have something working!

Sounds like your use case is that you have a client cert and key persisted on disk and you want Enarx to use that keypair for all outgoing connections, is that right?

PR #2375, opened 2022-12-06
feat(sgx): add a page fault handler to do lazy mmap

haraldh:
This will allow to map the whole 2GB address range of the wasm application, which will speedup execution a lot.

### DIFF
--- a/src/backend/sgx/thread.rs
+++ b/src/backend/sgx/thread.rs
@@ -6,12 +6,12 @@ use crate::backend::execute_gdb;
 use crate::backend::Command;
 
 use std::arch::asm;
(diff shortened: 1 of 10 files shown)


src/backend/sgx/thread.rs
```diff
@@ -171,9 +190,9 @@ impl super::super::Thread for Thread {
         //
         // After the sallyport rework, we can test the sallyport itself and
         // remove this logic.
```
bstrie, line 192: "After the sallyport rework", eh? :P

haraldh: Interesting

Issue #1042, opened 2021-10-21
Allow `--log-target` and `--log-filter` settings to apply inside the keep


wgwoods:
The `--log-target` and `--log-filter` CLI arguments set logging parameters for the CLI itself, but those settings are lost once we enter the keep, because we don't have a way to pass those arguments into the keep.

(We may also need to add something like `--keep-log-target` so that the code inside the keep can choose a log target that doesn't go to the host system - but that's a different issue.)

rvolosatovs:
Refs #2398

PR #2391, opened 2022-12-14
feat: add flame graph

### LINKED ISSUE #2176: [Feature]: Benchmarks and Profiling Support

### Description

Add benchmarking and profiling to Enarx so developers can learn more about their application's performance.

### Acceptance Criteria

Ideas:
* Add a stopwatch to `exec-wasmtime` which shows how long a program ran when it exits, enabled by a flag. This would show the program's runtime without any start-up or attestation time from Enarx.
* Add profiling which can be viewed live from a browser, similar to Go's [net/http/pprof](https://pkg.go.dev/net/http/pprof).
  * Memory profiling
  * CPU profiling
  * Anything else?

### Suggestions for a technical implementation

Checkout:
* https://nnethercote.github.io/perf-book/profiling.html
* https://github.com/flamegraph-rs/flamegraph

ishme-al:
partially closes #2176 
Refs https://github.com/enarx/enarx/issues/1042 (the `log-filter` part, `log-target` is still TODO)

produce flame graph with (requires inferno crate)
`cat tracing.folded | inferno-flamegraph > tracing-flamegraph.svg;`

tracing-flamegraph

notes:
add more functions to flame graph?
logging might have been broken (enarx -v)
includes wiggle

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ target/
 output/
 .idea/
 .vscode/
-tracing.folded
-*.svg
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ target/
 output/
 .idea/
 .vscode/
+tracing.folded
+*.svg
--- a/crates/exec-wasmtime/Cargo.toml
+++ b/crates/exec-wasmtime/Cargo.toml
@@ -7,6 +7,9 @@ authors = ["The Enarx Project Developers"]
 repository = "https://github.com/enarx/enarx"
 license = "Apache-2.0"
 
+[features]
+bench = ["dep:tracing-flame"]
+
 [dependencies]
 anyhow = { workspace = true }
 cap-std = { workspace = true }
@@ -41,7 +44,7 @@ zeroize = { workspace = true }
 io-extras = { workspace = true }
 
 [target.'cfg(unix)'.dependencies]
-tracing-flame = { workspace = true }
+tracing-flame = { workspace = true, optional = true }
 
 [target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
 sallyport = { workspace = true }
--- a/crates/exec-wasmtime/Cargo.toml
+++ b/crates/exec-wasmtime/Cargo.toml
@@ -40,6 +40,9 @@ zeroize = { workspace = true }
 [target.'cfg(windows)'.dependencies]
 io-extras = { workspace = true }
 
+[target.'cfg(unix)'.dependencies]
+tracing-flame = { workspace = true }
+
 [target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
 sallyport = { workspace = true }
 
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -54,7 +54,7 @@ pub fn execute() -> anyhow::Result<()> {
     use anyhow::Context;
     use tracing::level_filters::LevelFilter;
     use tracing_subscriber::prelude::*;
-    use tracing_subscriber::{fmt, registry};
+    use tracing_subscriber::registry;
 
     // This is the FD of a Unix socket on which the host will send the TOML-encoded execution arguments
     // and shutdown the write half of it immediately after.
@@ -87,9 +87,12 @@ pub fn execute() -> anyhow::Result<()> {
     };
     let log_level: LevelFilter = log_level.map(Into::into).into();
     {
-        // TODO: Default to a secure log target
-        // https://github.com/enarx/enarx/issues/1042
-        let registry = registry().with(fmt::layer().with_filter(log_level));
+        let fmt_layer = tracing_subscriber::fmt::layer()
+            // TODO: Default to a secure log target
+            // https://github.com/enarx/enarx/issues/1042
+            .with_writer(std::io::stderr)
+            .with_filter(log_level);
+        let registry = registry().with(fmt_layer);
         #[cfg(feature = "bench")]
         let registry = registry.with(flame_layer);
         let _guard = registry.set_default();
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -30,7 +30,7 @@ pub struct Args {
     pub log_level: Option<log::Level>,
 
     /// Profile
-    #[cfg(unix)]
+    #[cfg(all(unix, feature = "bench"))]
     pub profile: Option<std::os::unix::prelude::RawFd>,
 
     /// Package
@@ -46,7 +46,6 @@ pub fn execute_package(pkg: Package) -> anyhow::Result<()> {
 /// Execute with arguments read from file descriptor 3.
 #[cfg(unix)]
 pub fn execute() -> anyhow::Result<()> {
-    use std::fs::File;
     use std::io::Read;
     use std::mem::forget;
     use std::os::unix::io::FromRawFd;
@@ -54,7 +53,6 @@ pub fn execute() -> anyhow::Result<()> {
 
     use anyhow::Context;
     use tracing::level_filters::LevelFilter;
-    use tracing_flame::FlameLayer;
     use tracing_subscriber::prelude::*;
     use tracing_subscriber::{fmt, registry};
 
@@ -72,13 +70,16 @@ pub fn execute() -> anyhow::Result<()> {
 
     let Args {
         log_level,
+        #[cfg(feature = "bench")]
         profile,
         package,
     } = toml::from_str(&args).context("failed to decode arguments")?;
 
+    #[cfg(feature = "bench")]
     let (flame_layer, _guard) = if let Some(profile) = profile {
+        use std::fs::File;
         let profile = unsafe { File::from_raw_fd(profile) };
-        let flame_layer = FlameLayer::new(profile);
+        let flame_layer = tracing_flame::FlameLayer::new(profile);
         let guard = flame_layer.flush_on_drop();
         (Some(flame_layer), Some(guard))
     } else {
@@ -86,10 +87,12 @@ pub fn execute() -> anyhow::Result<()> {
     };
     let log_level: LevelFilter = log_level.map(Into::into).into();
     {
-        let _guard = registry()
-            .with(fmt::layer().with_filter(log_level))
-            .with(flame_layer)
-            .set_default();
+        // TODO: Default to a secure log target
+        // https://github.com/enarx/enarx/issues/1042
+        let registry = registry().with(fmt::layer().with_filter(log_level));
+        #[cfg(feature = "bench")]
+        let registry = registry.with(flame_layer);
+        let _guard = registry.set_default();
         execute_package(package)?;
     }
     Ok(())
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -17,6 +17,8 @@ pub use workload::{Package, Workload, PACKAGE_CONFIG, PACKAGE_ENTRYPOINT};
 
 use runtime::Runtime;
 
+use wiggle::tracing::instrument;
+
 /// The Arguments
 // NOTE: `repr(C)` is required, otherwise `toml` serialization fails with `values must be emitted before tables`
 #[derive(Debug)]
@@ -36,6 +38,7 @@ pub struct Args {
 }
 
 /// Execute package
+#[instrument]
 pub fn execute_package(pkg: Package) -> anyhow::Result<()> {
     Runtime::execute(pkg).map(|_| ())
 }
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -6,9 +6,13 @@
 #![deny(clippy::all)]
 #![warn(rust_2018_idioms)]
 
+#[cfg(unix)]
+mod log;
 mod runtime;
 mod workload;
 
+#[cfg(unix)]
+pub use log::Level as LogLevel;
 pub use workload::{Package, Workload, PACKAGE_CONFIG, PACKAGE_ENTRYPOINT};
 
 use runtime::Runtime;
@@ -19,25 +23,37 @@ use runtime::Runtime;
 #[cfg_attr(unix, derive(serde::Deserialize, serde::Serialize))]
 #[repr(C)]
 pub struct Args {
+    /// Log level
+    #[cfg(unix)]
+    pub log_level: Option<log::Level>,
+
+    /// Profile
+    #[cfg(unix)]
+    pub profile: Option<std::os::unix::prelude::RawFd>,
+
     /// Package
     pub package: Package,
 }
 
-/// Execute
-pub fn execute_with_args(args: Args) -> anyhow::Result<()> {
-    Runtime::execute(args.package).map(|_| ())
+/// Execute package
+pub fn execute_package(pkg: Package) -> anyhow::Result<()> {
+    Runtime::execute(pkg).map(|_| ())
 }
 
-/// Execute
-///
-/// with configuration read from file descriptor 3.
+/// Execute with arguments read from file descriptor 3.
 #[cfg(unix)]
 pub fn execute() -> anyhow::Result<()> {
-    use anyhow::Context;
+    use std::fs::File;
     use std::io::Read;
     use std::mem::forget;
     use std::os::unix::io::FromRawFd;
     use std::os::unix::net::UnixStream;
+
+    use anyhow::Context;
+    use tracing::level_filters::LevelFilter;
+    use tracing_flame::FlameLayer;
+    use tracing_subscriber::prelude::*;
+    use tracing_subscriber::{fmt, registry};
 
     // This is the FD of a Unix socket on which the host will send the TOML-encoded execution arguments
     // and shutdown the write half of it immediately after.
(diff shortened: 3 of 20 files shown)


crates/exec-wasmtime/src/lib.rs
```diff
@@ -42,7 +42,7 @@ pub fn execute() -> anyhow::Result<()> {
     // This is the FD of a Unix socket on which the host will send the TOML-encoded execution arguments
     // and shutdown the write half of it immediately after.
     // TODO: Use the write half of the socket to write logs/errors to the host
-    let mut host = unsafe { UnixStream::from_raw_fd(3) };
+    let mut host = unsafe { UnixStream::from_raw_fd(4) };
```
rvolosatovs, line 45: We shouldn't change this

rvolosatovs (changes requested):
Good start!

I don't think we should change the FD number expected by `exec-wasmtime`, a simple trick we could use to ensure FD 3 stays "reserved"
```diff
diff --git a/crates/exec-wasmtime/src/lib.rs b/crates/exec-wasmtime/src/lib.rs
index a2c03384..345ee6e5 100644
--- a/crates/exec-wasmtime/src/lib.rs
+++ b/crates/exec-wasmtime/src/lib.rs
@@ -42,7 +42,7 @@ pub fn execute() -> anyhow::Result<()> {
     // This is the FD of a Unix socket on which the host will send the TOML-encoded execution arguments
     // and shutdown the write half of it immediately after.
     // TODO: Use the write half of the socket to write logs/errors to the host
-    let mut host = unsafe { UnixStream::from_raw_fd(4) };
+    let mut host = unsafe { UnixStream::from_raw_fd(3) };
 
     let mut args = String::new();
     host.read_to_string(&mut args)
diff --git a/src/cli/mod.rs b/src/cli/mod.rs
index b325a9a9..07a65a93 100644
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -16,6 +16,7 @@ mod user;
 
 use crate::backend::{Backend, BACKENDS};
 
+use std::fs::File;
 use std::ops::Deref;
 use std::process::ExitCode;
 use std::str::FromStr;
@@ -178,8 +179,12 @@ impl LogOptions {
             .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
             .with_filter(env_filter);
 
+        let file = File::open("/dev/null");
+
         let (flame_layer, _guard) = FlameLayer::with_file("./tracing.folded").unwrap();
 
+        drop(file);
+
         tracing_subscriber::registry()
             .with(fmt_layer)
             .with(flame_layer)
diff --git a/src/exec/mod.rs b/src/exec/mod.rs
index 3215f926..236d32ee 100644
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -163,7 +163,7 @@ pub fn run_package(
 
     assert_eq!(
         exec_sock.as_raw_fd(),
-        4, // The first free FD after STDOUT, STDIN and STDERR are assigned.
+        3, // The first free FD after STDOUT, STDIN and STDERR are assigned.
         "exec-wasmtime expects the Unix socket to be at FD 3"
     );
 ```
 
 There should be better way to do this and we should only attempt to open "something" on `unix`. This is just a quick sketch that passes tests
 
 Let's also make the flame graph optional - there should be a CLI flag, or even better, both a CLI flag and a feature

**Comment by rvolosatovs** [comment]
@ish-meal I've pushed a commit reworking the initialization a little bit and added a `--profile` flag to CLI.
I don't get any output in the specified, however, but I also didn't get any output with original implementation anyway.

I am not sure if there's some environment variable I have to specify or is that an implementation issue?

File: .gitignore [hunk 2]
```diff
@@ -2,3 +2,5 @@ target/
 output/
 .idea/
 .vscode/
+tracing.folded
```
rvolosatovs, line 5: Since the `--profile` now takes in a path, this should not be necessary

.gitignore
```diff
@@ -2,3 +2,5 @@ target/
...
+*.svg
```
rvolosatovs, line 6: I don't think we should add this, seems more of a local config

rvolosatovs (commented):
Not getting the graph generated, is there some environment variable that needs to be set perhaps (if so, it should be documented) or maybe a feature is missing in `Cargo.toml`?

.gitignore
ish-meal: I'm confused about how to pass the path in

i run --profile tempfile
and i get
error: profile 'tempfile' is not defined

what am i doing wrong?

rvolosatovs: Try `cargo run -- --profile <path> deploy examples/fibonacci-c:0.3.0`

haraldh (changes requested):
I want this to be guarded behind a feature and set the `DEBUG` flag for the keeps like https://github.com/enarx/enarx/blob/7b47875530658efbe8ec9419eb341dba32c8a5cc/crates/shim-sgx/src/lib.rs#L56-L58

Edit: ok, or not

ish-meal:
looks like naming the guard "_" was the issue (i guess it's completely ignored? we do want to hold on to it)

rvolosatovs:
> looks like naming the guard "_" was the issue (i guess it's completely ignored? we do want to hold on to it)

Could be, thanks, I learned something today! :)

src/cli/mod.rs
```diff
@@ -48,7 +50,38 @@ pub struct Options {
...
+            // Open `env::tempdir()` to reserve fd 3 on Unix, which `exec-wasmtime` will expect to read config from
+            // It will be dropped at the end of this block, freeing it for the following socketpair call
+            #[cfg(unix)]
+            let _reserved = matches!(
+                self.cmd,
+                Subcommands::Run(_)
+                    | Subcommands::Deploy(_)
+                    | Subcommands::Unstable(unstable::Subcommands::Exec(_))
+            )
+            .then_some(std::fs::File::open(std::env::temp_dir()))
```
rvolosatovs, line 70: I still feel like we can do better, is there some path that is guaranteed to exist cross-platform and is fast to open?
What if we try and open `current_exe` or `current_dir`? Those are probably more likely to be there than `temp_dir`, although potentially still could be deleted once the executable is started?

@haraldh any ideas?

`/proc/self/exe` comes to mind, but I am not sure if that's really generic across all Unix'es

Maybe something to do with memory?

Was `/dev/null` used in the initial iteration the best solution?

All we really need is to get an fd 3 assigned by the kernel, we don't really care about anything on FS here.

Should we just open the `profile` path twice?

haraldh: /dev/null is fine

rvolosatovs:
By some weird reason I failed to reproduce on this locally or on a lab machine - all I get is an empty profile file, but @ish-meal did just show it working in WSL.
Not sure if that's something to do with the OS, maybe some weird tracing crate bug? I'll take over this to get it over the line and will add a GitHub action to generate the profile as well

rvolosatovs:
Ah... The issue is the backend - on WSL/mac `nil` backend, which means that all `tracing` events occur within the same binary, but in case of `kvm`/`sgx`/`sev` backend, the `exec-wasmtime` emitting these attributes is executed as a separate binary and these events do not bleed through the binary boundary.
I'll see how to work around this

rvolosatovs:
Here's an example SVG
tracing-flamegraph

nickvidal:
Looks great, thanks!

rvolosatovs:
Looks like this is also blocked on https://github.com/rust-lang/cargo/pull/11478

crates/exec-wasmtime/Cargo.toml
```diff
@@ -7,6 +7,9 @@ authors = ["The Enarx Project Developers"]
 repository = "https://github.com/enarx/enarx"
 license = "Apache-2.0"
 
+[features]
+bench = ["dep:tracing-flame"]
```
bstrie, line 11: This is not a default feature, and I don't see this feature being enabled for the exec-wasmtime dependency in enarx's Cargo.toml. Where is this enabled? Is it intended to be manually specified when running benchmarks only on this crate?

crates/exec-wasmtime/Cargo.toml
```diff
@@ -40,6 +43,9 @@ zeroize = { workspace = true }
 [target.'cfg(windows)'.dependencies]
 io-extras = { workspace = true }
 
+[target.'cfg(unix)'.dependencies]
```
bstrie, line 46: Is this cfg necessary? I believe we treat the normal, un-cfg'd dependency list as the "unix" dependencies.

crates/exec-wasmtime/src/lib.rs
```diff
@@ -51,10 +68,36 @@ pub fn execute() -> anyhow::Result<()> {
...
-
+    let Args {
+        log_level,
+        #[cfg(feature = "bench")]
+        profile,
+        package,
+    } = toml::from_str(&args).context("failed to decode arguments")?;
+
+    #[cfg(feature = "bench")]
+    let (flame_layer, _guard) = if let Some(profile) = profile {
```
bstrie, line 79: I would prefer `let foo = match bar {` over `let foo = if let Some(foo) = bar {` in cases like this where you handle all the variants anyway.

crates/exec-wasmtime/src/lib.rs
```diff
@@ -51,10 +68,36 @@ pub fn execute() -> anyhow::Result<()> {
...
+    {
+        let fmt_layer = tracing_subscriber::fmt::layer()
+            // TODO: Default to a secure log target
+            // https://github.com/enarx/enarx/issues/1042
+            .with_writer(std::io::stderr)
+            .with_filter(log_level);
+        let registry = registry().with(fmt_layer);
+        #[cfg(feature = "bench")]
+        let registry = registry.with(flame_layer);
+        let _guard = registry.set_default();
```
bstrie, line 98: What are these two guards guarding, and why are we having one shadow the other?

src/cli/mod.rs
```diff
@@ -48,12 +52,82 @@ pub struct Options {
...
+                match self.logger.log_target {
+                    LogTarget::Stdout => Box::new(io::stdout()),
+                    LogTarget::Stderr => Box::new(io::stderr()),
+                }
+            })
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .with_filter(env_filter);
+
+        #[cfg(feature = "bench")]
+        let (flame_layer, _guard, profile) = if let Some(ref profile) = self.logger.profile {
```
bstrie, line 74: Same suggestion for `match` as above.

crates/exec-wasmtime/Cargo.toml
rvolosatovs: I don't get what you mean, this will never be used on anything but Unix, why include it on e.g. Windows?

crates/exec-wasmtime/src/lib.rs
rvolosatovs: The outermost needs to be dropped to ensure the profile is written to the file. Innermost reverts the log setting once it's dropped, which is relevant for nil backends where the log may have been previously initialized

crates/exec-wasmtime/Cargo.toml
bstrie: I see `enarx-exec-wasmtime/bench` there, but shouldn't it also have `enarx-shim-kvm/bench` and `enarx-shim-sgx/bench`?

crates/exec-wasmtime/src/lib.rs
rvolosatovs: I disagree with this, because that would add an unnecessary indentation level

src/cli/mod.rs
rvolosatovs: same as above, I don't think we need to add another indentation level here. what's the benefit?

crates/exec-wasmtime/Cargo.toml
bstrie: Why is it excluded from Windows in the first place?

rvolosatovs: Because there's no shim on non-unix platforms, it runs within the same binary and logger is only ever initialized once, on the host

crates/exec-wasmtime/src/lib.rs
bstrie: My criterion for when to use `match` over `if let` is whenever I'm handling all of the variants of an enum. I find it to be more readable, despite the extra indentation level. Do you not use `match` at all?

crates/exec-wasmtime/Cargo.toml
rvolosatovs: On Unix platforms, `execute` is used, which will set up a logger for the duration of execution of `exec-wasmtime` and set it back to previous value once execution is done.
On non-Unix platforms, `execute_package` is called directly and, therefore, no logger initialization within `exec-wasmtime` is taking place.
This is working as intended.

https://github.com/enarx/enarx/blob/462c8821a3db8ed1b765c5101c07ced89610950c/src/backend/nil.rs#L105-L115

crates/exec-wasmtime/src/lib.rs
rvolosatovs: I use `match` when there are more than two cases to match. Matching an `Option` using a `match` is redundant in my opinion. That's also why multitude of helpers exist for handling this (`.map*`, `.unwrap*` etc).

rvolosatovs: There are also plenty of `let x = if let Some(..) = {} else {}` examples in our codebase

rvolosatovs:
```
$ rg 'if let Some\('
tests/integration.rs
151:    if let Some(stdin) = stdin {
244:    if let Some(expected_stdout) = expected_stdout {
256:    if let Some(expected_stderr) = expected_stderr {

tests/client/util.rs
88:    if let Some(msg) = failed_test {

tests/crates/enarx_syscall_tests/src/lib.rs
281:    let arg1 = if let Some(ref nonce) = nonce {
287:    let arg0 = if let Some(nonce) = nonce {
293:    let arg3 = if let Some(ref buf) = buf {
299:    let arg2 = if let Some(buf) = buf {

tests/crates/enarx_exec_tests/src/bin/sgx_attestation.rs
47:    let nonce_ptr = if let Some(ref nonce) = nonce {
53:    let nonce_len = if let Some(ref nonce) = nonce {
59:    let buf_len = if let Some(ref buf) = buf {
65:    let buf_ptr = if let Some(buf) = buf {

tests/crates/enarx_exec_tests/src/bin/sev_attestation.rs
134:    let arg1 = if let Some(ref nonce) = nonce {
140:    let arg0 = if let Some(nonce) = nonce {
146:    let arg3 = if let Some(ref buf) = buf {
152:    let arg2 = if let Some(buf) = buf {

tests/wasm/mod.rs
37:            if let Some(conf) = conf {

crates/shim-kvm/src/hostcall.rs
688:    if let Some(mut host_call) = HostCall::try_new(SHIM_LOCAL_STORAGE.write().deref_mut()) {

crates/shim-kvm/src/debug.rs
133:        if let Some(rip_rbp) = rbp.checked_add(size_of::<usize>() as _) {
138:            if let Some(rip) = rip
202:        if let Some(rip_rbp) = rbp.checked_add(size_of::<usize>() as _) {
211:            if let Some(rip) = rip.checked_sub(1) {
215:                if let Some(rip) = rip.checked_sub(shim_offset) {
219:                    if let Some(rip) = rip.checked_sub(EXEC_VIRT_ADDR.read().as_u64() as _) {

crates/shim-sgx/src/handler/mod.rs
345:        if let Some(addr) = addr {
472:                if let Some(block) = block {

crates/exec-wasmtime/src/workload.rs
89:    let entry = if let Some(entry) = dir.get(&PACKAGE_CONFIG) {
211:                let config = if let Some(conf) = conf.as_mut() {

crates/exec-wasmtime/src/lib.rs
79:    let (flame_layer, _guard) = if let Some(profile) = profile {

crates/exec-wasmtime/src/runtime/mod.rs
53:        let certs = if let Some(url) = steward {

crates/sallyport/src/item/block.rs
20:        let item_size = if let Some(item_size) = item_count.checked_mul(LARGEST_ITEM_SIZE) {
26:        let size = if let Some(size) = item_size.checked_add(data_size) {

crates/sallyport/src/guest/handler.rs
580:        if let Some(oldact) = oldact {
583:        if let Some(act) = act {
982:                    if let Some(oldact) = oldact {

crates/sallyport/src/guest/call/syscall/ioctl.rs
54:        let argp = if let Some(argp) = self.0.argp {

crates/sallyport/src/guest/call/syscall/nanosleep.rs
42:        let (rem, rem_offset) = if let Some(rem) = self.rem {

crates/sallyport/src/guest/call/enarxcall/park.rs
29:        let (timeout, timeout_offset) = if let Some(timeout) = self.timeout {

crates/sallyport/src/guest/call/syscall/clock_getres.rs
27:        let (res, res_offset) = if let Some(res) = self.res {

src/drawbridge.rs
140:    if let Some(token) = provided_token {
142:    } else if let Some(helper) = helper {
153:        } else if let Some(code) = output.status.code() {
182:    if let Some(ca_bundle_path) = ca_bundle {
232:    if let Some(helper) = helper {
252:            if let Some(code) = output.code() {

src/backend/kvm/thread.rs
163:                            if let Some(Item::Enarxcall(enarxcall, data)) =

src/cli/sign.rs
140:        let binary = if let Some(ref path) = self.binpath {
155:            let exec = if let Some(ref e) = binary {
157:            } else if let Some(e) = EXECS.iter().find(|w| w.with_backend(backend)) {
185:        if let Some(path) = self.out {

src/exec/mod.rs
118:    if let Some(conf) = conf {

src/backend/sev/snp/launch/linux.rs
138:            ma_uaddr: if let Some(addr) = start.ma_uaddr {

src/cli/mod.rs
73:        let (flame_layer, _guard, profile) = if let Some(ref profile) = self.logger.profile {
172:        if let Some(ref name) = self.backend {

src/cli/key/sgx/create.rs
31:        if let Some(path) = self.out {

src/cli/key/sgx/digest.rs
52:        if let Some(path) = self.out {

src/backend/sgx/enarxcall.rs
45:                    if let Some(thread) = keep.clone().spawn()? {
59:                if let Some(mut thread) = thread {

src/cli/key/sev/create.rs
30:        if let Some(path) = self.out {

src/backend/sev/hasher.rs
42:        if let Some(src) = src {

src/cli/platform/info.rs
120:                if let Some(ref info) = datum.info {
127:                if let Some(mesg) = datum.mesg.as_ref() {

src/backend/sgx/builder.rs
135:        let signature = if let Some(signatures) = builder.cnfg.signatures {

src/cli/key/sev/digest.rs
65:        if let Some(path) = self.out {

src/backend/sev/builder.rs
206:        let finish = if let Some(signatures) = signatures {
```

bstrie (approved):

PR #2395, opened 2022-12-15
build: make commissery no longer check PR titles

bstrie:
We have no need for this feature and it causes no end of annoyance.


rjzak (approved):

dpal (approved):
This looks OK but please let me know when you are making changes to the workflows that will be shared.

PR #2396, opened 2022-12-16
feat(sgx): handle CSSA 0-3 with 2 sallyport blocks

haraldh:
CSSA 1 and 2 get their own block.
CSSA 0 and 3 don't need a block.

Normal block usage looks like this:
* Enclave CSSA 0 fills in registers and executes syscall op.
* Host sees exception, calls CSSA 1 with sallyport block 0.
  * Enclave CSSA 1 fills in sallyport block, executes syscall op.
  * Host sees exception, handles sallyport block 0 and calls CSSA 2.
    * Enclave CSSA 2 confirms CSSA 1 syscall op and increases instruction pointer of CSSA 1. Exits to the host.
    * Host sees clean EEXIT of CSSA 2, calls CSSA 1.
  * Enclave CSSA 1 continues after syscall op and handles sallyport block 0, then increases CSSA 0 instruction pointer and   exits to the host.
  * Host sees clean EEXIT of CSSA 1, calls CSSA 0.
* Enclave CSSA 0 handles result of syscall.

Now, if there was lazy memory allocation for CSSA 0, with page fault handling, the following situation could occur: While enclave CSSA 1 handles the result of sallyport block 0 and copies data back to the userspace memory, a page fault could occur. Therefore we need a second sallyport block, which could be uses by CSSA 2 to handle the page fault occuring in CSSA 1.

This situation looks like this:
* Enclave CSSA 0 fills in registers and executes syscall op.
* Host sees exception, calls CSSA 1 with sallyport block 0.
  * Enclave CSSA 1 fills in sallyport block, executes syscall op.
  * Host sees exception, handles sallyport block 0 and calls CSSA 2.
    * Enclave CSSA 2 confirms CSSA 1 syscall op and increases instruction pointer of CSSA 1. Exits to the host.
    * Host sees clean EEXIT of CSSA 2, calls CSSA 1.
  * Enclave CSSA 1 continues after syscall op and handles sallyport block 0, and encountes a page fault for touching not yet mapped memory of CSSA 0.
  * Host sees page fault exception, calls CSSA 2.
    * CSSA 2 sees page fault exception, does some memory operations using sallyport block 1 and syscall op.
    * Host sees syscall exception, handles sallyport block 1 and calls CSSA 3.
      * Enclave CSSA 3 confirms CSSA 2 syscall op and increases instruction pointer of CSSA 2. Exits to the host.
      * Host sees clean EEXIT of CSSA 3, calls CSSA 2.
    * Enclave CSSA 2 continues after syscall op and handles sallyport block 1, then increases CSSA 1 instruction pointer and exits to the thost.
    * Host sees clean EEXIT of CSSA 2, calls CSSA 1.
  * Enclave CSSA 1 continues handling sallyport block 0, now that the page
  is correctly mapped, then increases CSSA 0 instruction pointer and exits
  to the host.
  * Host sees clean EEXIT of CSSA 1, calls CSSA 0.
* Enclave CSSA 0 handles result of syscall.

### DIFF
--- a/crates/shim-sgx/src/main.rs
+++ b/crates/shim-sgx/src/main.rs
@@ -26,8 +26,14 @@ use enarx_shim_sgx::thread::{
 };
 use enarx_shim_sgx::{
     entry, handler, shim_address, ATTR, BLOCK_SIZE, CSSA_0_STACK_SIZE, ENARX_EXEC_START,
-    ENARX_SHIM_ADDRESS, ENCL_SIZE, ENCL_SIZE_BITS, MISC,
+    ENARX_SHIM_ADDRESS, ENCL_SIZE, ENCL_SIZE_BITS, MISC, NUM_SSA,
 };
+use noted::noted;
+use primordial::Page;
+use sallyport::util::ptr::is_aligned_non_null;
+use sallyport::{elf::note, REQUIRES};
+use sgx::parameters::{Attributes, MiscSelect};
+use sgx::ssa::{GenPurposeRegs, StateSaveArea};
 
 #[panic_handler]
 #[cfg(not(test))]
@@ -37,12 +43,6 @@ fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
 }
 
 // ============== REAL CODE HERE ===============
-
-use noted::noted;
-use primordial::Page;
-use sallyport::{elf::note, REQUIRES};
-use sgx::parameters::{Attributes, MiscSelect};
-use sgx::ssa::{GenPurposeRegs, StateSaveArea};
 
 noted! {
     static NOTE_REQUIRES<note::NAME, note::REQUIRES, [u8; REQUIRES.len()]> = REQUIRES;
@@ -275,19 +275,13 @@ global_asm!(
         CSSA_0_STK_TCS_SZ = const CSSA_0_STACK_SIZE + Page::SIZE,
 );
 
-unsafe extern "C" fn main(
-    block: &mut [usize; BLOCK_SIZE / core::mem::size_of::<usize>()],
-    ssas: &mut [StateSaveArea; 3],
-    cssa: usize,
-    tcb: &mut MaybeUninit<Tcb>,
-) -> i32 {
-    // Enable exceptions:
-    ssas[cssa].extra[0] = 1;
+fn validate_block_ptr(ptr: *mut u8) -> &'static mut [usize; BLOCK_SIZE / size_of::<usize>()] {
+    is_aligned_non_null::<[usize; BLOCK_SIZE / size_of::<usize>()]>(ptr as usize).unwrap();
 
     // As host is a separate application from the shim, all the data coming from
     // it needs to be validated explicitly.  Thus, check that the Sallyport
     // block is outside the shim address space:
-    let block_start = block.as_ptr() as usize;
+    let block_start = ptr as usize;
     let block_end = block_start + BLOCK_SIZE;
     let shim_start = shim_address();
     let shim_end = shim_start + ENCL_SIZE;
@@ -295,8 +289,20 @@ unsafe extern "C" fn main(
     if (block_start >= shim_start && block_start < shim_end)
         || (block_end > shim_start && block_end <= shim_end)
     {
-        panic!();
+        panic!("Sallyport block is inside the shim address space");
     }
+
+    unsafe { &mut *(ptr as *mut [usize; BLOCK_SIZE / size_of::<usize>()]) }
+}
+
+unsafe extern "C" fn main(
+    block_ptr: *mut u8,
+    ssas: &mut [StateSaveArea; NUM_SSA],
+    cssa: usize,
+    tcb: &mut MaybeUninit<Tcb>,
+) -> i32 {
+    // Enable exceptions:
+    ssas[cssa].extra[0] = 1;
 
     let mut ret = 0;
 
@@ -337,6 +343,7 @@ unsafe extern "C" fn main(
         1 => {
             // cssa == 0 already initialized the TCB
             let tcb = tcb.assume_init_mut();
+            let block = validate_block_ptr(block_ptr);
             handler::Handler::handle(
                 &mut ssas[0],
                 block.as_mut_slice(),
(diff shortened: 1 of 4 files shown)


crates/shim-sgx/src/main.rs
```diff
@@ -337,17 +345,24 @@ unsafe extern "C" fn main(
...
+            let tcb = tcb.assume_init_mut();
+            let block = validate_block_ptr(block_ptr);
+            handler::Handler::finish(&mut ssas[1], Some(block.as_mut_slice()), tcb)
+        }
+        3 => {
             let tcb = tcb.assume_init_mut();
-            handler::Handler::finish(&mut ssas[n - 1], block.as_mut_slice(), tcb)
+            handler::Handler::finish(&mut ssas[2], None, tcb)
         }
+        _ => panic!("CSSA > 3"),
```
bstrie, line 365: Do we actually want to panic here? The panic handler at this point is an infinite loop, right? Is this the safest/most useful approach?

crates/shim-sgx/src/main.rs
```diff
@@ -275,14 +276,9 @@ global_asm!(
...
-unsafe extern "C" fn main(
-    block: &mut [usize; BLOCK_SIZE / core::mem::size_of::<usize>()],
-    ssas: &mut [StateSaveArea; 3],
-    cssa: usize,
-    tcb: &mut MaybeUninit<Tcb>,
-) -> i32 {
-    // Enable exceptions:
-    ssas[cssa].extra[0] = 1;
+fn validate_block_ptr(ptr: *mut u8) -> &'static mut [usize; BLOCK_SIZE / size_of::<usize>()] {
+    is_aligned_non_null::<[usize; BLOCK_SIZE / size_of::<usize>()]>(ptr as usize).unwrap();
```
bstrie, line 280: Similar to my previous comment, is panicking the appropriate thing to do here? Is panicking-on-error normal for the shims?

crates/shim-sgx/src/main.rs
```diff
@@ -275,14 +276,9 @@ global_asm!(
...
+    let block = unsafe { &mut *(ptr as *mut [usize; BLOCK_SIZE / size_of::<usize>()]) };
```
bstrie, line 281: Safety comment. Is the assertion on the previous line enough to guarantee that this is safe? Can we move the construction of the block below the `if (block_start ...)` check below? Instead of `block.as_ptr() as usize` below can we just use the original `ptr` instead for that check?

src/backend/sgx/thread.rs
```diff
@@ -171,9 +190,9 @@ impl super::super::Thread for Thread {
         //
         // After the sallyport rework, we can test the sallyport itself and
         // remove this logic.
-        if self.cssa > 0 {
+        if self.cssa == 1 || self.cssa == 2 {
```
bstrie, line 193: Just a reminder to follow up with Roman and see if we can follow the above comment and remove this.

crates/shim-sgx/src/main.rs
haraldh: It's safe, but not useful. I might change panic to loop on the `hlt` opcode in the future.

crates/shim-sgx/src/main.rs
haraldh: Yes, definitely panic, if the host is playing games with us here, because it is a malicious host.

crates/shim-sgx/src/main.rs
haraldh: good point!

haraldh: fixed

src/backend/sgx/thread.rs
```diff
@@ -121,10 +130,20 @@ impl super::super::Thread for Thread {
         let exit_status = exit_status as i32;
 
         self.how = match run.function as usize {
-            EENTER | ERESUME if run.vector == Vector::InvalidOpcode => EENTER,
-
-            #[cfg(feature = "gdb")]
-            EENTER | ERESUME if run.vector == Vector::Page => EENTER,
+            EENTER | ERESUME if run.vector == Vector::InvalidOpcode => {
+                //trace!("Vector::InvalidOpcode cssa={} EENTER", self.cssa);
```
rvolosatovs, line 134: Why commented-out trace log? Either remove or uncomment

src/backend/sgx/thread.rs
rvolosatovs: Yes, we can parse the sallyport block to determine what requests are there, but why would removing this conditional be a desirable thing? How would you differentiate between the two different blocks?
Even if there'd only be a single block, it's well-defined that sallyport is only executed in CSSA 1 (and in this PR, 1 and 2), so why even bother parsing anything in CSSA 0 or 3 ?

By the way, does the block get reset to 0 on each CSSA state decrease? (and if not, don't we run into a risk of executing an already-executed block again in CSSA 3?)

From the context provided by @haraldh today at the status meeting, I think the implementation is correct and we should simply remove the comment.

crates/shim-sgx/src/main.rs
```diff
@@ -40,6 +40,7 @@ fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
 
 use noted::noted;
 use primordial::Page;
+use sallyport::util::ptr::is_aligned_non_null;
 use sallyport::{elf::note, REQUIRES};
```
rvolosatovs, line 44: inconsistent import styling

src/backend/sgx/thread.rs
haraldh: removed stale debug log

crates/shim-sgx/src/main.rs
haraldh: fixed

src/backend/sgx/thread.rs
haraldh: I think we should test, if the sallyport does contain s.th. useful before processing it in that CSSA.

haraldh: And the "reset" should happen inside the shim after processing the host return data in the sallyport crate.

haraldh: `for item in block` ... this iterator should not return anything, if the sallyport was not used/was reset.

haraldh: removed the comment, though and added a new issue https://github.com/enarx/enarx/issues/2400

rvolosatovs (approved):

PR #2398, opened 2022-12-16
fix: remove unused crt0 environment variables

rvolosatovs:
This also removes hard-coded `debug` level log in `exec-wasmtime`

### DIFF
--- a/crates/shim-kvm/src/exec.rs
+++ b/crates/shim-kvm/src/exec.rs
@@ -131,20 +131,6 @@ fn crt0setup(
     builder.push("/init").unwrap();
     let mut builder = builder.done().unwrap();
     builder.push("LANG=C").unwrap();
-    // FIXME - v0.1.0 KEEP-CONFIG HACK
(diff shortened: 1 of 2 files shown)


haraldh (approved):

haraldh:
Are you planning to add a `wasmtime_log=` config to Enarx.toml?
The evaluation should be guarded with `cfg(feature=dbg)` then.

rvolosatovs:
> Are you planning to add a `wasmtime_log=` config to Enarx.toml?
> The evaluation should be guarded with `cfg(feature=dbg)` then.

I have added `log_level` and `profile` as arguments passed via fd 3 to `exec-wasmtime`, where the level is just taken from `ENARX_LOG`.
Since arguments should be measured, that should not posses any security risks.
I'll file an issue around this, where we can discuss whether we want a more fine-grained control or expose it via Enarx.toml.
You're also welcome to review the PR, once I update it on Monday

PR #2406, opened 2022-12-19
feat(sgx): add page fault handling and lazy mmap

haraldh:
This will allow to map the whole 2GB address range of the wasm application, which will speedup execution a lot.

### DIFF
--- a/crates/shim-sgx/src/handler/mod.rs
+++ b/crates/shim-sgx/src/handler/mod.rs
@@ -28,8 +28,7 @@ pub(crate) mod key;
 pub(crate) mod usermem;
 
 use crate::handler::usermem::UserMemScope;
-use crate::heap::Access;
-use crate::heap::Heap;
+use crate::heap::{Access, Heap};
 use crate::thread::{
     NewThread, NewThreadFromRegisters, Tcb, Tcs, ThreadMem, NEW_THREAD_QUEUE, THREADS_FREE,
     THREAD_ID_CNT,
@@ -44,11 +43,12 @@ use core::arch::x86_64::CpuidResult;
 use core::ffi::{c_int, c_size_t, c_ulong, c_void};
 use core::fmt::Write;
 use core::mem::size_of;
+use core::ops::Deref;
 use core::ptr::read_unaligned;
 use core::ptr::NonNull;
 use core::sync::atomic::{AtomicU32, Ordering};
-use mmledger::{Region, Span};
 
+use mmledger::{Record, Region, Span};
 use primordial::{Address, Offset, Page};
 use sallyport::guest::{self, Handler as _, Platform, ThreadLocalStorage};
 use sallyport::item::enarxcall::sgx::{Report, ReportData, TargetInfo, TECH};
@@ -57,11 +57,13 @@ use sallyport::libc::{
     off_t, pid_t, CloneFlags, SYS_clock_gettime, EACCES, EAGAIN, EINVAL, EIO, EMSGSIZE, ENOMEM,
     ENOSYS, ENOTSUP, MAP_ANONYMOUS, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE, STDERR_FILENO,
 };
+use sallyport::Error;
 use sgx::page::{Class, Flags};
 use sgx::ssa::StateSaveArea;
 use sgx::ssa::Vector;
 use spin::{Lazy, RwLock};
 use x86_64::addr::VirtAddr;
+use x86_64::structures::idt::PageFaultErrorCode;
 use x86_64::structures::paging::Page as PageAddr;
 
 // Opcode constants, details in Volume 2 of the Intel 64 and IA-32 Architectures Software
@@ -82,6 +84,42 @@ static ZERO: Page = Page::zeroed();
 
 fn is_prot_allowed(prot: c_int) -> bool {
     prot == PROT_READ || prot == (PROT_READ | PROT_WRITE) || prot == (PROT_READ | PROT_EXEC)
+}
+
+fn flags_from_access(access: Access) -> Flags {
+    let mut flags = Flags::empty();
+
+    if access.contains(Access::READ) {
+        flags |= Flags::READ;
+    }
+
+    if access.contains(Access::WRITE) {
+        flags |= Flags::WRITE;
+    }
+
+    if access.contains(Access::EXECUTE) {
+        flags |= Flags::EXECUTE;
+    }
+
+    flags
+}
+
+fn libc_from_access(access: Access) -> c_int {
+    let mut flags = 0;
+
+    if access.contains(Access::READ) {
+        flags |= PROT_READ;
+    }
+
+    if access.contains(Access::WRITE) {
+        flags |= PROT_WRITE;
+    }
+
+    if access.contains(Access::EXECUTE) {
+        flags |= PROT_EXEC;
+    }
+
+    flags
 }
 
 fn flags_from_libc(prot: c_int) -> Flags {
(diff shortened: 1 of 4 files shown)


crates/shim-sgx/src/handler/mod.rs
```diff
@@ -815,37 +956,48 @@ impl<'a> Handler<'a> {
...
+            }
 
-        for i in 0..pages {
-            let virt_addr = VirtAddr::new((addr.raw() + i * Page::SIZE) as u64);
-            // # Safety
-            //
-            // The address must be page aligned.
-            let page_addr = unsafe { PageAddr::from_start_address_unchecked(virt_addr) };
+            let region = &record.region;
+            let addr = NonNull::new(region.start.as_mut_ptr() as *mut c_void).unwrap();
```
bstrie, line 968: Should `Region` have `addr()`, `len()`, and `pages()`, methods? Or maybe `Address` could benefit from a built-in conversion to `NonNull`, since this also happens several times below.

crates/shim-sgx/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub const CSSA_1_PLUS_STACK_SIZE: usize =
     0x800000 - Page::SIZE - NUM_SSA * core::mem::size_of::<StateSaveArea>(); // 8MB - TCS - SSA
 
 /// FIXME: doc
-pub const ENCL_SIZE_BITS: u8 = 32;
+pub const ENCL_SIZE_BITS: u8 = 34;
```
bstrie, line 41: While you're here, can you address the FIXME comments above and below this?

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -996,4 +1144,117 @@ impl<'a> Handler<'a> {
 
         Ok(())
     }
+
+    fn handle_page_fault(&mut self, addr: usize, error_code: PageFaultErrorCode) -> Result<(), ()> {
+        let tid = self.tcb.tid;
+        let addr = addr & !(Page::SIZE - 1);
```
bstrie, line 1150: Is page size guaranteed to be a power of 2? If not, what happens here?

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -996,4 +1144,117 @@ impl<'a> Handler<'a> {
...
+
+        debugln!(self, "[{tid}] handle_page_fault: {addr:#x} {error_code:?}");
+
+        // check for non-valid flags
```
bstrie, line 1154: Would it be more future-proof to check for valid flags instead?

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -996,4 +1144,117 @@ impl<'a> Handler<'a> {
...
+            PROT_READ | PROT_WRITE,
+        ) {
+            debugln!(
+                self,
+                "[{tid}] handle_page_fault: ERROR mmap_host() = {e:#?}"
+            );
+            panic!();
+        }
+
+        self.mmap_guest(addr, length, flags_from_access(access));
```
bstrie, line 1237: Is this operation infallible?

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -120,6 +158,14 @@ fn access_from_libc(prot: c_int) -> Access {
     access
 }
 
+#[derive(PartialEq)]
+enum MMapStrategy {
+    /// Memory should be mmap'ed directly
+    Direct,
```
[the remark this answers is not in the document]
haraldh: Most of the time, when the shim needs memory itself in CSSA > 0, or when it knows, that the memory is consumed immediately.

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -715,6 +744,68 @@ impl<'a> Handler<'a> {
...
+        if len == 0 {
+            return Err(EINVAL);
+        }
+
+        let length = Offset::from_items((len + Page::SIZE - 1) / Page::SIZE);
+        let access = access_from_libc(prot);
+        let mut heap = HEAP.write();
+        if let Some(addr) = heap.mmap(None, length, access) {
+            let ret = NonNull::new(addr.raw() as *mut c_void).unwrap();
+            debugln!(
```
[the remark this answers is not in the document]
haraldh: the inner is guarded with 
```rust
if $crate::DEBUG {
// …
}
```
so the optimizer should remove all the stuff anyway.

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -715,6 +744,68 @@ impl<'a> Handler<'a> {
...
+        lazy: MMapStrategy,
+    ) -> Result<NonNull<c_void>, Error> {
+        if len == 0 {
+            return Err(EINVAL);
+        }
+
+        let length = Offset::from_items((len + Page::SIZE - 1) / Page::SIZE);
+        let access = access_from_libc(prot);
+        let mut heap = HEAP.write();
+        if let Some(addr) = heap.mmap(None, length, access) {
```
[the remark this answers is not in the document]
haraldh: nice

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -440,20 +445,29 @@ impl<'a> Handler<'a> {
 
     /// Finish handling an exception
     pub fn finish(ssa: &'a mut StateSaveArea, block: Option<&'a mut [usize]>, tcb: &'a mut Tcb) {
-        if let Some(Vector::InvalidOpcode) = ssa.vector() {
-            if let OP_SYSCALL | OP_CPUID = unsafe { read_unaligned(ssa.gpr.rip as _) } {
-                // Skip the instruction.
-                ssa.gpr.rip += 2;
-                return;
+        match ssa.vector() {
```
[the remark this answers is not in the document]
haraldh: yes

crates/shim-sgx/src/handler/mod.rs
```diff
@@ -497,15 +511,30 @@ impl<'a> Handler<'a> {
...
-                unreachable!()
+                let error_code =
+                    PageFaultErrorCode::from_bits_truncate(h.ssa.misc.exinfo.errcd as u64);
+                if h.handle_page_fault(h.ssa.misc.exinfo.maddr as usize, error_code)
+                    .is_ok()
+                {
+                    return;
+                }
+
+                #[cfg(feature = "gdb")]
```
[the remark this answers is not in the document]
haraldh: With `if cfg!(feature = "gdb") {`
```
540 |                     h.gdb_session();
    |                       ^^^^^^^^^^^ method not found in `handler::Handler<'_>`
```

crates/shim-sgx/src/handler/mod.rs
haraldh: added

crates/shim-sgx/src/lib.rs
haraldh: done

crates/shim-sgx/src/handler/mod.rs
haraldh: yes, on x86_64 it's always 4096.

crates/shim-sgx/src/handler/mod.rs
haraldh: good point! Thanks!

crates/shim-sgx/src/handler/mod.rs
haraldh: It unwraps internally and should really panic, because it means something is really wrong (maybe malicious host).

crates/shim-sgx/src/handler/mod.rs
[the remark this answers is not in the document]
bstrie: How would you feel about starting to use cfg-if? I bet we can find several places where it would be an improvement over our current cfgs.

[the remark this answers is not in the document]
haraldh: yeah, separate PR then

bstrie (approved):

PR #2419, opened 2022-12-21
build: make test crates part of workspace

rvolosatovs:
Note, that `cargo test --workspace` is already broken in latest `main`, a fix PR for that will follow shortly


tests/crates/enarx_syscall_tests/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-goblin = { version = "0.6", features = ["elf64"], default-features = false }
-rcrt1 = { git = "https://github.com/enarx/rcrt1", rev = "54c556fc611754521ba8825033dc8e0574d81be9", default-features = false }
-sallyport = { path = "../../../crates/sallyport", default-features = false }
-numtoa = "0.2.4"
+goblin = { workspace = true, features = ["elf64"] }
```
bstrie, line 8: The feature here should be unnecessary, the goblin specified in the workspace has this feature enabled.

rvolosatovs: Yes, but I want to make sure that the set of features required for dependencies is specified explicitly.
I don't think `workspace.dependencies` should define any features at all

bstrie (approved):

PR #2423, opened 2022-12-21
fix: make shims `cargo test --workspace` compatible

haraldh:
Closes #1824 

Signed-off-by: Harald Hoyer <harald@profian.com>

PR #2424, opened 2022-12-21
ci: various improvements in relation to `--workspace` support

rvolosatovs:
Blocked by #2423

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    after_n_builds: 9 # adjust to the number of jobs in .github/workflows/coverage.yml
+    after_n_builds: 4 # adjust to the number of jobs in .github/workflows/coverage.yml
     require_ci_to_pass: false
     wait_for_ci: false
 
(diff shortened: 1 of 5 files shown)


haraldh:
Please also coverage.yml

haraldh:
And adjust `.codecov.yml` line 3 with the number of matrix jobs in the coverage workflow

haraldh (changes requested):

rvolosatovs:
Looks like this PR is blocked on changes to https://github.com/profianinc/infra-terraform/blob/be9bd1afd473dfdb9087551ed883b5d2dac59ff1/github_branchprotections_enarx.tf#L60-L96 @puiterwijk ?

bstrie:
Does every clippy check need to be its own action? Once one action has already done a full `cargo check --workspace`, the subsequent `cargo check -p enarx-shim-kvm -p enarx-shim-sgx --target x86_64-unknown-none` takes about 4 seconds, which doesn't seem worth the compute/transfer costs of spinning up a separate runner just do do those.

haraldh (approved):

PR #2429, opened 2022-12-23
feat(sev): add lazy memory mapping

haraldh:
and configure a fast wasmtime


dpal:
I am not the right person to review this PR. Reassigned the review to Ben and Roman.

bstrie (approved):

jarkkojs:
I did check the patch set and the logic using zero page looks pretty defacto way of doing this so +1 from me.

PR #2437, opened 2023-01-08
feat: only log `enarx` workspace events

rvolosatovs:

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -24,7 +24,7 @@ use std::str::FromStr;
 use anyhow::{anyhow, bail};
 use clap::{ArgAction, Args, Parser, Subcommand};
 use tracing::info;
-use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::filter::{filter_fn, FilterExt, LevelFilter};
 use tracing_subscriber::fmt::format::FmtSpan;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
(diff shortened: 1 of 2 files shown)


src/cli/mod.rs
```diff
@@ -59,6 +59,17 @@ impl Options {
             .max_level_hint()
             .and_then(LevelFilter::into_level)
             .map(Into::into);
+
+        let target_filter = filter_fn(|meta| match meta.target() {
+            "enarx" | "enarx_exec_wasmtime" | "enarx_shim_kvm" | "enarx_shim_sgx" => true,
+            target if target.starts_with("enarx::") => true,
+            target if target.starts_with("enarx_exec_wasmtime::") => true,
+            target if target.starts_with("enarx_shim_kvm::") => true,
+            target if target.starts_with("enarx_shim_sgx::") => true,
```
bstrie, line 68: Can this `match` just become `if meta.target().starts_with("enarx") {`? This way we won't need to manually update this when adding new crates, as long as we continue to prefix them with `enarx`, and in the meantime I find it unlikely that we'll encounter any crates that start with `enarx` that aren't from us.

bstrie: > We don't add new crates too often and it'd be just a 2 line change to add

I personally doubt we'll remember to add it, but sure. :P

bstrie (approved):

src/cli/mod.rs
rvolosatovs: That's a valid concern, but then again should be easy to track down and fix.
I think vetting the list of crates we log from is more important.
I take full responsibility if it comes back biting us in the future

PR #2438, opened 2023-01-08
refactor: remove `-v` argument

rvolosatovs:
We already have `ENARX_LOG` env variable for verbosity level

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -22,9 +22,9 @@ use std::process::ExitCode;
 use std::str::FromStr;
 
 use anyhow::{anyhow, bail};
-use clap::{ArgAction, Args, Parser, Subcommand};
+use clap::{Args, Parser, Subcommand};
 use tracing::info;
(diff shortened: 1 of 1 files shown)


rjzak:
It seems to me that `-v` is worthwhile as it's a normal argument to various applications to increase verbosity.

bstrie:
@rjzak Personally, the `-v` idiom never really sat right with me, it's one thing to support just `-v`, but once you get into `-vv`, `-vvv`, `-vvvv`, it all seems a bit too cute and I'd rather see people use named log levels, which make it easier to understand what each level adds. I'm sympathetic to the idea of someday having a (non-repeating) `-v` flag in certain subcommands which might print out slightly more status info for that command, but I see that as something separate from the generalized logging infrastructure. I'm fine with removing this for now, but I'll file a UX issue for thinking about how we should expose logging in the long-term.

bstrie (approved):

PR #2447, opened 2023-01-09
feat: store CRL & originating URL

rjzak:
Cache the URL of the CRL with the CRL, with updates to the `enarx platform info` checks.

Addresses: https://github.com/profianinc/steward/pull/159#discussion_r1064699070


rvolosatovs (changes requested):
I don't think `add_crl` is the right approach. Mutating data is not a good idea in general, especially for a use case like this.

I'm not familiar with `der` or `x509` family of crates, but from looking at the docs and from what I see in this PR, looks like you need to implement `FromIterator` on `CachedCrl`, which takes in `impl IntoIterator<Item=(String, CertificateList<'a>)>`

I think the right solution would look something like (pseudocode for SNP):
```rust
# use `crls` array you already have
let crls: CachedCrl = [GENOA, MILAN].into_iter().zip(crls).collect();
let crls = crls.to_vec().context(...)
```

rvolosatovs:
> I don't think `add_crl` is the right approach. Mutating data is not a good idea in general, especially for a use case like this.
> 
> I'm not familiar with `der` or `x509` family of crates, but from looking at the docs and from what I see in this PR, looks like you need to implement `FromIterator` on `CachedCrl`, which takes in `impl IntoIterator<Item=(String, CertificateList<'a>)>`
> 
> I think the right solution would look something like (pseudocode for SNP):
> ```rust
> // use `crls` array you already had originally containing `CertificateList<'a>` values
> let crls: CachedCrl = zip([GENOA, MILAN], crls).collect();
> let crls = crls.to_vec().context(...)
> ```

Actually, you can also do it without iterators by using something like `struct CachedUrl<'a, const N: usize>([CertificateList<'a>; N])` and constructing it just like above in the original implementation in this file
`&'a [CertificateList<'a>]` should probably work as well

rvolosatovs (changes requested):
See https://github.com/rjzak/enarx/pull/4 for more useful `CrlList` structure

tests/crates/enarx_exec_tests/src/bin/sgx_attestation.rs
```diff
@@ -39,11 +39,22 @@ impl TryFrom<u64> for TeeTech {
...
+pub struct CrlPair<'a> {
+    pub url: String,
+    pub crl: CertificateList<'a>,
+}
+
+#[derive(Sequence)]
+pub struct CachedCrl<'a> {
+    pub crls: Vec<CrlPair<'a>>,
+}
+
```
rvolosatovs, line 52: You should be importing `CrlList` structure from `enarx`

tests/crates/enarx_exec_tests/src/bin/sev_attestation.rs
```diff
@@ -126,10 +126,21 @@ impl TryFrom<u64> for TeeTech {
...
+#[derive(Debug, Sequence)]
+pub struct CrlPair<'a> {
+    pub url: String,
+    pub crl: CertificateList<'a>,
+}
+
+#[derive(Debug, Sequence)]
+pub struct CachedCrl<'a> {
+    pub crls: Vec<CrlPair<'a>>,
+}
```
rvolosatovs, line 138: ditto

src/cli/mod.rs
```diff
@@ -5,7 +5,7 @@ mod deploy;
 #[cfg(enarx_with_shim)]
 mod key;
 mod package;
-mod platform;
+pub(crate) mod platform;
```
rvolosatovs, line 8: Instead of doing this, move `caching` module to top-level and make it public, then you can import it in `backend`, `cli` and the tests

rvolosatovs (changes requested):
Overall, looks good, just a small code organization issue.

tests/crates/enarx_exec_tests/src/bin/sgx_attestation.rs
rjzak: I'd have to add `enarx = { path = "../../../" }` to `tests/crates/enarx_exec_tests/Cargo.toml`, is that right? SEV structures were copied over for `sev_attestation.rs`, so there's a precedent for copying the structure over.

rjzak: Also, Enarx `src/` only has `main.rs`, no `lib.rs`.

src/cli/mod.rs
rjzak: Done

tests/crates/enarx_exec_tests/src/bin/sgx_attestation.rs
rjzak: Copied to the `lib.rs` in the exec tests crate, created an issue to track moving some common items to a common items crate in Enarx, https://github.com/enarx/enarx/issues/2451.

rvolosatovs (approved):

rvolosatovs:
I still don't understand why the structures are not duplicated, but called differently than in the original file, but whatever, let's move forward

tests/crates/enarx_exec_tests/src/lib.rs
```diff
@@ -34,3 +37,15 @@ macro_rules! musl_fsbase_fix {
         }
     }
 }
+
+// These structures are copied from `src/caching.rs`
```
rvolosatovs, line 41: This comment is not true

tests/crates/enarx_exec_tests/src/bin/sev_attestation.rs
rjzak: Resolved in the above comment

tests/crates/enarx_exec_tests/src/lib.rs
rjzak: Renamed.

haraldh (approved):

Issue #2451, opened 2023-01-11
[Feature]: Shared internal crate


rjzak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Introduce an internal crate which would store common items for various components, including unit tests. At present, some code is copy & pasted between Enarx and some integration tests, which can be reduced with such a crate.

### Acceptance Criteria

New crate in `crates/`

haraldh:
@rvolosatovs does not seem to work:
```
❯ cargo build
error: cyclic package dependency: package `enarx v0.6.4 (/home/harald/git/enarx/enarx)` depends on itself. Cycle:
package `enarx v0.6.4 (/home/harald/git/enarx/enarx)`
    ... which satisfies path dependency `enarx` (locked to 0.6.4) of package `enarx-shim-kvm v0.6.4 (/home/harald/git/enarx/enarx/crates/shim-kvm)`
    ... which satisfies path dependency `enarx-shim-kvm` (locked to 0.6.4) of package `enarx v0.6.4 (/home/harald/git/enarx/enarx)`
```

haraldh:
@rvolosatovs and yes, I guarded the dependencies with:
```
[target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
```
and
```
[target.'cfg(target_os = "none")'.dependencies]
```

rvolosatovs:
@haraldh we may need to keep some of the crates internally and reexport instead to break the cycles.
I will take a look at this as well later.

I don't have enough context from these logs to really say anything

haraldh:
my guess is that the dependency cycle checker does not take the target into account... bindeps problem

PR #2460, opened 2023-01-13
feat: factor out cpuidpage to a shared crate

haraldh:
Related: #2451


rjzak (approved):
Looks good to me

PR #2469, opened 2023-01-17
fix: set nonblocking on TCP

puiterwijk:
This makes the behaviour closer to wasmtime, and also unblocks .NET from running. .NET is expecting nonblocking sockets.


puiterwijk:
wasmtime source: https://github.com/bytecodealliance/wasmtime/blob/ebb693aa18b5ed772e8a85a8c5afeb4b18350f6a/src/commands/run.rs#L282

haraldh:
for the reference https://github.com/bytecodealliance/wasmtime/blob/main/src/commands/run.rs#L282

haraldh:
This commit removed the nonblocking:
https://github.com/enarx/enarx/pull/1500/commits/1460c0e66fc421be0588f4cba28d7d8da1964288

PR #2472, opened 2023-01-18
fix: bring release helper script up to date

bstrie:


dpal (approved):
I think this looks good to me.

PR #2480, opened 2023-01-20
feat(cli): Regularize enarx cpu info output

### LINKED ISSUE #1962: Regularize `enarx cpu info` output
Currently the default `enarx cpu info` output looks like this:

```
✗ Backend: sgx
  ✗ Driver: /dev/sgx_enclave
  ✗ AESM Daemon Socket: /var/run/aesmd/aesm.socket
  ✔ CPU: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz | GenuineIntel
  ✗  SGX Support
  ✗   Version 1
  ✗   Version 2
  ✗   FLC Support
  ✔   Max Size (32-bit): 1
  ✔   Max Size (64-bit): 1
  ✔   MiscSelect: (empty)
  ✔   Features: (empty)
  ✔   Xfrm: (empty)
  ✔   EPC Size: 0
```

Which isn't totally ideal, but is fine enough since this is just meant for human consumption.

However, if a user asks for JSON, they get stuff this:

```
      "backend": "sgx",
      "data": [
        {
          "name": "Driver",
          "pass": false,
          "info": "/dev/sgx_enclave",
          "mesg": null
        },
        {
          "name": "AESM Daemon Socket",
          "pass": false,
          "info": "/var/run/aesmd/aesm.socket",
          "mesg": null
        },
        {
          "name": "CPU",
          "pass": true,
          "info": "11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz | GenuineIntel",
          "mesg": null
        },
        {
          "name": " SGX Support",
          "pass": false,
          "info": null,
          "mesg": null
        },
        {
          "name": "  Version 1",
          "pass": false,
          "info": null,
          "mesg": null
        },
```

Some obvious problems: whereas the output pretends to be a hierarchy, the structure of the hierarchy is represented here by indenting the value of the name field. Furthermore, whereas consumers of this data would probably prefer to have proper key/value pairs, for most of this data all the values are just stored directly in the name field, with the info field being null.

Figure out a better representation for the JSON output to make it more regular to process and consume.

gandhisamay:
Formatted the CPU info output for the json option.

Fixes: https://github.com/enarx/enarx/issues/1962

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -22,6 +22,8 @@ mod parking;
 
 #[cfg(enarx_with_shim)]
 use binary::{Binary, Loader, Mapper};
+use serde_json::json;
+use serde_json::Value;
 
 use std::fs::File;
 use std::io::Read;
@@ -184,14 +186,62 @@ pub trait Backend: Sync + Send {
 
 impl Serialize for dyn Backend {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        fn leading_white_spaces(str: &str) -> usize {
+            str.chars().take_while(|c| c.is_whitespace()).count()
+        }
+
+        fn rec_insert(mut val: &mut Value, datum_json: Value, counter: usize) {
+            //each unwrap will never fail
+            if counter != 1 {
+                val = val
+                    .get_mut("data")
+                    .unwrap()
+                    .as_array_mut()
+                    .unwrap()
+                    .last_mut()
+                    .unwrap();
+                rec_insert(val, datum_json, counter - 1);
+            } else {
+                val.get_mut("data")
+                    .unwrap()
+                    .as_array_mut()
+                    .unwrap()
+                    .push(datum_json);
+            }
+        }
+
+        //Logic
+        let data = self.data();
+        let mut ser = Vec::<Value>::new();
+
+        for datum in data {
+            //Count the number of whitespaces.
+            let white_spaces = leading_white_spaces(&datum.name);
+
+            let datum_json = json!({
+                "name": &datum.name.trim(),
+                "pass": &datum.pass,
+                "info": &datum.info.unwrap_or("null".into()),
+                "mesg": &datum.mesg.unwrap_or("null".into()),
+                "data": [],
+            });
+
+            if white_spaces != 0 {
+                //This unwrap will never fail
+                rec_insert(ser.last_mut().unwrap(), datum_json, white_spaces);
+            } else {
+                ser.push(datum_json);
+            }
+        }
+
         let mut backend = serializer.serialize_struct("Backend", 2)?;
         backend.serialize_field("backend", self.name())?;
-        backend.serialize_field("data", &self.data())?;
+        backend.serialize_field("data", &ser)?;
         backend.end()
     }
 }
 
-#[derive(serde::Serialize)]
+#[derive(serde::Serialize, Debug)]
 pub struct Datum {
     /// The name of this datum.
     pub name: String,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,9 +268,9 @@ checksum = "7a40729d2133846d9ed0ea60a8b9541bccddab49cd30f0715a1da672fe9a2524"
 
 [[package]]
 name = "async-trait"
-version = "0.1.61"
+version = "0.1.63"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "705339e0e4a9690e2908d2b3d049d85682cf19fbd5782494498fbf7003a6a282"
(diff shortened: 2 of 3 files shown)


haraldh (changes requested):

src/backend/mod.rs
```diff
@@ -184,14 +186,78 @@ pub trait Backend: Sync + Send {
...
+        }
+
+        fn rec_insert(mut val: &mut Value, datum_json: Value, counter: i32) {
+            //each unwrap will never fail
+            if counter != 1 {
+                rec_insert(val.get_mut("data").unwrap(), datum_json, counter - 1)
+            } else {
+                let is_object = val.is_object();
+
+                if is_object {
```
haraldh, line 208:
```suggestion
                if val.is_object() {
```

src/backend/mod.rs
```diff
@@ -184,14 +186,78 @@ pub trait Backend: Sync + Send {
...
+                        last_prop.push(datum_json);
+                    } else {
+                        val = last_prop.last_mut().unwrap();
+                        rec_insert(val, datum_json, counter)
+                    }
+                }
+            }
+        }
+
+        //Logic
```
haraldh, line 229: Logic what?

haraldh (dismissed):

haraldh:
Don't use merge commits and squash your commits into logical ones. Don't have commits only for formatting.

haraldh:
Thanks for submitting this PR!

src/backend/mod.rs
gandhisamay: Noted!

src/backend/mod.rs
gandhisamay: Just for anyone trying to understand the code, that the overall logic can be seen from here.

gandhisamay:
I have made all the suggested changes.

haraldh:
Thanks! Please squash all your commits into one. That is sufficient for this PR.

gandhisamay:
> Thanks! Please squash all your commits into one. That is sufficient for this PR.

Done!!

nickvidal:
Thank you @gandhisamay!

bstrie:
This looks good to me. I do have two comments:

1. The approach here determines the depth of the JSON nesting based on counting leading whitespaces. However, those leading whitespaces were already determined elsewhere based on an existing hierarchy of data. In the long term, I would prefer to *only* insert whitespaces in the "pretty" output, and the JSON should be generated directly from the underlying data instead. However, that's not a user-visible change, so I'm fine leaving that for the future.
2. I'd like to have a test for the output of `enarx platform info`. The problem is that the output of this command is highly platform-specific. However, we just need the test to ensure the general structure of the JSON output, and we don't actually care about the platform info itself. This means that we could generate a dummy backend as a part of the test, implement `Backend` for it, and then have a test that serializes the dummy backend and checks the output. I'm fine leaving this for the future, but if @gandhisamay is looking for something to work on next, this would be a good place to start. :)

bstrie (approved):

bstrie:
Ah, we might need to update the nix flake in order to get the CI to pass.

gandhisamay:
@bstrie 
1)  The main problem is that the struct `Datum` in itself doesn't represent the actual hierarchy and the whitespaces are actually the ones which represent that hierarchy currently for a particular backend, so maybe what we could do is just like the `data` field present in the json structure if we have a vector of datum present in the datum struct itself then we would spare ourselves from the all this whitespaces related logic and the json output serialization will as well simply significantly and seems correct for a longer as well. (If agreed upon by everyone, I'm willing to implement this)
2) Do we want the test with a dummy backend as a part of this PR itself or create a separate issue for that?

bstrie:
> If agreed upon by everyone, I'm willing to implement this

If you'd be willing to try, then go ahead. :)

> Do we want the test with a dummy backend as a part of this PR itself or create a separate issue for that?

If you'd like to do this right now, then feel free to make it a part of this PR. However, if you're busy with other things, then we can leave that for the future and merge this PR. I would make this a higher priority than the Datum refactor.

gandhisamay:
> @gandhisamay
> 
> > If agreed upon by everyone, I'm willing to implement this
> 
> If you'd be willing to try, then go ahead. :)
> 
Cool, I will give it a try
> > Do we want the test with a dummy backend as a part of this PR itself or create a separate issue for that?
> 
> If you'd like to do this right now, then feel free to make it a part of this PR. However, if you're busy with other things, then we can leave that for the future and merge this PR. I would make the test a higher priority than the Datum refactor.
Okay then, I will write the test now and update this PR

gandhisamay:
@bstrie I have implemented the dummy backend test, could you please review it?

bstrie (approved):

PR #2488, opened 2023-01-25
fix(shim-kvm): protect against interrupt injection

haraldh:
Because the current SEV-SNP linux kernel patchset does not allow to disable interrupt injection, mitigate against:
- #VMM with error_code 0x72 (VMEXIT_CPUID)
- page_fault handler without error_code or invalid error_code


Freax13 (approved):

rvolosatovs (approved):
I trust @Freax13 's judgement

PR #2489, opened 2023-01-25
feat(shim-kvm): move consts to lib.rs

haraldh:

### DIFF
--- a/crates/shim-kvm/src/exec.rs
+++ b/crates/shim-kvm/src/exec.rs
@@ -8,6 +8,9 @@ use crate::random::random;
 use crate::shim_stack::init_stack_with_guard;
 use crate::snp::cpuid;
 use crate::usermode::usermode;
+use crate::{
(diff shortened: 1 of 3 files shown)


crates/shim-kvm/src/exec.rs
```diff
@@ -12,32 +12,21 @@ use crate::usermode::usermode;
 use core::convert::TryFrom;
 use core::sync::atomic::{AtomicBool, Ordering};
 
+use crate::{
```
rvolosatovs, line 15: nit; inconsistent import styling (there's already a `crate` import group at the top)

rvolosatovs (dismissed):

crates/shim-kvm/src/exec.rs
haraldh: oops.. bad IDE!

bstrie (approved):

PR #2527, opened 2023-03-25
chore: upgrade to wasmtime 5.0.1

rjzak:
Upgrade to Wasmtime 5.0.1 to address a security concern from `cargo deny`.

THe biggest change I see is that Wasmtime 5 made a lot of changes to how error types are handled, which made the bulk of the changes in `crates/exec-wasmtime/src/runtime/net/tls.rs`.

However, it seems that this has broken the `sgx` and `kvm` backends, in that output isn't shown on stdout. But `nil` backend works.


haraldh (approved):

rjzak:
Thank you @haraldh!

PR #2528, opened 2023-03-29
New documentation for Rust safety block comments

Tyldah:
Rust safety block comments are an essential tool for ensuring memory safety in Rust code. By marking code blocks as "safe," developers can signal to the compiler and other developers that the code has been carefully designed to avoid memory unsafety bugs. Enarx project uses a standardized format for Rust safety block comments to make them more effective and easier to read.
The above is An example of a code snippet that demonstrate the use of Rust safety block comments using the new standardized format proposed.


rjzak:
How does this signal anything to the compiler? Rust code is safe unless it's code in an `unsafe` block, or an `unsafe` function. I don't see how this helps anything.

Also, a suggestion should be an Issue, not a PR.

PR #2545, opened 2023-08-26
build(nix): update dependencies

rvolosatovs:


rvolosatovs:
@rjzak you'll have to fix the `cargo deny` failing here

rjzak:
@rvolosatovs it looks like the `cargo deny` issue would be fixed with a `cargo update` and/or updating some old crates. I'll take a look.

PR #2546, opened 2023-09-04
Fix suggested action for VCEK cache update

jethrogb:


rjzak:
@haraldh can you also force merge this since the tests are having issues, and it's a simple change? Thank you! :)

PR #2554, opened 2023-11-14
fix: calculate correct amount of EPC memory size

haraldh:
shift edx by 32bit because:

ecx: contains bits 31:12 of size in the high bits
edx: contains bits 51:32 of size in the low bits

### DIFF
--- a/src/backend/sgx/data.rs
+++ b/src/backend/sgx/data.rs
@@ -143,9 +143,9 @@ pub fn epc_size(max: u32) -> Datum {
                 break;
             }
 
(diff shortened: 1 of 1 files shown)


haraldh:
lint error not in the PR

rjzak:
Thank you, and I can see I have some changes needed on the SGX CI runner (SNP runner still a WiP).

rjzak (approved):

rjzak:
Merging since the SGX tests all pass, except for one test complaining about a blocking port for TLS, which is a known issue.

PR #2555, opened 2023-11-14
fix: use `from_bits_truncate()`

haraldh:
If enarx or the x86_64 crate does not keep up with CPU tech, using `from_bits` will prevent the usage of enarx on newer CPUs, because the checker will bailout here, although it's mostly used as an informative output.

Happens here with newer XEONs, which have the XTILECFG and XTILEDATA flags (bit 17 and 18) in Xfrm.


haraldh:
lint error not in the PR

rjzak (approved):

rjzak:
Force merged since the SGX tests passed on the SGX runner. Failures are not due to the PR, and the SEV runner doesn't work (yet).

PR #2566, opened 2023-12-16
feat: update `wasmtime` from 5 to 14

rjzak:
Wasmtime is very much outdated. However, the API has changed, especially with handling files. `wasi_common::file::FileCaps` is gone, and seemingly replaced by bit flags with far fewer options. I'm not 100% I'm doing this write.

Also making this complicated is that writing to files with trait `WasiFile` expects to be able to write to the file with a non-mutable self reference.


rjzak:
Related: https://github.com/enarx/enarx/issues/2567

rjzak:
@haraldh Am I on the right track? I'm confused by how the file capability and features was used/needed for exec-wasmtime, and what was removed by the latest Wasmtime release.

haraldh:
Hmm... `FileCaps` and `FileAccessMode` are two different things... you can't replace one with the other.

crates/exec-wasmtime/src/runtime/io/mod.rs
```diff
@@ -4,15 +4,20 @@
...
+
+pub fn stdio_file(mut file: impl WasiFile + 'static) -> (Box<dyn WasiFile>, FileAccessMode) {
     // Ensure wasmtime can detect the TTY.
-    let caps = if file.isatty() {
-        FileCaps::all().difference(FileCaps::TELL | FileCaps::SEEK)
+    let mode = if file.isatty() {
+        FileAccessMode::all()
     } else {
-        FileCaps::all()
+        FileAccessMode::READ
```
haraldh, line 20: Why not `all()`? I thought the main purpose was to take away TELL and SEEK in case of a tty... of course that does not cover the case of a pipe.

crates/exec-wasmtime/src/runtime/mod.rs
```diff
@@ -75,8 +75,8 @@ impl Runtime {
         let mut names = vec![];
         for (fd, file) in files.iter().enumerate() {
             names.push(file.name());
-            let (file, caps): (Box<dyn WasiFile>, _) = match file {
-                File::Null(..) => (Box::new(Null), FileCaps::all()),
+            let (file, access): (Box<dyn WasiFile>, _) = match file {
+                File::Null(..) => (Box::new(Null), FileAccessMode::WRITE),
```
haraldh, line 79: why not `all()`?

crates/exec-wasmtime/src/runtime/io/mod.rs
haraldh: stdio_file(stdout) only returning `FileAccessMode::READ` makes no sense somehow

haraldh: If there is no `FileCaps` anymore you can maybe remove this function altogether.

rjzak: I wasn't sure since I don't have the context as to why some decisions were made, or what was intended to be accomplished.

crates/exec-wasmtime/src/runtime/mod.rs
rjzak: There is no `all`.

```rust
pub struct FileAccessMode : u32 {
    const READ = 0b1;
    const WRITE= 0b10;
}
```

crates/exec-wasmtime/src/runtime/io/mod.rs
haraldh: It was done to mark the capabilities of the file descriptor... so a pipe does not have `tell` and `seek` capabilities.

FileAccessMode might not make sense here.

crates/exec-wasmtime/src/runtime/mod.rs
haraldh: I guess it can be removed completely then

haraldh: caps != access

PR #2617, opened 2026-02-06
use rust/vm-memory

jarkkojs:
Use the prime crate for mm stuff for the sake of steady stream of fixes and updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1285,6 +1285,7 @@ dependencies = [
  "ureq",
  "url",
  "vdso",
+ "vm-memory",
  "wat",
  "x509-cert",
  "x86_64",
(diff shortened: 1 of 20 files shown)


jarkkojs:
@rjzak, i did this partly to demonstrate to show where vm-memory is lacking. So no rush to merge but I think this is overall right. Actually I think it would be better to sort out this limit in vm-memory first and make this change later on complete.

It is better to stick what we have over here rather than be half between two worlds :-) I still wonder why CI is exploding?

jarkkojs:
@rjzak, that said I can do a separate PR of the bug fix as it is somewhat no-brainer. It addressed a failing TLS test.

PR #2618, opened 2026-05-26
Integrate lind boot

Yaxuan-w:


rjzak:
What happened to the PR? I was looking forward to reviewing, since I was having some issues dealing with the related issues on my own.

Yaxuan-w:
> What happened to the PR? I was looking forward to reviewing, since I was having some issues dealing with the related issues on my own.

@rjzak I was supposed to merge to another project:

https://github.com/Lind-Project/TriSeal/pull/14

This is an another project but based on Enarx. If you're interested in reviewing it, that would be fantastic :)

rjzak:
Would there be interest in supporting Enarx instead of having a fork? Or maybe both: support Enarx and have a small/lightweight fork? We haven't been active for a while, but I do have CI servers for AMD SNP-SEV and Intel SGX. This way we both benefit.

